PR #29, opened 2016-04-06
Angular2 initial support

fjozsef:
It adds a new menu item (Keymaps-ng2) which shows an angular2 component. This will be the starting point to the Keymaps component written in ng2.


mondalaci:
Excellent job, and a memorable first commit. Thank you!

Issue #30, opened 2016-04-10
Implement every class of the configuration to be serialized


mondalaci:
First of all, thank you very much for offering your help regarding this, @srang!

I've just finished putting together a [Google Sheet](https://docs.google.com/spreadsheets/d/1qsUgSyPOcEI3zLLjTlvH4Mo4pbCBCTjxk9LQE_vRWJw/edit#gid=0) that describes the individual classes and properties of the configuration in fair detail.

The [documentation](https://github.com/UltimateHackingKeyboard/agent/blob/master/config-serializer/README.md) that I've written up on the configuration serializer should also help.

I think it'd be a good strategy to move upwards in the tree step by step, starting from the currently implemented KeyActions class. If so, the next one would be Module, then Modules, then Layer, Layers, Keymap, Keymaps, and finally UhkConfiguration. Then you could implement ModuleConfigurations, and finally the Macros tree.

Please let me know if anything is unclear and let's discuss things here, so that everybody can see the conversation as it unfolds.

Thanks a ton!

srang:
Would you prefer a single pr with all of the serialization functionality that we review periodically or split it over multiple prs?

mondalaci:
I certainly vote for multiple PRs because that way code gets committed in regular intervals. Also gives a better sense of progress.

srang:
Sounds good I'll have Module and Modules up soon

srang:
Is pointerRoles supposed to be a single pointerRole enum or an array of pointerRole enums? In the semantics, the first module has it as a single and the description supports that but in the second and third modules, its an array, and it could make sense for a pointer module to be able to scroll and point. Additionally the name is plural so I figured it was worth checking.

mondalaci:
Good point! `pointerRole` is supposed to be a single enum, not an array. Originally, it was an array but I figured that we may never use multiple pointers per module, so an array would be overkill. Looks like I already updated `uhk-config.json`, but I forgot to update the spreadsheet, so I've just fixed it.

srang:
So I've come across an interesting bug while working on this. I've got the serializer working with a single module, however what I'm seeing is that when there are multiple items in a `ClassArray` of the same type (eg `Serializable<Module>`) the previous items in the array get overwritten with the properties of the new pushed element. For example:
```
                    "modules": [
                        {
                            "moduleId": 0,
                            "pointerRole": "move",
                            "keyActions": []
                        },
                        {
                            "moduleId": 1,
[...]

PR #31, opened 2016-04-16
Module and Modules

srang:
Here is the implementation for the Module and Modules classes. As mentioned in the [issue](https://github.com/UltimateHackingKeyboard/agent/issues/30) the assertions are commented out due to another issue with objects being overwritten when pushed into the ClassArray. Also added vim swap files to the gitignore


mondalaci:
Thanks Sam, this looks great!

Do the comparisons succeed at the end of the script?

srang:
Yup and the tslint is now clean (at least for config-serializer/**/*.ts)

mondalaci:
Awesome, music to my ears! :+1: Well done!

PR #32, opened 2016-04-17
Feature/layer layers

srang:
Add Layer and Layers (towards #30). Pending comments on name change to SwitchLayerAction->Layer enum


mondalaci:
Excellent progress, Sam - keep it up! :+1:

PR #33, opened 2016-04-18
Feature/keymap keymaps

srang:
one thing to note, lumped together keymap.id and keymap.isDefault into single uint8. I can take it out if we want but I figured we're not going to need 120+ keymaps
*edit this is for #30


mondalaci:
Excellent work. Thank you, Sam!

PR #34, opened 2016-04-18
Feature/macros

srang:
I know this is a little bit of a larger pr but I was just in a groove and wanted to hammer it out. Feel free to take extra time to make sure everything is as you expected. Also #33 is merged into this for the readString fix.

 One thing to note: in the [spec sheet ](https://docs.google.com/spreadsheets/d/1qsUgSyPOcEI3zLLjTlvH4Mo4pbCBCTjxk9LQE_vRWJw) for delay macro action it says the type is uint16 but has it in quotes, I assumed the uint16. 

Also I'm open to making macro.id a uint16 and then putting isLooped and isPrivate as flags at the beginning. It would add complexity but put to use an otherwise wasted 6 bits. On the other hand I'm still not sure how I feel about truncating IDs to be not even btye sized (would make assertions more difficult), and do we really need that many macros... Thoughts for another day
ref #30


mondalaci:
Just wow... You are a machine! I'm seriously impressed by your progress. Is there anything left at all? :)

srang:
All that is left is the full configuration. So ModuleConfigurations might be a little interesting (if you have other examples or types, could you add them) but the rest of it is a solved problem :smile: time to starting thinking validation or tying in with the gui

PR #35, opened 2016-04-19
use full byte for each bool

srang:
based on comment in #30
and because its probably a good idea


mondalaci:
Great idea on adding boolean-specific utility functions, but how about adding `UhkBuffer.{read,write}Boolean` instead? I think that'd be a more integrated approach.

Also, how about simply using 1 for true, and 0 for false?

srang:
Heh yeah that would make a lot more sense since its not being tacked on anymore. I'll update that

Issue #36, opened 2016-04-20
Validating the properties of serializables


mondalaci:
[It became apparent](https://github.com/UltimateHackingKeyboard/agent/issues/30#issuecomment-210833221) that the current implementation of `@assert*` decorators causes objects of the same class to have identical property values, which is clearly buggy behavior.

I've looked into the implementation of these decorators (within `assert.ts`) and haven't managed to fix them. But that doesn't mean that they can't be fixed because I'm not a TypeScript expert.

First, it'd make sense to create a minimal test case (saved as a Gist, for example) that reproduces the issue. It could be as simple as a class with an `@assert*` decorated property, and then creating 2 objects of that class and setting their properties to different values, and then printing those values would reveal the issue.

There's an excellent and mind-bending multi-part article on the subject: http://blog.wolksoftware.com/decorators-reflection-javascript-typescript

Given that there's some heavy TypeScript / JavaScript magic involved, this is very nontrivial. You're welcome to ask for help on Stack Overflow, Sam. Tagging the question with the typescript tag should make the question visible to knowledgable folks.

If it'll turn out that this can't be done then I have a plan B, which involves writing per property setters and getters manually. That shouldn't be too bad either but I'd like to avoid boilerplate code if possible.

srang:
https://gist.github.com/srang/eb4b28d2dcae4bab409e0ed659d2a5bb
Simple example of how the bug manifests. Note how the `surname` variable is not overwritten/shared between objects but the `name` variable is.

srang:
So it seems that the crux of this issue is that for a property decorator, the new getter and setter is only defined once per class and then the declared value is shared between all instances of that class.

mondalaci:
Excellent analysis, Sam! Could you please make a Stack Overflow question out of this, and tag it with the typescript tag? Maybe a TypeScript wizard can make this decorator work in some way.

It'd probably be useful to put all this code into a single file, so that folks would be able copy & paste it into a script, into Stack Overflow, or to https://www.typescriptlang.org/play/ easily.

Actually, I've just noticed that http://codepen.io supports TypeScript, so it'd make sense to create a pen over there for people to try it out in the browser (and also include the original version into the Stack Overflow question).

Thank you!

srang:
[...]

PR #37, opened 2016-05-07
changing how assertions access instance variables 

srang:
to make them non-static-like. Based on http://stackoverflow.com/questions/35493375/typescript-class-with-property-decorator-acts-as-if-static

Let me know if there is a better way to import the assertions
fix for #36

### DIFF
--- a/config-serializer/assert.ts
+++ b/config-serializer/assert.ts
@@ -1,68 +1,78 @@
-function assertUInt8(target: any, key: string) {
+export function assertUInt8(target: any, key: string) {
     return assertInteger(target, key, 0, 0xFF);
 }
 
-function assertInt8(target: any, key: string) {
+export function assertInt8(target: any, key: string) {
     return assertInteger(target, key, -0x80, 0x7F);
 }
 
-function assertUInt16(target: any, key: string) {
+export function assertUInt16(target: any, key: string) {
     return assertInteger(target, key, 0, 0xFFFF);
 }
 
-function assertInt16(target: any, key: string) {
+export function assertInt16(target: any, key: string) {
     return assertInteger(target, key, -0x8000, 0x7FFF);
 }
 
-function assertUInt32(target: any, key: string) {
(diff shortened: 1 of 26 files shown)


mondalaci:
Excellent work, Sam! This looks great to me, but I'd like @fjozsef to also take a look at it as our TypeScript expert.

config-serializer/assert.ts
```diff
@@ -1,68 +1,80 @@
...
     }
+
+    function setter(newVal: any) {
+        if (this[priv] !== newVal) {
+            if (newVal < min || newVal > max) {
+                throw `${target.constructor.name}.${key}: ` +
+                    `Integer ${newVal} is outside the valid [${min}, ${max}] interval`;
+            }
+            this[priv] = newVal;
+            this._dirty = true;
```
fjozsef: Why does it need to set `this._dirty` ?

config-serializer/assert.ts
```diff
@@ -1,68 +1,80 @@
...
+            return this[priv];
         }
+
+        function setter(newVal: any) {
+            if (this[priv] !== newVal) {
+                if (enumerated[newVal] === undefined) {
+                    throw `${target.constructor.name}.${key}: ${newVal} is not enum`;
+                }
+                this[priv] = newVal;
+                this._dirty = true;
```
fjozsef: Here the same.

config-serializer/assert.ts
srang: Honestly, no idea. I was following the post I linked, but I can try removing it.

fjozsef: `_dirty` is a member variable of the `TestClass` you found on the stackoverflow page since we don't have it, I think you can remove this line, and the other below at my comment.

srang: good catch. removing

fjozsef:
Squash the commits, please.

srang:
Haven't done much rebasing before but I think I got it squashed

PR #62, opened 2016-06-19
Package.json - Tslint and Stylelint fixes

NejcZdovc:
Fixed ts linter for windows and added style linter into dependencies. Renamed deprecated rule from **number-zero-length-no-unit** to **length-zero-no-unit**. 

@fjozsef is this working on linux as well?


fjozsef:
Unfortunately it doesn't wok on linux. It needs the parameters as string (between apostrophe or quotation mark).
And there is an other problem, if there is any error in the first path, then the second one isn't evaluated. 
We should run a single tslint command with multiple path.

NejcZdovc:
Ok, will try to find out another approach to make it work

fjozsef:
Okay and don't forget to rebase.

fjozsef:
We should either find a way to make it work on both windows and linux or postpone the fix later and close this.

NejcZdovc:
Let's close it for now

Issue #63, opened 2016-06-20
Side menu


NejcZdovc:
I would like to start work on side menu. I have a couple of questions before I can start.

- Can I move all content from HTML cutouts to one big div and angular to another, because now we can't toggle only content anymore, we need to toggle the whole page? I would personally suggest to split content into two html files at this point. Index.html for angular and Demo.html for static content.
- Main app need change. I suggest moving content from the current main app to new "master" keymap component and introduce new main app with only sidemenu, router and outline.
- I would suggest to start using router v3, so we need to update project to rc2 and all necessary dependencies. This is [documentation ](https://angular.io/docs/ts/latest/guide/router.html)for it.
- Is first keymap default page (home page) of the agent or will there be welcome screen / other home page?

@fjozsef and @mondalaci what do you think?

fjozsef:
- I thought we could have 3 htmls, one for the angular2, one for the keymap and one for the macro editor. I would wrap the jQuery codes into an angular2 component which would contain only an iframe. What do you think?
- Yes, we definitily need a new main app. I think you should rename the current one and create a new one. Maybe you should move some of the logic (like selecting the keymap) from the current main app to the new one, but I am sure you will see what should be moved.
- router v3: +1. I really hope that it's stable enough. I had a bad experience with the previous ones when I tried to migrate to v2 unsuccessfully.
- For now the first keymap should be the home page, but in the future maybe we'll have authentication services, who knows.

NejcZdovc:
- Which jquery code are you referring too and which iframe?
- Ok
- Same here, fingers crossed.
- Ok

I would suggest to do this in three phases. 
1) Split HTML files
2) Reorganize main app without router
3) Add router

This way we can merge PR (phases) to the master branch faster, so there will not be a lot of conflicts. @fjozsef What do you think?

mondalaci:
@NejcZdovc: Regarding the iframes, @fjozsef meant to extract to code of the legacy keymaps, and the code of macros into separate files, and loading those files within the iframe of a separate LegacyKeymap, and a Macro Angular 2 component. That way, Angular could supervise the whole DOM, and we would nicely separate (and eventually port) the legacy code. Makes sense?

Your roadmap looks good to me, but I'm interested about @fjozsef's opinion, too.

Thank you!

fjozsef:
@NejcZdovc Yes, exactly as @mondalaci described.

Issue #56, opened 2016-06-12
Popover with flexbox layout


csanyiarpad:
Rewrite the popover's layout to use `flexbox` instead of hardcoded widths.

csanyiarpad:
I've found out that the main problem was that Bootstrap defines a `max-width: 276px;` on the `.popover`. Removing this more-or-less solved the problem. Will check if there's other "_Bootstrap goodie_" that has to be overridden or this is just enough.

csanyiarpad:
On the keymap tab the placeholder image breaks the layout, this has to be fixed.

PR #65, opened 2016-06-22
Side menu

NejcZdovc:
Implementing side menu as discussed in #63.

First step is done. So what I did:

- Moved html macro and keymap to iframe
- Removed JS that will not be used in Sidemenu, because it will be implemented into TS (some JS is still there and will be removed when implementation is finished)
- Added header to main-app component, which will be than moved to real key map component (step 2)
- Added menu section in side menu Legacy (other items will be added in step 3)

@fjozsef and @mondalaci please check if this is something that you had in mind for legacy code. Code will improve dramatically when everything will be in proper components, but I wanted to add this step so that move will be progressive.

**Still to do:**

Step 2:

- [x] Create basic component for side menu (static HTML)
- [x] Create key-map component and move related html from main-app component
- [x] Clean up main-app component

Step 3:

- [x] Add routing/router
- [x] Implement logic into side-menu component
- [x] Remove unused JS code from cutouts that was replaced by TS
- [x] Code optimization


mondalaci:
Great progress, @NejcZdovc! Maybe it'd make sense to merge this as it is if @fjozsef likes it. By committing the individual steps we could minimize the chance of having to resolve large conflicts in the near future.

NejcZdovc:
Agree, that's why I splitted side menu move to angular in three phases. I hope that today/tomorrow I can finish phase two.

fjozsef:
I thought switching an angular 2 based menu can be done in a much easier way.
Now with this code, the UI has changed a lot since the original one and I am not satisfied with that. 

weird popup2
no_keyboard1

@NejcZdovc Are they supposed to look like these?
 The angular 2 based keymap doesn't work to me, popover doesn't work when I tried to click on the mockup either. Maybe I should wait for the phase 2 or 3, but now it cannot be merged.

NejcZdovc:
@fjozsef that's strange. For me everything is working without any problems. Angular based key map working OK also legacy code and popups. From your print screens I can see that there is something really wrong, because you don't have header and label on left side of mods. Seems like you don't have complete version. Do you have the latest version? On which browser are you trying this?

This is my view:
Angular2 keymap
image

Legacy code keymap
image


UI has not changed, I just added elements from mocks to angular code. Current scss angular code is very different then in mock scss. So I tried to merge them together as much as possible. If this is not correct approach and scss in angular is up to date, then I can easily revert changes regarding scss, but for me things just didn't work with current angular scss. Problem was that header was not there and everything in angular scss in based on 100% without header. That's why I needed to add some things. I also added missing parts from mocks to angular.

fjozsef:
The bug has occured to me, when I clicked on Keymap title on the left. When I click on it, the keyboard disappears.

NejcZdovc:
That's possible, because title was just copied from mockups. I will fix this in phase two when I will connect it with data.

NejcZdovc:
I tried clicking on everything that is there, but can't reproduce your error. Everything is fine here.

fjozsef:
With the latest change I cannot either. Maybe my browser's cache tricked me. Sorry, I forget the double check it. Now it works.

So about the changes:
I thought we could keep the original menu structure and we could switch to angular 2 without the user noticing the change.  
As far as I see the angular 2 keymap is the default and with the menu you can switch to the legacy ones, but you cannot switch back without refreshing the page. The legacy keymap and the angular 2 keymap should be explicitly accessible from the menu and the legacy one should be the default.

NejcZdovc:
Agree, I added this to the legacy menu for now. In phase 3 I will connect everything with data, so in legacy menu will only be iframes. Everything else will be live and connected from real data.

NejcZdovc:
So phase two is finished. What was done:

- created side-menu component and move static html there
- created key-map component and move all the logic and style from main-app to this component
- clean up index.html and main-app component
- optimise legacy component
- removed handlebars from package.json as we are not using it anymore
- added loader for html in webpack and moved html to separate files as dictated in [style guide](https://angular.io/styleguide#!#05-04) for components that I was modifying

I had quite a lot of problems with CSS, because when you add component wrappers quite a lot of things stopped working. I fixed it, but this should be checked with @csanyiarpad if everything is OK. Popup is still not aligned correctly right now. I didn't fixed this, because @csanyiarpad will implement flex in #56.

**UX change**
The only thing that I added by myself is position of key map. I am using 27" monitor and as you can see at [this comment](https://github.com/UltimateHackingKeyboard/agent/pull/65#issuecomment-228450170) it looks quite strange for me (legacy code vs my modification). That's why I aligned it to the center of the screen and set max-width, so that keyboard is not to big on big screens. What do you think @mondalaci, @fjozsef and @csanyiarpad?

@fjozsef We should fixed webpack loader, so that we could use relative URL paths for templates and not full paths.

src/components/keymap/key-map.component.ts
```diff
@@ -0,0 +1,107 @@
...
+import { UhkConfigurationService } from '../../services/uhk-configuration.service';
+
+@Component({
+    selector: 'key-map',
+    templateUrl: 'src/components/keymap/key-map.component.html',
+    styles: [require('./key-map.component.scss')],
+    directives: [SvgKeyboardPopoverComponent],
+    providers: [UhkConfigurationService]
+})
+export class KeyMapComponent implements OnInit, AfterViewInit {
```
fjozsef: Rename to KeymapComponent since keymap is one word.

src/components/keymap/key-map.component.ts
```diff
@@ -0,0 +1,107 @@
+import { Component, Input, ViewChildren, QueryList, ElementRef, OnInit, AfterViewInit, Renderer } from '@angular/core';
+import { SvgKeyboardPopoverComponent } from '../svg-keyboard-popover.component';
+import { Layers } from '../../../config-serializer/config-items/Layers';
+import { UhkConfigurationService } from '../../services/uhk-configuration.service';
+
+@Component({
+    selector: 'key-map',
```
fjozsef: Rename to keymap.

fjozsef:
Great job! I am excited about the phase 3 :)

src/main-app.component.ts
```diff
@@ -1,131 +1,17 @@
...
-        </div>
-        <div>
-            <svg-keyboard-popover *ngFor="let layer of layers.elements"
-                            [moduleConfig]="layer.modules.elements"
-                            (animationend)="onKeyboardAnimationEnd($event)"
-                            hidden>
-            </svg-keyboard-popover>
-        </div>
-    `,
+    templateUrl: 'src/main-app.component.html',
```
fjozsef: Why don't use template with require like at the styles? Then you could use relative path.

src/components/keymap/key-map.component.ts
NejcZdovc: Fixed.

src/components/keymap/key-map.component.ts
NejcZdovc: Fixed.

src/main-app.component.ts
NejcZdovc: It wasn't working for me, that's why I used full path. Can you look into it? I didn't pay too much attention about it right now. It's on my todo :)

NejcZdovc: Fixed d227946. Thanks @fjozsef

mondalaci:
@NejcZdovc So glad to see the great progress! I'd rather not center the keymap because there will be additional content below the keyboard. The idea is that the add-on modules will be visualized there. This makes sense because they are remappable on a per-keymap, per-layer basis, just like the main keyboard. It may be a good idea to slightly enlarge the keyboard just as you suggested. Looks a bit better.

NejcZdovc:
@mondalaci Ok, will remove center alignment, but leave enlargement.

package.json
```diff
@@ -36,7 +36,6 @@
     "browser-stdout": "^1.3.0",
     "es6-shim": "^0.35.1",
     "font-awesome": "^4.6.3",
-    "handlebars": "^4.0.5",
```
fjozsef: `popup__iframe.html` still uses:
`<script src="node_modules/handlebars/dist/handlebars.js"></script>`

src/main-app.component.html
```diff
@@ -0,0 +1,6 @@
+<side-menu></side-menu>
+<div id="main-content" class="split split-horizontal container-fluid main-content">
+    <key-map></key-map>
+    <legacy [link]="'keymapLegacy.html'" [class]="'keymap-legacy--edit'"></legacy>
```
fjozsef: Why bind class instead of `class="keymap-legacy--edit"`?

src/main-app.component.html
```diff
@@ -0,0 +1,6 @@
...
+    <legacy [link]="'macroLegacy.html'" [class]="'macro-legacy--edit'"></legacy>
```
fjozsef: Same problem as one line above.

NejcZdovc: This are leftovers from previous implementation. This will be removed when I will implement router, so here will only be `<router-outlet></router-outlet>`

package.json
NejcZdovc: Added back to package.json.

NejcZdovc:
@fjozsef merged with master (webpack fix). There were also two bugs fixed, that were found today.

package.json
```diff
@@ -23,7 +24,7 @@
     "webpack": "^1.13.1",
     "webpack-dev-server": "^1.14.1",
     "webpack-stream": "^3.2.0",
-    "webpack-svgstore-plugin": "^2.2.2",
+    "webpack-svgstore-plugin": "^3.0.0",
```
fjozsef: I couldn't make it work with 3.0.0. I think you should revert this to 2.2.2 for now.

package.json
```diff
@@ -44,7 +46,7 @@
     "rxjs": "5.0.0-beta.6",
     "select2": "^4.0.3",
     "sortablejs": "^1.4.2",
-    "typescript": "^1.8.10",
+    "typescript": "1.9.0-dev.20160626-1.0",
```
fjozsef: Could you update to the latest?

src/components/sidemenu/side-menu.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    constructor(private uhkConfigurationService: UhkConfigurationService) {
+    }
+
+    ngOnInit() {
+        this.keymaps = this.uhkConfigurationService.getUhkConfiguration().keymaps.elements;
+        this.macros = this.uhkConfigurationService.getUhkConfiguration().macros.elements;
+    }
+
+    private toggleHide(event: Event, view: ElementRef) {
+        let classesToggle: string[] = event.srcElement.className.split(' ');
```
fjozsef: Shouldn't we use `target` instead of `srcElement`?

src/components/sidemenu/side-menu.component.ts
```diff
@@ -0,0 +1,54 @@
...
+        // Slide down
+        else {
+            classesToggle.splice(classesToggle.indexOf('fa-chevron-down'), 1);
+            classesToggle.push('fa-chevron-up');
+
+            let positionView: number = classesView.indexOf('slide-up');
+            classesView.splice(positionView, 1);
+        }
+
+        event.srcElement.className = classesToggle.join(' ');
```
fjozsef: Here also.

src/components/keymap/keymap.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    private buttons: ElementRef[];
+    private keyboards: ElementRef[];
+    private selectedLayerIndex: number;
+    private keymapId: number = 0;
+
+    private layers: Layers;
+    private keymap: any;
+
+    private numAnimationInProgress: number;
+    private subParams: any;
```
fjozsef: Change `any` to `Subscription`.

src/components/keymap/keymap.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    private subQuery: any;
```
fjozsef: Change `any` to `Subscription`.

src/components/keymap/keymap.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    @ViewChildren(SvgKeyboardPopoverComponent, { read: ElementRef })
+    keyboardsQueryList: QueryList<ElementRef>;
+
+    private buttons: ElementRef[];
+    private keyboards: ElementRef[];
+    private selectedLayerIndex: number;
+    private keymapId: number = 0;
+
+    private layers: Layers;
+    private keymap: any;
```
fjozsef: Change `any` to `Keymap` imported from `config-serializer/config-items/Keymap.ts`.

src/components/keymap/keymap.component.ts
```diff
@@ -0,0 +1,138 @@
...
+
+            if (!isNaN(id)) {
+                this.keymapId = id;
+            }
+
+            this.keymap = this.uhkConfigurationService.getUhkConfiguration().keymaps.elements[this.keymapId];
+            this.layers = this.keymap.layers;
+        });
+
+
```
fjozsef: Remove this line.

src/components/sidemenu/side-menu.component.scss
```diff
@@ -0,0 +1,23 @@
...
+ul ul {
+    max-height: 200px;
+    transition: max-height 0.5s ease-in;
+    overflow: hidden;
+}
+
+ul ul.slide-up {
+    max-height: 0;
+    transition: max-height 0.5s ease-out;
+}
```
fjozsef: Add new line to fix stylelint error.

fjozsef:
@NejcZdovc  I have reviewed, please consider my comments.

@mondalaci @csanyiarpad Could one of you also review?

mondalaci:
@fjozsef Just reviewed the commit and your suggestions do make sense to me. I have nothing to add.

Excellent work, guys!

NejcZdovc:
@fjozsef Implemented your notes and optimized side menu toggle. Please check it out.

package.json
NejcZdovc: I can't because typescript@latest/next is already version 2.0

fjozsef: I know but If we already using the dev of typescript then why couldn't we use 2.0?

src/components/keymap/keymap.component.ts
```diff
@@ -4,6 +4,8 @@ import { ActivatedRoute } from '@angular/router';
 import { SvgKeyboardPopoverComponent } from '../svg-keyboard-popover.component';
 import { Layers } from '../../../config-serializer/config-items/Layers';
 import { UhkConfigurationService } from '../../services/uhk-configuration.service';
+import { Keymap } from '../../../config-serializer/config-items/Keymap';
+import { Subscription } from 'rxjs/Rx';
```
fjozsef: Change to `import { Subscription } from 'rxjs/Subscription';`
Importing from `rxjs/Rx` imports the whole Rx lib as far as I know.

package.json
NejcZdovc: I didn't analysed 2.0 yet, so I don't know what impact will this have on the whole project.

fjozsef: Okay, then we'll update later.

PR #66, opened 2016-07-08
Webpack

NejcZdovc:
This is now a clean branch forked from sidemenu with new webpack, so that can be tested. If everything is OK, I will remove all other commits and leave only webpack related things in this pull request, but for now, for testing purposes, I left it inside.

Source of the new configuration: https://github.com/angular/angular2-seed

@fjozsef Please check it out, it's the last commit (c9a04db) that is related to the webpack, everything else is sidemenu.


NejcZdovc:
Now only webpack is inside this PR. @fjozsef

typings.json
```diff
@@ -1,7 +1,9 @@
 {
+ "version": false,
+ "dependencies": {},
  "globalDependencies": {
-  "es6-shim": "registry:dt/es6-shim#0.31.2+20160602141504",
-  "jquery": "registry:dt/jquery#1.10.0+20160628074423",
-  "node": "registry:dt/node#6.0.0+20160621231320"
+  "core-js": "registry:dt/core-js#0.0.0+20160602141332",
+  "jquery": "registry:dt/jquery#1.10.0+20160701124751",
+  "node": "registry:dt/node#4.0.0+20160423143914"
```
fjozsef: Don't downgrade the node typings, if not neccessarry.

NejcZdovc: Fixed 913a6ed

fjozsef:
I think I have fixed the problem with webpack in https://github.com/UltimateHackingKeyboard/agent/commit/d20101c9817134201a084c2cdd5c3194acf996e6.
Could you also test it? If mine works then this can be closed.
I think the problem was that es6-shim was loaded after zone.js.
[Angular docs](https://angular.io/docs/ts/latest/guide/webpack.html) says:
"LOADING POLYFILLS
Load Zone.js early, immediately after the other ES6 and metadata shims."

NejcZdovc:
Fix is working. Closing this PR.

PR #68, opened 2016-07-13
Structural refactoring

NejcZdovc:
- [x] Move html in scss out of ts files
- [x] Create folders for component
- [x] Move logic from html file to ts file


NejcZdovc:
@fjozsef Please check

PR #72, opened 2016-08-13
Move style to angular

NejcZdovc:
- [x] Move app.scss to angular app
- [x] Create dummy notification component
- [x] Move style.scss to angular app
- [x] Create dummy html for macro, so that macro.scss can be moved into app
- [x] Fix click on sidemenu, so that link is the whole line (reported by @mondalaci)


NejcZdovc:
@fjozsef @mondalaci @csanyiarpad please check it out

mondalaci:
The commit looks great @NejcZdovc, but I rather wait for @fjozsef to confirm. Thanks for your contribution!

src/main-app.component.ts
```diff
@@ -1,14 +1,16 @@
...
 import { SideMenuComponent } from './components/sidemenu/side-menu.component';
+import { NotificationComponent } from './components/notification/notification.component';
 
 @Component({
     moduleId: module.id,
     selector: 'main-app',
     template: require('./main-app.component.html'),
     styles: [require('./main-app.component.scss')],
-    directives: [SideMenuComponent, ROUTER_DIRECTIVES]
+    directives: [SideMenuComponent, ROUTER_DIRECTIVES, NotificationComponent],
```
fjozsef: Please move the NotificationComponent to AppModule declarations. Over the time all the directives should be there since rc.5.

fjozsef:
You should delete the sass folder and remove the gulp task accordingly. When this commit is merged then I think we could get rid of gulp, too! :)

fjozsef:
I am wrong, the mockups are still using them, nevermind.

src/main-app.component.ts
```diff
@@ -1,14 +1,16 @@
...
+    encapsulation: ViewEncapsulation.None
```
fjozsef: Why have you set this to None?

src/components/popover/popover.component.scss
```diff
@@ -41,10 +42,19 @@
     padding: 10px 24px;
 }
 
-mouse-tab {
```
fjozsef: Why did you remove these?

src/components/popover/tab/keypress/keypress-tab.component.html
```diff
@@ -26,4 +26,10 @@
     <select2 #longPressSelect [data]="longPressGroups" [selectedValue]="selectedLongPressIndex.toString()" (valueChanged)="onLongpressChange($event)"
         [width]="140"></select2>
     <icon name="question-circle" title="This action happens when the key is being held along with another key."></icon>
+</div>
+
+<div class="disabled-state--text">
```
fjozsef: It is hidden, why is it added?

src/main-app.component.ts
NejcZdovc: This is our main entry point into our application. So I changed this to none, so we can have some global CSS styles like select2 style. Select2 drop down is added dynamically before the end of the body tag. So it's outside out main-app tag and we can't style it. I would use main-app.component.scss only for this type of styling, everything else is inside correct component.

src/components/popover/tab/keypress/keypress-tab.component.html
NejcZdovc: This is currently hidden, but we will toggle visibility when logic is added later. This text is shown if user select key as layer switcher key.

NejcZdovc:
@fjozsef If I delete sass files legacy files (css) will not work anymore. Can I delete all legacy things (html files, legacy component, gulp, sass folder)? If yes I would create another PR only for that.

src/main-app.component.ts
NejcZdovc: Moved.

fjozsef:
@NejcZdovc There some functionalities in the legacy pages we haven't migrated to angular 2, yet. (e.g. ordering changing in macros.) First we have to implement those, then we can remove the mentioned things.

Issue #78, opened 2016-08-15
Add keymap UI


mondalaci:
I've just created a mockup of the add keymap UI:

![agent-add-keymap](https://cloud.githubusercontent.com/assets/300402/17671638/55d1953a-6318-11e6-879d-b515c29d6cda.png)

The UI should be shown when clicking on the + button of the Keymaps row of the side menu.

The Search input should function as an incremental filter and only reveal the relevant filtered keymaps. The "x / y keymaps shown" text should also update accordingly.

The description field is not yet included into the keymap data structure but maybe it'd be useful all across Agent. I can see it being used when sharing keymaps in the future.

The keymap itself should be visualized just as when being edited but here no edit mode should be provided. The viewer should be put into view mode, so the popover should not pop up when clicking on the keys. Merely a tooltip should appear when hovering above a key showing relevant key mapping information.

Upon clicking the "Add keymap" button the relevant keymap should be added.

If a keymap with the same name already exist then let's append " (2)" to its end. If that's already exists then " (3)", and so forth.

If a keymap with the same abbreviation already exists then let's use "2" as the last letter, then increment it. For example "QWR" -> "QW2" -> "QW3" -> ... -> "QW9" -> then go over the alphabet: "QWA" -> "QWB" -> .. -> "QWZ" -> then just start from "AAA" and increment ("AAA" -> "AAB" -> ... -> "AAZ" -> ABA -> ...) until a free abbreviation is found.

This is a rough mockup and many details could probably be more polished. Smaller layer buttons may look better and maybe the "Add keymap" button should be full width.

Suggestions are welcome!

NejcZdovc:
Sounds good, but I have some question:

1. Should we add description to the keymap when implementing add keymap?
2. Should slider be included in this list? (so you can slide between layers)
3. What relevant key mapping information would you like to show in popup?
4. Where would we have list of all keymaps available for adding? Will this be a new JSON that needs to be added to the project?

mondalaci:
Yours are excellent questions, Nejc. Exactly the stuff that I should have included into the issue. :)

1. For the time being, let's only use descriptions for the hardcoded keymaps of the Add new keymap page.
2. I'm not sure what you mean. Switching between layers is already possible with the Base, Mod, Fn, Mouse buttons per layer.
3. Something like this should be displayed in a tooltip tailored on a per key action basis: Key action: keypress[newline]Scancode: A[newline]Modifiers: none[newline]Long press action: None
[...]

PR #80, opened 2016-08-17
Add keymap

NejcZdovc:
Implementing key map add screen. More info here #78.


fjozsef:
What is the progress of this? Can I review?

NejcZdovc:
No, we need to talk about this, because we need animation on key maps here as well. We would need to refactor KeymapComponent so that will be a separate component that will already include animations and layers at top. What was done so far is basically everything except animations and current list length. Will ping you tomorrow on chat to discus it further.

NejcZdovc:
Working on #82 to separate components that can be then used here.

NejcZdovc:
Everything is now done except text in the tool tip. I was checking out how it's done in popover and I think that we need helper (maybe I just didn't find it) that will took keyAction in and return all properties (key action, scan code, long press, modifiers). @fjozsef Do we have something like this?

src/components/keymap/add/keymap-add.component.ts
```diff
@@ -0,0 +1,31 @@
+import { Component } from '@angular/core';
+import { Keymap } from '../../../config-serializer/config-items/Keymap';
+import { Keymaps } from '../../../config-serializer/config-items/Keymaps';
+import { SvgKeyboardComponent } from '../../svg/keyboard/svg-keyboard.component';
+import { UhkConfigurationService } from '../../../services/uhk-configuration.service';
+
+@Component({
+    directives: [SvgKeyboardComponent],
```
fjozsef: Move to app.module

src/components/keymap/add/keymap-add.component.ts
```diff
@@ -0,0 +1,31 @@
...
+    providers: [UhkConfigurationService],
```
fjozsef: Order: selector, template, styles, providers

src/components/keymap/add/keymap-add.component.ts
NejcZdovc: Done

src/components/keymap/add/keymap-add.component.ts
NejcZdovc: Done

src/components/svg/wrap/svg-keyboard-wrap.component.scss
```diff
@@ -10,12 +10,34 @@ svg-keyboard {
     max-width: 1400px;
     position: absolute;
     left: 0;
-    transform: translateX(-100%);
+    transform: translateX(-101%);
```
fjozsef: Why did you changed it?

NejcZdovc: It was changed, because there where some traces of keyboard that supposed to be hidden.

mondalaci:
Am I right that `all-keymaps.json` contains the keymaps shown on the add keymaps page? If so, I think `preset-keymaps.json` would be a better name. Even better, how about creating a `preset-keymaps` directory and putting files per keymaps into it like `vim.json`? Then the build process would merge these files to a single file. It'd result in much better maintainability over the long term.

NejcZdovc:
@fjozsef wanted to rename all to default, so it was just renamed :smile:

PR #82, opened 2016-08-27
Layers/Keyboard component

NejcZdovc:
Move navigation of the key map to the new component, and move keyboard with animation to the svg keyboard popover component.


NejcZdovc:
@mondalaci @fjozsef Can you please checkout this branch and play around with layer and key map switching. I replaced old animation with angular one and would like to test if every single scenario is working.

NejcZdovc:
Please note that code is not done yet and still needs last checks.

src/components/layers/layers.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="text-center">
+    <span class="uhk__layer-switcher--wrapper" data-title="Layers: ">
+        <button *ngIf="len > 0" #baseButton type="button" class="btn btn-default btn-lg btn-primary" (click)="selectLayer(0)">
```
mondalaci: Do we check here and below whether all the 4 layers are present within a keymap? I think we shouldn't because all the 4 layers should always be present in every keymap.

mondalaci:
According to my understanding a new component has been created which includes the layer buttons and the keyboard and now this component is used for keymaps. Is it right?

I assume that clicking on the mouse layer results in no action because there is no mouse layer yet in the model itself. Right?

The only minor issue I can notice is that if the animation is running and I click on another layer button then the click event gets ignored. It's quite an edge case but maybe it'd be worth handling.

src/components/layers/layers.component.html
NejcZdovc: Done in 6bf6de2

NejcZdovc:
Yes, new component was created so that this component can be reused on edit and add key map page. Current implementation couldn't be reused in multiple places.

Now all layers have key map, but there is no mapping added for keys. It's only simple key map, so that layer switching is working correctly.

I can disable clicking on buttons while animation is happening. Should I do this? Currently you can interrupt active animation and switch to the new one without any problems. @mondalaci

mondalaci:
I've just realized that previously I messed up the git commands and haven't even checked your branch. Silly me!

Now I did actually check out your branch and the animation works perfectly, even when clicking across the layer buttons in rapid succession.

The animation delay feels a bit long, though. The original value of the mockup was faster and felt more responsive. I think we should use the original value all across Agent.

NejcZdovc:
Animation delay and type was fixed in e19926f. Can you please try it again? @mondalaci

mondalaci:
Now the delay feels pretty darn good. Thank you!

How about extracting the 400ms delay in some way? We could reference it then instead of copy-pasting it everywhere.

NejcZdovc:
As far as I know, correct me if I am wrong, but we only use animation here, so I don't see real reason to extract it. The only scenario that can be useful if extracted is if we would add this to the setting page, so that user can define animation time for key map slider.

mondalaci:
For the time being, it's ok they way it is. I'm rather concerned about the state of the codebase over the long term. If we end up using animation delays across the whole UI here and there let's make sure that we reference the same value instead of copy-pasting it, or even worse: accidentally using different, arbitrary values.

NejcZdovc:
I agree with you 100% regarding code base. The only thing is that animations can variate depending on the elements or things that you want to do. So we are not using same speed and transition for all animations. For the slider we use 400ms, for tooltips we have 150ms, etc. This was my point that I wanted to make. Animation is something that I don't think is that straight forward and global.

mondalaci:
I wasn't conscious that different components should have different values. You're right on the money! :)

src/components/keymap/keymap.component.ts
```diff
@@ -1,47 +1,32 @@
...
 import { Keymap } from '../../../config-serializer/config-items/Keymap';
 import { Subscription } from 'rxjs/Subscription';
+import { LayersComponent } from '../layers';
 
 @Component({
     selector: 'keymap',
     template: require('./keymap.component.html'),
     styles: [require('./keymap.component.scss')],
-    directives: [SvgKeyboardPopoverComponent],
+    directives: [SvgKeyboardWrapComponent, LayersComponent],
```
fjozsef: New directives should be added to the NgModule, because adding this way (in the component) is deprecated since rc.5 and we want to avoid the increasing of deprecated code.

config-serializer/config-items/Layer.ts
```diff
@@ -5,6 +5,7 @@ import {UhkBuffer} from '../UhkBuffer';
 export class Layer extends Serializable<Layer> {
 
     modules: Modules;
+    animation: string = 'none';
```
fjozsef: We should restrict the animation type with the valid ones. Typescript allows the define a type as a set of strings like:
`animation: 'none' | 'leftIn' | 'leftOut' | 'rightIn' | 'rightOut';`. The correct types should be defined elsewhere not inlined as a described here, but right now I don't remember the correct syntax to define it with `type`.

fjozsef:
The bottom of the keymap is not visible to me. What do I do wrong?
bottom

src/components/popover/popover.component.scss
```diff
@@ -3,6 +3,8 @@
     flex-direction: column;
     max-width: none;
     padding: 0;
+    left: calc(50% - 285px);
```
fjozsef: Where is this 285 px come from? Please add comment or constant to the code the explain the value.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        this.currentLayer = 0;
+    }
+
+    ngOnInit() {
+        this.layers[0].animation = 'leftIn';
+    }
+
+    ngOnChanges() {
+        this.currentLayer = 0;
+        this.layers.map((layer: Layer) => layer.animation = 'none');
```
fjozsef: Use `forEach` instead of `map`.

mondalaci: I agree, and would like to add an explanation. In this case, the intention is to assign a value to the property of objects which is indeed achievable with `map`, but `map` is rather supposed to be used for mapping objects to other objects. `forEach` will make the intention of the code easier to understand in this case. It's easy to spot these cases because the return value of `map` is unused.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -0,0 +1,156 @@
...
+})
+export class SvgKeyboardWrapComponent implements OnInit {
+    @Input() layers: Layer[];
+    @Input() popoverEnabled: boolean = true;
+    @Input() animationEnabled: boolean = true;
+
+    private popoverShown: boolean;
+    private keyEditConfig: { moduleId: number, keyId: number };
+    private popoverInitKeyAction: KeyAction;
+    private module: Module[];
```
fjozsef: Isn't `module` the same as `this.layers[this.currentLayer].modules.element` ? If it is, then I don't undestand why we should maintain a shortcut for it.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        ])
+    ]
+})
+export class SvgKeyboardWrapComponent implements OnInit {
+    @Input() layers: Layer[];
+    @Input() popoverEnabled: boolean = true;
+    @Input() animationEnabled: boolean = true;
+
+    private popoverShown: boolean;
+    private keyEditConfig: { moduleId: number, keyId: number };
```
fjozsef: I wouldn't store this config as a member.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        this.popoverInitKeyAction = keyAction;
+        this.popoverShown = true;
+    }
+
+    hidePopover(): void {
+        this.popoverShown = false;
+        this.popoverInitKeyAction = undefined;
+    }
+
+    changeKeyAction(keyAction: KeyAction): void {
```
fjozsef: It should get the keyEditConfig as a parameter.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -0,0 +1,156 @@
...
+import { SvgKeyboardComponent } from '../keyboard';
+import { PopoverComponent } from '../../popover';
+import { Module } from '../../../../config-serializer/config-items/Module';
+import { Layer } from '../../../../config-serializer/config-items/Layer';
+
+@Component({
+    selector: 'svg-keyboard-wrap',
+    template: require('./svg-keyboard-wrap.component.html'),
+    styles: [require('./svg-keyboard-wrap.component.scss')],
+    directives: [SvgKeyboardComponent, PopoverComponent],
```
fjozsef: Move this directives to the NgModule, too.

src/components/svg/wrap/svg-keyboard-wrap.component.scss
```diff
@@ -0,0 +1,21 @@
+:host {
+    width: 100%;
+    height: 100%;
+    position: relative;
+    display: block;
+}
+
+svg-keyboard {
+    width: 95%;
+    max-width: 1400px;
```
fjozsef: Where is this value come from?

src/components/keymap/keymap.component.ts
NejcZdovc: Done

src/components/popover/popover.component.scss
NejcZdovc: It was changed, so this if not fixed anymore.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: I think the code is more readable for example:

`this.modules.elements.keyActions.elements[keyId]`

this is more readable for me than this:

`this.layers[this.currentLayer].modules.elements.keyActions.elements[keyId]`

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: This is your code :smile:

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: This is your code :smile: , didn't change anything here.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: Done

src/components/svg/wrap/svg-keyboard-wrap.component.scss
NejcZdovc: This value is used to limit keyboard width on big screens. This is not a new code.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
fjozsef: Sorry, you're correct, it's okay for now. I will refactor it after this lands.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    onKeyClick(moduleId: number, keyId: number, module: Module[]): void {
+        if (!this.popoverShown) {
+            this.keyEditConfig = {
+                moduleId,
+                keyId
+            };
+
+            this.module = module;
```
fjozsef: How about changing this line to `let modules: Module[] = this.layers[this.currentLayer].modules;` ?

src/components/svg/wrap/svg-keyboard-wrap.component.ts
fjozsef: See my comment below.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
fjozsef: Old me, why did you do this? :D

config-serializer/config-items/Layer.ts
NejcZdovc: Done

src/components/svg/wrap/svg-keyboard-wrap.component.scss
fjozsef: Okay, I have found it source.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: I don't really agree with this, but it was fixed.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -103,8 +100,14 @@ export class SvgKeyboardWrapComponent implements OnInit {
 
     ngOnChanges() {
         this.currentLayer = 0;
-        this.layers.map((layer: Layer) => layer.animation = 'none');
-        this.layers[0].animation = 'leftIn';
+        if (this.layers.length > 0) {
```
fjozsef: It is unnecessary.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -103,8 +100,14 @@ export class SvgKeyboardWrapComponent implements OnInit {
...
+            this.layers.forEach((element) => {
+                element.animation = 'none';
+
+                return element;
+            });
+            this.layers[0].animation = 'leftIn';
```
fjozsef: Only this should be in the if.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: Fixed.

NejcZdovc:
@fjozsef fixed keyboard height

PR #85, opened 2016-08-27
refactor: Layers

fjozsef:
@NejcZdovc please review!

### DIFF
--- a/src/components/layers/layers.component.ts
+++ b/src/components/layers/layers.component.ts
@@ -17,6 +17,10 @@ export class LayersComponent {
     }
 
     selectLayer(index: number) {
+        if (this.current === index) {
+            return;
+        }
+
         this.select.emit({
             oldIndex: this.current,
             index: index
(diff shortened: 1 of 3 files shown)


src/components/layers/layers.component.ts
```diff
@@ -9,43 +7,21 @@ import {
...
-            this.renderer.setElementClass(this.buttons[this.current].nativeElement, 'btn-primary', true);
-            this.selectedLayerIndex = 0;
-        }
+    constructor() {
+        this.buttons = ['Base', 'Mod', 'Fn', 'Mouse'];
+        this.current = 0;
     }
 
     selectLayer(index: number) {
-        if (index === this.selectedLayerIndex) {
```
NejcZdovc: Please add this back, so that if you click on current button, you don't trigger the whole tree.

NejcZdovc:
Great refactor, this was on my to-do, because this code really needed it.

src/components/layers/layers.component.ts
fjozsef: Done

fjozsef:
Yes, I don't understand how could I not notice earlier.

PR #86, opened 2016-08-28
Folder restructuring

fjozsef:
@mondalaci @NejcZdovc Please review!


mondalaci:
Looks pretty good to me, although the scope of these sweeping change is rather overwhelming, so it's easy to miss problems.

What about the new `vendors` directory? Is it the new hipster thing to use it instead of `node_modules`? :)

fjozsef:
The directory name `vendors` is a convention you can read about it [here](https://tommcfarlin.com/the-vendor-directory/). 
As you can see in the webpack.config.js I am collecting the files we need from `node_modules`, so if we want to deploy all the files will be ready in the dist folder.

mondalaci:
The referenced article talks about the `vendors` directory mostly in the context of WordPress. How usual it is to use in node context?

What is the hard rule for this? Is `node_modules` supposed to be totally replaced by `vendors`?

fjozsef:
It isn't WordPress specific, It is common in all over the web apps. 
See angular material 2 [demo index](https://github.com/angular/material2/blob/master/src/demo-app/index.html), they are using it, too.
And no it's not the replacement of `node_modules`. If you have problem about distingiush the two then if you want to then we could talk about it on hangout. But I am sure that you can find other articles about it on google.

mondalaci:
So according to my understading `vendors` contains the assets that are needed to build the application, right? Either way, I trust your judgement and I'm not against it.

But there's a very important question to decide about: should we use it's singular `vendor` or plural `vendors` variant? :) The former is used the the Angular 2 demo you referenced.

fjozsef:
No, we have an `assets` folder, too.  In the `assets` folder will be anything which is ours, but not code (I am not sure where the jsons should be) and in the `vendor` folder there will be anything which is 3rd party.

I am considering renameing it to singular since I saw it on material :)

mondalaci:
`assets` seems to be a good place for the json files. I don't worry too much about them because they'll be removed eventually anyways.

`vendor` is the way to go then, as God intended. :)

fjozsef:
I don't think that every json will be removed, but this is not the purpose of this pr to discuss that.

mondalaci:
Didn't mean every json, only `agent-state.json` and `uhk-state.json`. I agree that let's not talk about this right now.

fjozsef:
So I am still waiting for @NejcZdovc confirmation before I merge this. I modify "assignees" according to this. We should get to use it to use these features :)

NejcZdovc:
@fjozsef Looks good.

Issue #42, opened 2016-05-08
Migrate font icons to SVG.


csanyiarpad:
Font Awesome uses font icons whose quality is [not great](https://css-tricks.com/icon-fonts-vs-svg/).

We purchased the [IcoMoon Ultimate Font Pack](https://icomoon.io/preview-ultimate.html) and we're about to migrate to it and remove Font Awesome from the project. Going forward, Font Awesome should not be used.

According to our license, we're not permitted to upload every IcoMoon Ultimate font to the repo but we can upload the icons that are needed. Contributors, please let us know which IcoMoon Ultimate font you need and we'll add it to the repo.

Here's a non-complete list of font icons to be migrated:
- [ ] layer
- [ ] none action
- [ ] trash
- [ ] pencil (macro edit action)
- [ ] star
- [ ] cog

In the meantime, we've also purchased Font Awesome 5 Pro, so let's wait until it's released and then migrate to it.

NejcZdovc:
@mondalaci this is still something that need's to be done right?

mondalaci:
@NejcZdovc Correct.

mondalaci:
Replace the current Font Awesome 4 font icons with Font Awesome 5 SVG icons according to the following rules:

1. Almost always, the icon name stays the same. Use the solid version of the icons.
2. Remove the `-o` part when the icon name ends in it. Please note that for the star icon, the solid icon version must be used for enabled state and the regular version for disabled state.
3. The following icons must be renamed:

* gear -> cog
* sliders -> sliders-h
* files-o -> copy

4. Replace glyphicons with the following Font Awesome 5 SVG icons:

* glyphicon-trash -> trash
* glyphicon-option-vertical -> grip-vertical
* glyphicon-pencil -> pen

PR #87, opened 2016-08-28
Macro UI

laxu:
Added partially working macro UI for  #73 . Mainly looking for input on what should be done differently as I haven't worked with Angular 2 before or if I'm doing something the wrong way or not using some existing component where I should.

Macro editing popover UI apes key remap UI so it should feel familiar and consistent to use.

### What works:

- Loading existing macro from UHK config file
- Listing macro actions
- Editing macro name
- Editing macro action
- Deleting macro action
- Saving macro action in UI and displaying the updates in macro actions list
- Editing "Delay" type macro action
- Editing "Text" type macro action
- Adding new macro to sidebar list

### Added features:
- Popover UI for editing macro action
- Contenteditable component for editing macro name
- More icon options to icon component
- clone method to MacroAction. It creates a copy of the MacroAction, used to edit the macro action without changing the original.
- Placeholders for all macro action types to MacroItemComponent

### Changed:
- MacroItemComponent icons for various actions to better represent the action. Less black squares. :)

### What doesn't work:
- New added macro cannot be edited/loaded (it's not added to the correct location to be loaded in MacroComponent).
- Editing key press/hold/release macro actions (needs component taking macro functionality into account).
- Editing any mouse macro actions (needs component taking macro functionality into account)
- Changing type of macro being edited. For example if it's delay type and you click text tab it should be changed to text type. Should probably just use the toJsObject method of MacroAction to handle data in the popover as generic data before turning it back to a real MacroAction upon save.
- Saving macros (to the keyboard I suppose) where they can be loaded again  when coming back to Agent.
- Bug in ng2-slider-component ignores the stepValue so delay slider doesn't move at 0.1s steps like it should.


src/components/macro/macro-popover/macro-popover.component.ts
```diff
@@ -0,0 +1,118 @@
+import {Component, OnInit, OnChanges, Input, Output, EventEmitter, ViewChild } from '@angular/core';
+import {NgIf, NgSwitch, NgSwitchCase } from '@angular/common';
```
fjozsef: Importing these not needed.

src/components/macro/macro-popover/macro-popover.component.ts
```diff
@@ -0,0 +1,118 @@
...
+@Component({
+    moduleId: module.id,
+    selector: 'macro-popover',
+    template: require('./macro-popover.component.html'),
+    styles: [
+        require('../../popover/popover.component.scss'),
+        require('./macro-popover.component.scss')
+    ],
+    host: { 'class': 'popover macro-action' },
+    directives: [
```
fjozsef: Don't add directives here. It is deprecated since rc.5 add them in app.module.ts

fjozsef:
@mondalaci Maybe you want to check the UI out.

From the angular2 side I think the code is okay more or less. 
Just one thing before I forget: I would move the MacroItemComponent to under the components/macro folder. 
Keep working on! :)

(If I may suggest to rebase before you continue your work.)

mondalaci:
Thank you for your contribution, @laxu! Yours is a definite step forward, and it's great to be able to interact with the macro editor.

Although the mockup of the macro UI is not fully ready yet, back in the days we agreed with Arpi that we shouldn't use a popover for this purpose. Please let me make some mockups.

Other than that, some comments:
* We [shouldn't use](https://github.com/UltimateHackingKeyboard/agent/issues/42) Font Awesome anymore but IcoMoon Ultimate.
* The macro editor should be implemented as a standalone component because it will be reused elsewhere in viewer mode. (I'm not sure how it's implemented right now but this is an important point.)
* The pencil icon shouldn't be featured next to the macro name above.
* The macro name above should have a dotted underline just as in the case of keymap names and abbreviations. The text should be editable the same manner, and the same styles should apply. The interaction is the same when editing a title of a Google Document.

Gonna follow up regarding the macro UI mockup.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -9,7 +9,7 @@ import {MacroTextTabComponent} from './tab/macro-text/macro-text.component';
 import {MacroDelayTabComponent} from './tab/macro-delay/macro-delay.component';
 import {MacroMouseTabComponent} from './tab/macro-mouse/macro-mouse.component';
 
-import {MacroItemComponent} from '../../popover/tab/macro/macro-item.component';
+import {MacroItemComponent} from '../macro-item/macro-item.component';
```
fjozsef: Create an index.ts in the macro-item folder export the MacoItemComponent from there and use '../macro-item' here as a path.

src/components/macro/macro-action-editor/tab/macro-delay/macro-delay.component.html
```diff
@@ -0,0 +1,31 @@
...
+                 step="0.1" 
+                 placeholder="Delay amount" 
+                 class="form-control" 
+                 [(ngModel)]="delay" 
+                 (ngModelChange)="setDelay($event)">
+        </div>
+    </div>
+    <div class="row macro-delay__presets">
+        <div class="col-xs-12">
+          <button class="btn btn-sm btn-default" (click)="setDelay(0.3)">0.3s</button>
```
fjozsef: Create an array in typescriőt and use ngFor here.

src/app.module.ts
```diff
@@ -13,17 +14,22 @@ import { MacroComponent } from './components/macro';
 import { LegacyLoaderComponent } from './components/legacy-loader';
 import { NotificationComponent } from './components/notification';
 import { SvgKeystrokeKeyComponent, SvgOneLineTextKeyComponent, SvgTwoLineTextKeyComponent } from './components/svg/keys';
+import { ContenteditableModel } from './components/directives/contenteditable.component';
+import { MacroActionEditorComponent } from './components/macro/macro-action-editor/macro-action-editor.component';
 
 @NgModule({
     declarations: [
+        FORM_DIRECTIVES,
```
fjozsef: I think we should rather use `FormsModule` according to [this](https://angular.io/docs/ts/latest/guide/forms.html).

src/components/directives/contenteditable.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (isPropertyUpdated(changes, this.lastViewModel)) {
+      this.lastViewModel = this.model;
+      this.refreshView();
+    }
+  }
+
+  onKeypress(event:any) {
+    if (this.updateOnEnter && (event.which === KEY_ENTER)) {
+      // Finish editing when Enter pressed
+      this.elRef.nativeElement.blur();
```
fjozsef: Use render's invokeElementMethod to call blur.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,112 @@
...
+    Delay
+};
+
+@Component({
+    moduleId: module.id,
+    selector: 'macro-action-editor',
+    template: require('./macro-action-editor.component.html'),
+    styles: [ require('./macro-action-editor.component.scss') ],
+    host: { 'class': 'macro-action-editor' },
+    directives: [
```
fjozsef: Move directives to app.module.ts

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,112 @@
...
+            // Make an editable clone of macro action so original isn't changed
+            this.editableMacroAction = this.macroAction.clone();
+            let tab: TabName = this.getTabName(this.editableMacroAction);
+            this.selectTab(tab); 
+        }
+    }
+
+    onCancelClick(): void {
+        this.isEnabled = false;
+        this.cancel.emit(undefined);
```
fjozsef: It could be called without passing a parameter to emit.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,112 @@
...
+        MacroMouseTabComponent,
+        MacroTextTabComponent,
+        MacroDelayTabComponent
+    ]
+})
+export class MacroActionEditorComponent implements OnInit {
+    @Input() macroAction: MacroAction;
+
+    @Output() save = new EventEmitter<any>();
+    @Output() cancel = new EventEmitter<any>();
```
fjozsef: Change the type to `EventEmitter<void>`

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,102 @@
...
+@Component({
+    moduleId: module.id,
+    selector: 'macro-mouse-tab',
+    template: require('./macro-mouse.component.html'),
+    styles: [
+        require('../../macro-action-editor.component.scss'),
+        require('./macro-mouse.component.scss')
+    ],
+    host: { 'class': 'macro__mouse' },
+    directives: [ FORM_DIRECTIVES ]
```
fjozsef: I think you can remove this line, since `FormsModule` will be imported in app.module.ts

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,92 @@
...
+import { MacroItemComponent } from './macro-item/macro-item.component';
+
+import { ContenteditableModel } from '../directives/contenteditable.component';
 
 @Component({
     selector: 'macro',
     template: require('./macro.component.html'),
-    styles: [require('./macro.component.scss')]
+    styles: [require('./macro.component.scss')],
+    providers: [UhkConfigurationService]
```
fjozsef: Since https://github.com/UltimateHackingKeyboard/agent/commit/dee9c1077bb06014be28f09fdcb15f36f94cc884 the `UhkConfigurationService˛is provided in the app.module.ts so you don't need to provide an other.

NejcZdovc:
@laxu I think that it's time that you do rebase, because you are doing things that are already in master branch. Also structure of the folders changed quite a bit. Just a friendly advise :wink:

fjozsef:
Please rebase again, we have updated our angular2 version to rc.6.

laxu:
Ok, status report time. Since my initial pull request I've changed a lot of stuff. 

### What's new 
- The macro action editor has been refactored into a component that appears under each action item as per @mondalaci's UI mockups. Only a single editor can be open at a time and clicking another action item will instead close the other editors and open the selected action's editor.
- MacroItemComponent has been moved under the macro folder. 
- Actions can have a custom title (editable by clicking the action title) or one generated by changing settings. Custom title, if set, always overrides generated title.
- Mouse and key macro actions have a UI.
- Move and scroll macro actions can be updated.
- Delay macro action now has a few useful presets for quickly setting the delay. The UI looked too empty without them. :)
- Macro actions can now be arranged by dragging thanks to the delightfully named Dragula component. :F
- Some texts in macro action titles have been changed to better represent the action or for readability (e.g. "x pixels right and y pixels downward" are now just "x pixels right and y pixels down")
- Many of fjozsef's recommendations have been applied. Thanks!

### Still to be done
- Saving edited macro and its actions somewhere. I'll probably leave this to you.
- Adding a new macro. Where should I add a method for adding a new macro to the data loaded from UhkConfigurationService?
- Changing type of macro action (e.g. from text to delay or from holdkey to presskey). I think the type should be changed when saving the edited action as that's really the only place where it matters. But how to do it is where I could use some help as you know the MacroAction stuff better.
- Saving selected keys or mouse buttons to the MacroAction. They seem to have a scancode or what I think is a bitmask field but how should I populate these?
- Confirmation dialog when deleting macro action (is this needed?)

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (changes[this.lastViewModel]) {
+      this.lastViewModel = this.model;
+      this.refreshView();
+    }
+  }
+
+  onKeypress(event: any) {
+    if (this.updateOnEnter && (event.which === KEY_ENTER)) {
+      // Finish editing when Enter pressed
+      this.elRef.nativeElement.blur();
```
fjozsef: Use renderer's `invokeElementMethod`.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    @ViewChild('tab') selectedTab: Tab;
+
+    public isEnabled: boolean; // Can be controlled from MacroComponent via local variable interaction (#macroPopover)
```
fjozsef: Rename to `enabled`.

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,73 @@
...
+        require('./macro-key.component.scss')
+    ],
+    host: { 'class': 'macro__mouse' }
+})
+export class MacroKeyTabComponent implements OnInit {
+    @Input() macroAction: MacroAction;
+    @ViewChild('tab') selectedTab: Tab;
+
+    private activeTab: TabName;
+    private tabName = TabName;
```
fjozsef: Change it to:
```
    /* tslint:disable:variable-name: It is an enum type. So it can start with uppercase. */
    /* tslint:disable:no-unused-variable: It is used in the template. */
    private TabName = TabName;
    /* tslint:enable:no-unused-variable tslint:enable:variable-name */
```

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,95 @@
...
+        require('./macro-mouse.component.scss')
+    ],
+    host: { 'class': 'macro__mouse' }
+})
+export class MacroMouseTabComponent implements OnInit {
+    @Input() macroAction: MacroAction;
+    @ViewChild('tab') selectedTab: Tab;
+
+    private activeTab: TabName;
+    private tabName = TabName;
```
fjozsef: Change it to:
```
/* tslint:disable:variable-name: It is an enum type. So it can start with uppercase. */
/* tslint:disable:no-unused-variable: It is used in the template. */
private TabName = TabName;
/* tslint:enable:no-unused-variable tslint:enable:variable-name */
```

src/components/macro/macro-item/macro-item.component.ts
```diff
@@ -0,0 +1,211 @@
...
+    @Output() cancel = new EventEmitter<any>();
+    @Output() edit = new EventEmitter<any>();
+    @Output() delete = new EventEmitter<any>();
+
+    @ViewChild('macroActionEditor') actionEditor: MacroActionEditorComponent;
+
+    private iconName: string;
+    private title: string;
+    private customTitle: string;
+    private isEditing: boolean = false;
```
fjozsef: Rename to `editing`.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,109 @@
...
+            this.macro = macros[id];
+       });
+    }
+
+    ngAfterViewInit() {
+       this.macroItemsSub = this.macroItems.changes.subscribe((data: any) => {
+           if (this.addedNewAction) {
+               // Open editor for newly added action
+               // Rather cludge way to do this, basically macroItems have to be updated before the editor can be opened
+               setTimeout(() => {
```
fjozsef: I think should be done otherway, maybe I will check that later.

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
laxu: I think the TabName written as tabName variable is a bit more understandable to someone looking at the code the first time.

fjozsef: We should be consistent. Enums should start with uppercase, however using enums directly in the template is not possible, since we can only use members, so we should hack it this way. I think the comment gives explanation for those, who see the code at the first time.

laxu: Fair enough. I'll change them.

mondalaci:
Looks like you've been making great progress, @laxu! I'm very impressed. My comments:

> Actions can have a custom title (editable by clicking the action title) or one generated by changing settings. Custom title, if set, always overrides generated title.

Adding custom descriptions to macro actions probably doesn't make a lot of sense because

1. macro actions are very simplistic in nature, so custom descriptions only add redundancy over the generated descriptions,
2. custom descriptions would occupy additional space in the EEPROM for no good reason,
3. it'd be more useful to trigger the editing of the macro action itself upon clicking its description text.

> Saving selected keys or mouse buttons to the MacroAction. They seem to have a scancode or what I think is a bitmask field but how should I populate these?

By "keys" do you mean modifiers? We should use a bitmask field the same way as in the key action popover. We should also use a bitmask field for mouse buttons. Left button should be the least significant bit, and then middle button, and right button.

> Confirmation dialog when deleting macro action (is this needed?)

Instead asking for confirmations, our UX policy is to offer undo wherever possible, as God intended. :)

@fjozsef Would you please help @laxu find Arpi's undo notification component? I cannot trigger it via the UI.

> Changing type of macro action (e.g. from text to delay or from holdkey to presskey). I think the type should be changed when saving the edited action as that's really the only place where it matters. But how to do it is where I could use some help as you know the MacroAction stuff better.

I do agree that the macro action's type and parameters should only be saved upon actually saving the action.

As for the how, would you please answer, @fjozsef? And also, would you answer @laxu's following question?

> Adding a new macro. Where should I add a method for adding a new macro to the data loaded from UhkConfigurationService?

Thanks for the great work!

NejcZdovc:
If I may answer to two questions.

1) Notification component is already in the agent, but it's hidden for now. This is something that I will implement when data layer is complete and will be automatically triggered when user do an action. So it will be universal and developer don't need to implement anything additional to his component. It will work out of the box.

2) For now don't implement save action until data layer #95 is complete.

laxu:
Thanks for the replies. I did the custom title because I swear I saw somewhere that it was something you wanted. I may have just been out of my mind as I didn't find it again. In any case it's easy to remove if that is the case. I think the only use for it would be adding a name like "Switch to next tab" which might not be obvious just from the keys pressed.

By keys I meant the hold/press/release for keys (including modifiers) as well as mouse buttons. I saw that the modifiers had their own macro actions but are these even needed considering the keypress tab component lets you select a scancode and x amount of modifiers as well as a long press action? Which method in the keypress tab component should I use to get the data needed to populate the macro action?

For undo I guess I need to implement some sort of simply undo list where removed actions are moved and if you save the macro then those are cleared. Should not be difficult.

NejcZdovc:
@laxu like I said undo will be done automatically, so you don't need to implement anything.

mondalaci:
The macro actions are certainly nontrivial. For example, `PressKeyMacroAction` and `PressModifiersMacroAction` should seemingly be implemented by one class. The reason it's implemented by multiple classes is that this way we can serialize the binary configuration in a more compact manner. But this could be better implemented.

Let me propose a refactor in this respect.

@fjozsef How about merging `{Hold,Press,Release}{Key,Modifiers}MacroAction.ts` to `KeyMacroAction.ts`? Then we'd add an `action` field for Hold|Press|Release and a `fields` field for Key|Modifiers|KeyAndModifiers. We could use 2 bits of the macro action type byte to store `action`, and another 2 to store `fields`. The length of the action item would be dependent on the value of `fields`. This space optimized, single-class strategy is the same that we're already using for key actions.

We could do the same kind of merging for `{Hold,Press,Release}MouseButtonsMacroAction.ts` to `MouseButtonsMacroAction.ts`.

@laxu For the time being you're welcome to create `KeyMacroAction.ts` without any of the above logic. It should store scancode and modifiers. Long press action doesn't make sense for macros.

Regarding technical questions, I'm waiting for @NejcZdovc or @fjozsef to reply. :)

laxu:
I added KeyMacroAction and MouseButtonMacroAction classes and refactored the MacroAction stuff quite a bit. I need you to check if any of it makes any sense or if I'm just doing bad choices there. It now checks the macroActionType to figure out if it's press/hold/release key but it could probably be done much better.

Now setting the key scancodes and modifiers works but keypress and modifier actions are still a bit separate and I should combine them so they show up as stuff like "Press/Hold/Release key left ctrl + tab" etc as that's really what the UI represents.

I could use some code review for setting and loading the bitmasks for mouse buttons. I have very little experience working with bitmasks so my solutions might be totally wrong. Should it be even possible to select multiple mouse buttons to press, hold or release in a macro?

fjozsef:
Sorry, I was busy. I will check this out today and going to reply asap.

laxu:
Still has issues where changing the type of macro action would be needed, have to rethink those a bit.

fjozsef:
I think to much refactoring is in this pr. I am sure creating the macro page wouldn't need to affect this many file changes. It would be better if you could seperate these refactors and send an individual pr for each or at least create different commits which would only contain the refactored code. 
For now it is really hard to see clearly what is going on (102 commit, 59 files changes).

fjozsef:
> Adding a new macro. Where should I add a method for adding a new macro to the data loaded from UhkConfigurationService?

Don't add any method now. Just use `this.getUhkConfiguration().macros.elements.push`, it is enough while #95 is not completed.

> Saving selected keys or mouse buttons to the MacroAction. They seem to have a scancode or what I think is a bitmask field but how should I populate these?

You can see an example in the popover/keypress-tab how to get the scancode. About bitmask fields:
```typescript
enum MouseButtons {
    Left = 1 << 0,
    Middle = 1 << 1,
    Right = 1 << 2
};

let x: number = 0;
x |= MouseButtons.Left; // Setting Left true: 000 | 001 = 001
x |= MouseButtons.Right; // Setting right true: 001 | 100 = 101
x &= ~MouseButtons.Left; // Setting left false: 101 & ~(001) = 101 & 110 = 100
```

About the rest of the questions I am a little bit lost what have not been answered, yet. Could you create a summary?

laxu:
The pull requests you need to look at for the MacroAction refactoring are the following:
https://github.com/UltimateHackingKeyboard/agent/pull/87/commits/3debd46b4c4f0a0295637b6919f6d83554d84e7a
https://github.com/UltimateHackingKeyboard/agent/pull/87/commits/f6045c12ea554e0552949160fb26b3e4af310e5f
https://github.com/UltimateHackingKeyboard/agent/pull/87/commits/446a0bfbaec74de1745aa7329fec385f0104318c

The two new classes combine together all the press/hold/release macro actions for keys and mouse buttons. Checking which action is used is done via the macroActionType property and this is changed when you click on different tabs when editing keys or mouse buttons. When your edited action is saved, a new instance of the right class is created based on the macroActionType set and it gets the correct values from the macro action used in editing. 

Now that I think of it this might be possible to simplify by having some sort of generic class that handles all the stuff needed for editing (essentially contains all the properties needed as well as the type) and pops out the correct class instance when saved. I'll have another look at it and see if it can be done like that.

In any case to me it seems there are barely any difference between the various press/hold/release classes for both keys and modifiers so I believe they could be easily combined, thus the KeyMacroAction class. If the keyboard only allows holding a key OR a bunch of modifiers then that will have to be taken into account in UI.

mondalaci:
> @laxu: Should it be even possible to select multiple mouse buttons to press, hold or release in a macro?

Yes, it should be possible, because it can save EEPROM space when multiple buttons are used.

> @fjozsef: I think too much refactoring is in this pr.

I tend to agree. I think we shouldn't aim for perfection for the first try. What we have is already a quantum leap forward compared to what we had, so I'd love to see it merged if it doesn't break anything. Having smaller PRs to review should improve the responsiveness and quality of the review process.

By the way, just tried to build this branch and after resolvig the conflicts I got error messages like the following:
```
ERROR in /home/laci/projects/uhk/agent/node_modules/@types/es6-shim/index.d.ts
(6,14): error TS2300: Duplicate identifier 'PropertyKey'.

ERROR in ./src/components/macro/macro.component.ts
Module not found: Error: Cannot resolve 'file' or 'directory' ../../config-serializer/config-items/TextMacroAction in /home/laci/projects/uhk/agent/src/components/macro
 @ ./src/components/macro/macro.component.ts 15:24-87
```

I suspected that the typings directory is the culprit, but after removing and reinstalling node_modules it's still not working.

Relying on your ultimate wisdom.

laxu:
I did some more work last night and made an EditableMacroAction class that simplifies the handling of the different types a lot. When you start editing a macro action, an EditableMacroAction is created containing the data from whatever macro action type it was originally. When you save the edited action it gets converted back to the right class based on what macroActionType is selected in the class. The macroActionType changes when you click on the tabs. So I might end up reverting some of the changes I did to the MacroAction classes and its subclasses.

I do still think having just a single class handling both mousebuttons and keypresses is a good idea as the only difference between for example ReleaseModifiersMacroAction and HoldModifiersMacroAction is that they have a different macroActionType.

@mondalaci probably a merge issue there.

fjozsef:
> @mondalaci:  I think we shouldn't aim for perfection for the first try. What we have is already a quantum leap forward compared to what we had, so I'd love to see it merged if it doesn't break anything. Having smaller PRs to review should improve the responsiveness and quality of the review process.

I agree. @laxu Please, let those smaller PRs come.

mondalaci:
> @laxu: Probably a merge issue there.

It's very possible. I'll take another look as soon as master gets merged to this branch.

> @fjozsef: I agree. @laxu Please, let those smaller PRs come.

When do you plan to merge this PR? Do you see any major problems with it that have yet to be fixed?

laxu:
I'd still like to have the mouse button handling fixed and the EditableMacroAction working perfectly, hopefully I have time to work on it today or tomorrow. I think then it should be good enough to merge. At this stage it's probably not worth adding additional pull requests but would be better to keep it to smaller PRs in the future.

fjozsef:
@laxu I've just talked with @mondalaci about it. Add those minimal changes you think it's necessary to be ready to merge. In the meantime I try to review it and if it not breaks anything then it can be merged. Afterwards I suggest to update #73 so we'll know what should be added after this lands in master. Those changes should come in smaller PRs. Is that okay for your? If not, then please suggest what should we do differently.

laxu:
Ok, now I believe it should work good enough to merge. I'll update #73 with info on what still needs to be done.

laxu:
Ok, now I believe it should work good enough to merge. I'll update #73 with info on what still needs to be done.

mondalaci:
@laxu Thanks for the update! Would you please merge master to this branch? Can't wait to give it a spin!

laxu:
Merged master but it seems the travis build fails because the testserializer doesn't understand the new KeyMacroAction and MouseButtonMacroAction classes.

fjozsef:
How hard would be revert those changes? I would rather add the KeyMacroAction and MouseButtonMacroAction in a different commit/pr.

laxu:
@fjozsef Would be pretty annoying to do as all the key and mouse button handling relies on those classes. I believe it would just be work that would have to be done again anyway.

mondalaci:
How about fixing the tests and merging this PR, guys?

fjozsef:
I'm not satisfied with the current KeyMacroAction. Give me some time and I'll send a pull request so you could see what would I do differently.

src/config-serializer/config-items/macro-action/KeyMacroAction.ts
```diff
@@ -0,0 +1,106 @@
...
+        return {
+            macroActionType: this.macroActionType,
+            scancode: this.scancode,
+            modifierMask: this.modifierMask
+        };
+    }
+
+    _toBinary(buffer: UhkBuffer) {
+        buffer.writeUInt8(this.getActionId());
+        const value = this.isKeyAction() ? this.scancode : this.modifierMask;
```
fjozsef: https://github.com/UltimateHackingKeyboard/agent/issues/98#issuecomment-244630409

src/components/macro/macro-item/macro-item.component.ts
```diff
@@ -0,0 +1,208 @@
...
+            actionType === macroActionType.ReleaseKeyMacroAction ||
+            actionType === macroActionType.ReleaseModifiersMacroAction
+        ) {
+            // Release key
+            this.iconName = 'hand-paper';
+            this.title = 'Release key: ';
+        }
+
+        if (action.scancode) {
+            const scancode = _find(scancodes, ['id', action.scancode.toString()]);
```
fjozsef: Why do we need an additional library to solve something simple as find? [Since ES6](http://exploringjs.com/es6/ch_arrays.html), the array has a find method. Please use that instead.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+               });
+           }
+       });
+    }
+
+    getMacro(id: number) {
+        const allMacros: Macro[] = this.uhkConfigurationService.getUhkConfiguration().macros.elements;
+        const macro = allMacros[id];
+        if (macro) {
+            return _cloneDeep(macro);
```
fjozsef: I don't think we should copy the macro, you can just return the same reference. (Otherwise if would need cloning in the future, then I would suggest to use copy constructors for class types.)

tslint.json
```diff
@@ -66,12 +66,7 @@
             "check-open-brace",
             "check-whitespace"
         ],
-        "ordered-imports": [
-            true,
-            {
-                "named-imports-order": "lowercase-last"
-            }
-        ],
+        "ordered-imports": false,
```
fjozsef: Let's not change this please. I want to keep the imports in order. I was talking about @NejcZdovc that current order is maybe not the best, but it is still better for me than random. If you want us to change it, open an issue and we can discuss the options there.

src/components/popover/widgets/icon/icon.component.scss
```diff
@@ -2,3 +2,21 @@
     display: flex;
     align-items: center;
 }
+
+.action {
```
fjozsef: Where are these styles used? Looking at the html they don't seem to match any element there.

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
```diff
@@ -0,0 +1,205 @@
...
+import {KeyAction, keyActionType} from '../key-action/KeyAction';
+import {KeystrokeAction} from '../key-action/KeystrokeAction';
+import {KeyMacroAction} from './KeyMacroAction';
+import {DelayMacroAction} from './DelayMacroAction';
+import {MouseButtonMacroAction} from './MouseButtonMacroAction';
+import {MoveMouseMacroAction} from './MoveMouseMacroAction';
+import {ScrollMouseMacroAction} from './ScrollMouseMacroAction';
+import {TextMacroAction} from './TextMacroAction';
+
+const macroActionTypeValues = _values(macroActionType);
```
fjozsef: [This](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Object/values) should work instead.

package.json
```diff
@@ -10,6 +10,7 @@
   "devDependencies": {
     "@types/core-js": "^0.9.32",
     "@types/jquery": "1.10.31",
+    "@types/lodash": "^4.14.34",
```
fjozsef: If you see my other comments we don't need the lodash or @types/lodash.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
fjozsef: Still waiting on for the change :)

src/components/macro/macro-item/macro-item.component.ts
```diff
@@ -0,0 +1,208 @@
...
+    host: { 'class': 'macro-item' }
+})
+export class MacroItemComponent implements OnInit, OnChanges {
+
+    @Input() macroAction: MacroAction;
+    @Input() editable: boolean;
+    @Input() deletable: boolean;
+    @Input() moveable: boolean;
+
+    @Output() save = new EventEmitter<any>();
```
fjozsef: They don't seem to emit any value. If that's the case then use `EventEmitter<void>` type.

src/components/macro/macro-item/macro-item.component.ts
laxu: I added it mostly for the cloneDeep and I imagined many of its other methods would be useful in the future. Good point about the find method though, was not aware of that.

src/components/macro/macro.component.ts
laxu: The reason it is copied is so that any changes to it can be easily discarded by just reloading the macro with its ID.

tslint.json
laxu: I disabled because tslint was giving me a really hard time about the ordering of things which to me looked like they were perfectly fine and in alphabetic order. It seemed a bit irrelevant to spend my time on arranging them as I was hoping to get my changes out.

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
laxu: Provided that nobody tries to use it on Safari (or the UI view version) or there is a polyfill for it. I used lodash because it avoids a fair bit of this kind of checking if an ES6 feature is implemented across browsers.

src/components/macro/macro-item/macro-item.component.ts
laxu: I left them as is mainly for future changes where a value might be needed.

fjozsef: Then it could be changed to any, but now is should be void.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+                   newAction.title = 'New macro action';
+                   newAction.actionEditor.toggleEnabled(true);
+                   this.hideOtherActionEditors(data.length - 1);
+                   this.addedNewAction = false;
+               });
+           }
+       });
+    }
+
+    getMacro(id: number) {
```
fjozsef: Looking it more closely I don't think that this helper function should be implemented before #95. So please remove it.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+                return handle.className.indexOf('action--movable') !== -1;
+            }
+        });
+        /* tslint:enable:no-unused-variable */
+    }
+
+    ngOnInit() {
+        this.sub = this.route.params.subscribe((params: any) => {
+            const id: number = params.id;
+            this.macro = this.getMacro(id);
```
fjozsef: Please use `this.uhkConfigurationService.getUhkConfiguration().macros.elements.find(id)` instead of getMacro.

src/components/macro/macro.component.ts
laxu: This has no bearing on #95 really. It only operates on the MacroItemComponent elements on the page. Basically this is needed for opening the editor for a new macro item. So when you click Add new action, a new MacroAction is added to the array of MacroActions, then this causes a new MacroItemComponent and the MacroActionEditor nested within it to be created. To open the editor after adding a new action, we need to wait for Angular to create the MacroItemComponents to the DOM. The setTimeout is to avoid the additional check done in development mode.

fjozsef:
@laxu Please see #99 
MouseButtonMacroAction could be created similarly.

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
fjozsef: We use [core-js](https://github.com/zloirock/core-js), so it should work.

laxu:
Looks good. Please merge that first and then I'll pull it into my branch and make the MouseButtonMacroAction based on that.

fjozsef:
Done

laxu:
Whee! It passes! You probably want to re-enable the ordered imports rule for tslint after this is merged and fix the order.

fjozsef:
Sorry, but I nor want to disable or reenable it. It is enabled for a reason, so please fix those linter errors.

laxu:
Done.

src/app.module.ts
```diff
@@ -36,17 +43,16 @@ import {
 } from './components/svg/keys';
 import { SvgModuleComponent } from './components/svg/module';
 import { SvgKeyboardWrapComponent } from './components/svg/wrap';
-
+import { MainAppComponent, appRoutingProviders, routing }  from './main-app';
 import { DataProviderService } from './services/data-provider.service';
```
fjozsef: Add a new line before services.

src/app.module.ts
```diff
@@ -36,17 +43,16 @@ import {
 } from './components/svg/keys';
 import { SvgModuleComponent } from './components/svg/module';
 import { SvgKeyboardWrapComponent } from './components/svg/wrap';
-
+import { MainAppComponent, appRoutingProviders, routing }  from './main-app';
```
fjozsef: Add a new line before services.

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
...
+export class ContenteditableModel implements OnChanges {
+  @Input('contenteditableModel') model: any;
+  @Input('contenteditableUpdateOnEnter') updateOnEnter: boolean;
+  @Output('contenteditableModelChange') update = new EventEmitter();
+
+  private lastViewModel: any;
+
+  constructor(private elRef: ElementRef, private renderer: Renderer) {}
+
+  ngOnChanges(changes: any) {
```
fjozsef: Change `any` to `SimpleChanges`.

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+@Directive({
+  selector: '[contenteditableModel]',
+  host: {
+    '(blur)': 'onBlur()',
+    '(keypress)': 'onKeypress($event)'
+  }
+})
+export class ContenteditableModel implements OnChanges {
+  @Input('contenteditableModel') model: any;
```
fjozsef: Use/create the proper type instead of `any`.

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  @Input('contenteditableUpdateOnEnter') updateOnEnter: boolean;
+  @Output('contenteditableModelChange') update = new EventEmitter();
+
+  private lastViewModel: any;
```
fjozsef: Same here, see comment above.

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(private elRef: ElementRef, private renderer: Renderer) {}
+
+  ngOnChanges(changes: any) {
+    if (changes[this.lastViewModel]) {
+      this.lastViewModel = this.model;
+      this.refreshView();
+    }
+  }
+
+  onKeypress(event: any) {
```
fjozsef: Use/create the proper type instead of `any`.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    styles: [ require('./macro-action-editor.component.scss') ],
+    host: { 'class': 'macro-action-editor' }
+})
+export class MacroActionEditorComponent implements OnInit {
+    @Input() macroAction: MacroAction;
+
+    @Output() save = new EventEmitter<MacroAction>();
+    @Output() cancel = new EventEmitter<void>();
+
+    @ViewChild('tab') selectedTab: any;
```
fjozsef: Use/create the proper type instead of `any`.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    public enabled: boolean; // Can be controlled from MacroComponent via local variable interaction (#macroPopover)
+    private editableMacroAction: EditableMacroAction;
+    private activeTab: TabName;
+    /* tslint:disable:variable-name: It is an enum type. So it can start with uppercase. */
+    /* tslint:disable:no-unused-variable: It is used in the template. */
+    private TabName = TabName;
+    /* tslint:enable:no-unused-variable */
+    /* tslint:enable:variable-name */
+
+    constructor() {
```
fjozsef: Remove empty `constructor` and `ngOnInit`.

src/components/macro/macro-action-editor/tab/macro-delay/macro-delay.component.ts
```diff
@@ -0,0 +1,34 @@
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
+import { AfterViewInit, Component, ElementRef, Input, OnInit, Renderer, ViewChild } from '@angular/core';
+
+const INITIAL_DELAY = 0.5; // 0.5 seconds
```
fjozsef: Change comment to `// in seconds`

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,73 @@
+import { KeyAction } from '../../../../../config-serializer/config-items/key-action/KeyAction';
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
+import { MacroSubAction } from '../../../../../config-serializer/config-items/macro-action/MacroAction';
+import { KeypressTabComponent } from '../../../../popover/tab/keypress';
+import { Tab } from '../../../../popover/tab/tab';
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
```
fjozsef: Move imports from angular to the first line and add a new line after it, so it can be in a seperated group.

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    constructor() {
+        this.buttonLabels = ['Left', 'Middle', 'Right'];
+        this.selectedButtons = Array(this.buttonLabels.length).fill(false);
+    }
+
+    ngOnInit() {
+        const tabName = this.getTabName(this.macroAction);
+        this.selectTab(tabName);
+        const buttonActions = [TabName.Click, TabName.Hold, TabName.Release];
+        if (buttonActions.indexOf(this.activeTab) !== -1) {
```
fjozsef: Maybe we could use [Array.prototype.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here.

src/components/macro/macro-action-editor/tab/macro-text/macro-text.component.ts
```diff
@@ -0,0 +1,30 @@
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
+import {
+    AfterViewInit,
+    Component,
+    ElementRef,
+    Input,
+    OnInit,
+    Renderer,
+    ViewChild
+} from '@angular/core';
```
fjozsef: Move angular imports to the first import block (new line with the end).

src/components/macro/macro-action-editor/tab/macro-text/macro-text.component.ts
```diff
@@ -0,0 +1,30 @@
...
+    styles: [require('./macro-text.component.scss')],
+    host: { 'class': 'macro__text' }
+})
+export class MacroTextTabComponent implements OnInit, AfterViewInit {
+    @Input() macroAction: EditableMacroAction;
+    @ViewChild('macroTextInput') input: ElementRef;
+
+    constructor(private renderer: Renderer) {}
+
+    ngOnInit() {}
```
fjozsef: Remove empty `ngOnInit`.

src/components/macro/macro-item/macro-item.component.ts
```diff
@@ -0,0 +1,188 @@
...
+
+import { MouseButtonMacroAction } from '../../../config-serializer/config-items/macro-action/MouseButtonMacroAction';
+import { MoveMouseMacroAction } from '../../../config-serializer/config-items/macro-action/MoveMouseMacroAction';
+import { ScrollMouseMacroAction } from '../../../config-serializer/config-items/macro-action/ScrollMouseMacroAction';
+import { TextMacroAction } from '../../../config-serializer/config-items/macro-action/TextMacroAction';
+
+import { MapperService } from '../../../services/mapper.service';
+import { MacroActionEditorComponent } from '../macro-action-editor/macro-action-editor.component';
+
+import { Component, EventEmitter, Input, OnChanges, OnInit, Output, ViewChild } from '@angular/core';
```
fjozsef: Move angular imports to the first import block (with new line at the end).

src/components/macro/macro-item/macro-item.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    private title: string;
+    private editing: boolean = false;
+
+    constructor(private mapper: MapperService) {}
+
+    ngOnInit() {
+        this.updateView();
+    }
+
+    ngOnChanges(changes: any) {
```
fjozsef: Replace `any` with `SimpleChanges`.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
-import { Component } from '@angular/core';
+import { Macro } from '../../config-serializer/config-items/Macro';
```
fjozsef: Move angular imports to the first block, rxjs import to the second, dragula the third, rest is the fourth.

src/components/popover/tab/mouse/mouse-tab.component.ts
```diff
@@ -1,8 +1,6 @@
-import {Component, Input, OnInit} from '@angular/core';
```
fjozsef: Not modify these.

src/components/side-menu/side-menu.component.html
```diff
@@ -19,7 +19,7 @@
     <li class="sidebar__level-1--item">
         <div class="sidebar__level-1">
             <i class="fa fa-play"></i> Macros
-            <a href="#" class="btn btn-default pull-right btn-sm">
+            <a (click)="addMacro()" class="btn btn-default pull-right btn-sm">
```
fjozsef: Move click after class.

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
```diff
@@ -0,0 +1,198 @@
...
+    moveY: number;
+    scrollX: number;
+    scrollY: number;
+    // Delay macro action properties
+    delay: number;
+    // Text macro action properties
+    text: string;
+
+    assertMacroActionType(jsObject: any) {
+        if (macroActionTypeValues.indexOf(jsObject.macroActionType) === -1) {
```
fjozsef: Use Array's include here, too.

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
```diff
@@ -0,0 +1,198 @@
...
+    moveX: number;
+    moveY: number;
+    scrollX: number;
+    scrollY: number;
+    // Delay macro action properties
+    delay: number;
+    // Text macro action properties
+    text: string;
+
+    assertMacroActionType(jsObject: any) {
```
fjozsef: Is `any` replacable with `JsObjectEditableMacroAction` here ? If yes, then replace it!

src/components/macro/macro-action-editor/tab/macro-delay/macro-delay.component.ts
```diff
@@ -0,0 +1,34 @@
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
+import { AfterViewInit, Component, ElementRef, Input, OnInit, Renderer, ViewChild } from '@angular/core';
```
fjozsef: Move imports from angular to the first line and add a new line after it, so it can be in a seperated group.

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,104 @@
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
+import { MacroSubAction, macroActionType } from '../../../../../config-serializer/config-items/macro-action/MacroAction';
+import { Tab } from '../../../../popover/tab/tab';
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
```
fjozsef: Move imports from angular to the first line and add a new line after it, so it can be in a seperated group.

src/config-serializer/config-items/macro-action/MouseButtonMacroAction.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    toString(): string {
+        return `<MouseButtonMacroAction mouseButtonsMask="${this.mouseButtonsMask}">`;
+    }
+
+    hasButtons(): boolean {
+        return this.mouseButtonsMask !== 0;
+    }
+
+    isHoldAction(): boolean {
```
fjozsef: Rename to `isOnlyHoldAction`.

src/config-serializer/config-items/macro-action/MouseButtonMacroAction.ts
```diff
@@ -0,0 +1,85 @@
...
+        return this.action === MacroSubAction.hold;
+    }
+
+    isPressAction(): boolean {
```
fjozsef: Rename to `isOnlyPressAction`.

src/config-serializer/config-items/macro-action/MouseButtonMacroAction.ts
```diff
@@ -0,0 +1,85 @@
...
+        return this.action === MacroSubAction.press;
+    }
+
+    isReleaseAction(): boolean {
```
fjozsef: Rename to `isOnlyReleaseAction`.

src/components/contenteditable/contenteditable.component.ts
laxu: I believe any should be there as it can be anything from an object to string etc.

fjozsef: Okay, then it can stay.

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
laxu: tslint seemed to regard the .. stuff as preceding the angular stuff. That's why I was against the imports ordering rule because it makes for less logical reading.

src/components/popover/tab/mouse/mouse-tab.component.ts
laxu: tslint has probably given an unused import error here.

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
fjozsef: True, but if we make groups, then tslint test passes. Feel free to create an issue about the import orders as I mentioned before, I admit, that the current ordering is not the best, but I want to avoid random orders.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+        this.sub = this.route.params.subscribe((params: any) => {
+            const id: number = Number(params.id);
+            this.macro = this.getMacro(id);
+            this.dragEnabled = true;
+            this.hasChanges = false;
+       });
+    }
+
+    ngAfterViewInit() {
+       this.macroItemsSub = this.macroItems.changes.subscribe((data: any) => {
```
fjozsef: If my calculations are right, then `data`'s type is ` QueryList<MacroItemComponent>` here. Please, replace it.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+           if (this.addedNewAction) {
+               // Open editor for newly added action
+               // Rather cludge way to do this, basically macroItems have to be updated before the editor can be opened
+               setTimeout(() => {
+                   let newAction = data.last;
+                   newAction.title = 'New macro action';
```
fjozsef: `title` is private member of  `MacroItemComponent`. Please, respect it!

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+
+    constructor(
+        private uhkConfigurationService: UhkConfigurationService,
+        private route: ActivatedRoute,
+        private dragulaService: DragulaService
+    ) {
+        /* tslint:disable:no-unused-variable: Used by Dragula. */
+        dragulaService.setOptions('macroActions', {
+            moves: function (el: any, container: any, handle: any) {
+                return handle.className.indexOf('action--movable') !== -1;
```
fjozsef: Use Array's includes.

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,132 @@
...
+            }
+        });
+        /* tslint:enable:no-unused-variable */
+    }
+
+    ngOnInit() {
+        this.sub = this.route.params.subscribe((params: any) => {
```
fjozsef: Replace `any` with `{id: string}`

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,113 @@
...
+})
+export class MacroActionEditorComponent implements OnInit {
+    @Input() macroAction: MacroAction;
+
+    @Output() save = new EventEmitter<MacroAction>();
+    @Output() cancel = new EventEmitter<void>();
+
+    @ViewChild('tab') selectedTab: any;
+
+    public enabled: boolean; // Can be controlled from MacroComponent via local variable interaction (#macroPopover)
```
fjozsef: A component never should assume its parent!

src/components/macro/macro-action-editor/macro-action-editor.component.ts
laxu: any is fine here because each tab is quite unique, only some of them implement Tab and I didn't want to create a completely redundant class or interface for these as there are no shared properties or methods.

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
+import {Directive, ElementRef, EventEmitter, Input, OnChanges, Output, Renderer, SimpleChanges } from '@angular/core';
```
fjozsef: Move this file to `src/directives/contenteditable/contenteditable.directive.ts`

src/components/contenteditable/contenteditable.component.ts
```diff
@@ -0,0 +1,45 @@
...
+const KEY_ENTER = 13;
+
+@Directive({
+  selector: '[contenteditableModel]',
+  host: {
+    '(blur)': 'onBlur()',
+    '(keypress)': 'onKeypress($event)'
+  }
+})
+export class ContenteditableModel implements OnChanges {
```
fjozsef: Rename class to `ContenteditableDirective`

fjozsef:
I have sent a [PR](https://github.com/laxu/agent/pull/1) about the changes that should be added. It was easier to implement them myself.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    onCancelClick(): void {
+        this.cancel.emit();
+    }
+
+    onSaveClick(): void {
+        try {
+            const action = this.editableMacroAction;
+            if (action.isKeyAction()) {
+                // Could updating the saved keys be done in a better way?
+                const tab = this.selectedTab as MacroKeyTabComponent;
```
fjozsef: The selected tab isn't always `MacroKeyTabComponent`.

laxu: It should work fine because of the isKeyAction check as changing the tab means also changing EditableMacroAction.macroActionType. So it should not be possible to select a different tab and have isKeyAction pass.

The real solution would be to update the EditableMacroAction every time something is changed in the key press component by providing an output event that gets run every time you change something in it. The reason I didn't do this is because I wanted to keep my changes in the macro components as much as possible and not mess with the other stuff.

fjozsef: Okay, now I understand, it is fine.

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,100 @@
...
+            console.error(e);
+        }
+    }
+
+    selectTab(tab: TabName): void {
+        this.activeTab = tab;
+        this.editableMacroAction.macroActionType = this.getTabMacroActionType(tab);
+    }
+
+    getTabName(action: EditableMacroAction) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,100 @@
...
+            case macroActionType.MouseButtonMacroAction:
+            case macroActionType.MoveMouseMacroAction:
+            case macroActionType.ScrollMouseMacroAction:
+                return TabName.Mouse;
+            default:
+                return TabName.Keypress;
+        }
+    }
+
+    getTabMacroActionType(tab: TabName) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/macro-action-editor.component.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (tab === TabName.Delay) {
```
fjozsef: Change these to switch-case.

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.defaultKeyAction = this.macroAction.toKeyAction();
+        this.selectTab(this.getTabName(this.macroAction));
+    }
+
+    selectTab(tab: TabName): void {
+        this.activeTab = tab;
+        this.macroAction.action = this.getActionType(tab);
+    }
+
+    getTabName(action: EditableMacroAction) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!action.action || action.isOnlyPressAction()) {
+            return TabName.Keypress;
+        } else if (action.isOnlyHoldAction()) {
+            return TabName.Hold;
+        } else if (action.isOnlyReleaseAction()) {
+            return TabName.Release;
+        }
+    }
+
+    getActionType(tab: TabName) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (tab === TabName.Keypress) {
```
fjozsef: Use switch-case

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,72 @@
...
+            return MacroSubAction.press;
+        } else if (tab === TabName.Hold) {
+            return MacroSubAction.hold;
+        } else if (tab === TabName.Release) {
+            return MacroSubAction.release;
+        }
+    }
+
+    getKeyAction() {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,105 @@
...
+            this.macroAction.action = this.getAction(tab);
+        }
+    }
+
+    setMouseClick(index: number) {
+        this.selectedButtons[index] = !this.selectedButtons[index];
+        this.macroAction.setMouseButtons(this.selectedButtons);
+    }
+
+    hasButton(index: number) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        return this.selectedButtons[index];
+    }
+
+    getAction(tab: TabName) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        if (tab === TabName.Click) {
```
fjozsef: Use switch-case

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,105 @@
...
+            return MacroSubAction.press;
+        } else if (tab === TabName.Hold) {
+            return MacroSubAction.hold;
+        } else if (tab === TabName.Release) {
+            return MacroSubAction.release;
+        }
+    }
+
+    getTabName(action: EditableMacroAction) {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,105 @@
...
+            }
+        } else if (action.macroActionType === macroActionType.MoveMouseMacroAction) {
+            return TabName.Move;
+        } else if (action.macroActionType === macroActionType.ScrollMouseMacroAction) {
+            return TabName.Scroll;
+        }
+        return TabName.Move;
+    }
+
+    getMacroActionType(tab: TabName) {
```
fjozsef: Add return type

src/components/macro/macro-item/macro-item.component.ts
```diff
@@ -0,0 +1,191 @@
+import { Component, EventEmitter, Input, OnChanges, OnInit, Output, SimpleChanges } from '@angular/core';
+
+import { KeyModifiers } from '../../../config-serializer/config-items/KeyModifiers';
+import { DelayMacroAction } from '../../../config-serializer/config-items/macro-action/DelayMacroAction';
```
fjozsef: Import all `MacroAction` from `../../../config-serializer/config-items/macro-action`.

src/components/macro/macro-action-editor/tab/macro-text/macro-text.component.ts
```diff
@@ -0,0 +1,28 @@
+import {
+    AfterViewInit,
+    Component,
+    ElementRef,
+    Input,
+    Renderer,
+    ViewChild
+} from '@angular/core';
+
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
```
fjozsef: Import from `../../../config-serializer/config-items/macro-action`

src/components/macro/macro.component.ts
```diff
@@ -1,12 +1,116 @@
...
+    ngOnInit() {
+        this.routeSubscription = this.route.params.subscribe((params: { id: string }) => {
+            const id: number = Number(params.id);
+            this.macro = this.getMacro(id);
+            this.dragEnabled = true;
+            this.hasChanges = false;
+        });
+    }
+
+    getMacro(id: number) {
```
fjozsef: Add return type

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
```diff
@@ -0,0 +1,190 @@
+import { KeyAction, keyActionType } from '../key-action/KeyAction';
```
fjozsef: import from `'../key-action'`

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
```diff
@@ -0,0 +1,190 @@
...
+        };
+    }
+
+    fromKeyAction(keyAction: KeyAction) {
+        let data = keyAction.toJsObject();
+        this.scancode = data.scancode;
+        this.modifierMask = data.modifierMask;
+    }
+
+    toKeyAction(): KeystrokeAction {
```
fjozsef: Rename `toKeystrokeAction`

src/config-serializer/config-items/macro-action/EditableMacroAction.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    setMouseButtons(buttonStates: boolean[]) {
+        let bitmask = 0;
+        for (let i = 0; i < buttonStates.length; i++) {
+            bitmask |= Number(buttonStates[i]) << i;
+        }
+        this.mouseButtonsMask = bitmask;
+    }
+
+    getMouseButtons() {
```
fjozsef: Add return type

src/config-serializer/config-items/macro-action/MouseButtonMacroAction.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    setMouseButtons(buttonStates: boolean[]) {
+        let bitmask = 0;
+        for (let i = 0; i < buttonStates.length; i++) {
+            bitmask |= Number(buttonStates[i]) << i;
+        }
+        this.mouseButtonsMask = bitmask;
+    }
+
+    getMouseButtons() {
```
fjozsef: Add return type

src/config-serializer/config-items/macro-action/MouseButtonMacroAction.ts
```diff
@@ -0,0 +1,85 @@
...
+    mouseButtonsMask: number;
+
+    _fromJsObject(jsObject: JsObjectMouseButtonMacroAction): MouseButtonMacroAction {
+        this.assertMacroActionType(jsObject);
+        this.action = MacroSubAction[jsObject.action];
+        this.mouseButtonsMask = jsObject.mouseButtonsMask;
+        return this;
+    }
+
+    _fromBinary(buffer: UhkBuffer): MouseButtonMacroAction {
```
fjozsef: Add return type

src/components/macro/macro-action-editor/tab/macro-delay/macro-delay.component.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
```
fjozsef: Import from `../../../../../config-serializer/config-items/macro-action`

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,72 @@
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
+
+import { KeyAction } from '../../../../../config-serializer/config-items/key-action/KeyAction';
```
fjozsef: Import from `../../../../../config-serializer/config-items/key-action`

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,72 @@
...
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
```
fjozsef: Import from `../../../../../config-serializer/config-items/macro-action`

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,105 @@
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
+
+import { EditableMacroAction } from '../../../../../config-serializer/config-items/macro-action/EditableMacroAction';
```
fjozsef: Import from `../../../../../config-serializer/config-items/macro-action`

fjozsef:
This needs to be fixed:
macro-tab

fjozsef:
The border between two macro-item is still doubled.

src/components/macro/macro-action-editor/tab/macro-key/macro-key.component.ts
```diff
@@ -0,0 +1,74 @@
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
+
+import { KeyAction } from '../../../../../config-serializer/config-items/key-action';
+import { EditableMacroAction, MacroSubAction } from '../../../../../config-serializer/config-items/macro-action';
+import { KeypressTabComponent } from '../../../../popover/tab/keypress';
```
fjozsef: `Tab` and `KeypressTabComponent` should be imported from `../../../../popover/tab/`

src/components/macro/macro-action-editor/tab/macro-mouse/macro-mouse.component.ts
```diff
@@ -0,0 +1,107 @@
+import { Component, Input, OnInit, ViewChild } from '@angular/core';
+
+import { EditableMacroAction, MacroSubAction, macroActionType } from '../../../../../config-serializer/config-items/macro-action';
+import { Tab } from '../../../../popover/tab/tab';
```
fjozsef: import from `../../../../popover/tab`

fjozsef:
The list of macro-items should be scrollable when there are too many of them

fjozsef:
"Add new action item" should be the part of the scrollable area and add some space after it.

laxu:
I disagree about the add new item being part of the scrollable area. This way it is always visible so you don't have to scroll all the way down the list to access it. The list also scrolls down to the editor if you add a new action so I'd consider it to have better usability this way.

I added a few breakpoints so that the size of the scrollable list changes accordingly and leaves some space at the bottom in various window sizes.

fjozsef:
You're right. It is better if it's not the part of it. I just wanted it to be visible when there is a long list.

laxu:
Yeah that was my bad for not checking how it handles on much smaller window size than what I normally use.

fjozsef:
Unused empty space marked with yellow. It should be filled.

empty space

laxu:
I left the space like that because of the fork me on github banner. Otherwise it will overlap the content, possibly hiding the edit button for example.

fjozsef:
It should overflow, the banner will not be there forever. Otherwise if we want it to not overflow, then it should be then on higher level in dom.

laxu:
So you want it to go over the banner? Or allow the macro list to go under the banner?

fjozsef:
Second option, under the banner

src/components/macro/macro.component.scss
```diff
@@ -59,18 +59,22 @@ h1 {
...
     overflow: auto;
-    max-height: 70vh;
+    max-height: 78vh;
     margin-bottom: 0;
     border-radius: 4px 4px 0 0;
     border: 1px solid #ddd;
     border-bottom: 0;
 
-    @media screen and (min-height: 1000px) {
+     @media screen and (min-height: 730px) {
```
fjozsef: These are not working for me, otherwise I think it should be done with flexbox, instead of media query.

laxu:
Tried flexbox but couldn't get it to work as expected in this case. Because the main content area is not scrollable content will go outside the window so I setup the breakpoints using vh values and that seemed to be the only way to keep the content as desired. If you can figure out a better way please make a pull request.

Issue #84, opened 2016-08-27
Select2 UI improvements


mondalaci:
On the Keypress tab of the key action editor:
- the text should be left-aligned
- "Functionkeys" should be written as two separate words
- alternative names like PgDn pagedown should be right-aligned and displayed with gray color

On the Layer tab of the key action editor:
- the filter input should be removed because there are only 2 or 3 options

In general:
- the height of select2 dropdowns should occopy all the available vertical space.

NejcZdovc:
@mondalaci I am not sure what this means:

> the height of select2 dropdowns should occopy all the available vertical space.

Can you explain a little bit please?

mondalaci:
Sure thing!

![ultimate hacking keyboard configurator](https://cloud.githubusercontent.com/assets/300402/18067993/982b64d8-6e3f-11e6-9ba7-d161b7dac076.png)

As you can see right now selects have a fixed maximum height instead of stretching till the bottom of the screen. It'd be nice to use all the available screen space.

NejcZdovc:
Ok, got it. This is not a straight forward thing. To achieve this we would need to add additional JS logic to it, because this can't be controlled with CSS. The problem is that drop down can be in multiple places (not only in popover) and that's why CSS can't help us. We would also need to add listener to resize event, because if user resizes window and drop down is opened we need to resize drop down accordingly.

I don't really think that this is a problem. You have search input which helps you navigate. One thing we can do is that we increase height from 200px to for example 300px.

![image](https://cloud.githubusercontent.com/assets/9574457/18077520/8a3e4dbc-6e86-11e6-8da7-af3cc25927e4.png)

mondalaci:
Thanks for the explanation! I wouldn't have thought that this is so hard to achieve. 300px sounds like a good default.

PR #92, opened 2016-08-29
Select2 improvements

NejcZdovc:
This PR is resolving issue #84.

### DIFF
--- a/src/components/popover/tab/keypress/keypress-tab.component.html
+++ b/src/components/popover/tab/keypress/keypress-tab.component.html
@@ -8,14 +8,20 @@
     <b class="setting-label">Modifiers:</b>
     <div class="btn-toolbar modifiers">
         <div class="btn-group btn-group-sm modifiers__left">
-            <button type="button" class="btn btn-default" *ngFor="let modifier of leftModifiers; let index = index" [class.btn-primary]="leftModifierSelects[index]"
-                (click)="toggleModifier(false, index)">
+            <button type="button" class="btn btn-default"
+                    *ngFor="let modifier of leftModifiers; let index = index"
+                    [class.btn-primary]="leftModifierSelects[index]"
+                    (click)="toggleModifier(false, index)"
+            >
                 {{modifier}}
             </button>
         </div>
         <div class="btn-group btn-group-sm modifiers__right">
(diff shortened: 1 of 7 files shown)


NejcZdovc:
@mondalaci please check it out

mondalaci:
Looks like a million dollars, @NejcZdovc! There are some issues I've noticed:

* Alternate names like "PgDn" are not filtered.
* NumberPad -> Numpad
* MediaKeys -> Media Keys
* Spacebar -> Space
* Enter should be primary on the left side, and Return should be secondary on the right side instead of the other way around

NejcZdovc:
All text mistakes are now fixed. Regarding filter there is sadly again no easy solution. We would need to add custom matching system You can read more about [here](https://select2.github.io/announcements-4.0.html#new-matcher). I can add it, but would need to add this functionality to wrapper. Would this feature be used only on this select2 for scan codes?

mondalaci:
Thanks for updating the texts!

It'd be worth to add a custom matching callback, so that users could search for alternate names, too.

For the time being, we'd only use this feature for the scancode select2. My crystal ball is broken, so I'm not sure whether we'll use it for other select2 instances, too. :)

NejcZdovc:
@mondalaci Ok, will also add this to the wrapper, but it will need to wait until more critical issues / tasks are done.

NejcZdovc:
I split remaining task to to new task so we can close this PR.

fjozsef:
left alignment
I think the left alignment should only be present on the Keypress tab and we should keep the default center alignment for the rest select2.
What do you think, @mondalaci ?

NejcZdovc:
@fjozsef I agree with you. I prefer center align

mondalaci:
Left align may look strange but I think it's because the width of the select2 widgets is overly large compared to the text content. How about reducing the width of select2 widgets to make them just slightly wider than the text content? Of course, we should maximize the width of select2 to fit within the popver.

mondalaci:
I'd go as far as left-aligning the content of every select2 widgets. We could fully left align categories and leave some left padding for the items inside of the categories, thereby emphasizing hierarchy beyond bold/normal font weight. I'm not fully confident that this will work, I'm just saying that we should give it a try.

mondalaci:
@NejcZdovc You're welcome to attach a couple of relevant screenshots. Maybe I'll come to the same conclusion.

NejcZdovc:
@fjozsef we will stick with a left alignment and 100% width.

PR #97, opened 2016-09-04
Side menu animation

NejcZdovc:
Improved animation of the side menu. Before we were limited to 200px height. Now animation is handled by angular, so there is no more limits. Also class toggling is now done with Renderer as dictated in style guides.

### DIFF
--- a/src/components/side-menu/side-menu.component.html
+++ b/src/components/side-menu/side-menu.component.html
@@ -5,9 +5,9 @@
             <a [routerLink]="['/keymap/add']" class="btn btn-default pull-right btn-sm">
                 <i class="fa fa-plus"></i>
             </a>
-            <i class="fa fa-chevron-up pull-right" (click)="toggleHide($event, keymapElement)"></i>
+            <i class="fa fa-chevron-up pull-right" (click)="toggleHide($event, 'keymap')"></i>
         </div>
-        <ul #keymapElement>
+        <ul [@toggler]="animation.keymap">
             <li *ngFor="let keymap of keymaps" class="sidebar__level-2--item">
                 <div class="sidebar__level-2">
                     <a [routerLink]="['/keymap', keymap.id]">{{keymap.name}}</a>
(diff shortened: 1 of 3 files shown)


NejcZdovc:
@fjozsef can you please check it out and merge it to the master?

src/components/side-menu/side-menu.component.ts
```diff
@@ -1,33 +1,56 @@
...
 
     ngOnInit() {
         this.keymaps = this.uhkConfigurationService.getUhkConfiguration().keymaps.elements;
         this.macros = this.uhkConfigurationService.getUhkConfiguration().macros.elements;
     }
 
-    /* tslint:disable:no-unused-variable: This function is used in the template */
-    private toggleHide(event: Event, view: Element) {
-        /* tslint:enable:no-unused-variable */
+    toggleHide(event: Event, type: string) {
```
fjozsef: Why have this become public?

src/components/side-menu/side-menu.component.html
```diff
@@ -5,9 +5,9 @@
             <a [routerLink]="['/keymap/add']" class="btn btn-default pull-right btn-sm">
                 <i class="fa fa-plus"></i>
             </a>
-            <i class="fa fa-chevron-up pull-right" (click)="toggleHide($event, keymapElement)"></i>
+            <i class="fa fa-chevron-up pull-right" (click)="toggleHide($event, 'keymap')"></i>
```
fjozsef: You could pass `$event.target` here. See my comment at `toogleHide`.

src/components/side-menu/side-menu.component.ts
NejcZdovc: In all examples on official [angular pages](https://angular.io/resources/live-examples/ngmodule/ts/plnkr.html) they use public for functions that are used in templates. Also this way you don't have a problem with tslint. 

I used only event, because if I would use event.target I would need to write it down in three different places in template (and if we add more menus, this number will be bigger). This way I can just write event.target in only one place.

src/components/side-menu/side-menu.component.html
NejcZdovc: Answered in previous comment.

src/components/side-menu/side-menu.component.ts
fjozsef: Okay

PR #104, opened 2016-09-20
Improve macro sorting visuals

### LINKED ISSUE #103: Make the drag & drop of macro actions more intuitive
Currently, it's rather hard to see which macro action is being dragged. The legacy macro UI is a pretty good example of how it could be implemented in a more intuitive manner. Any way make the new UI behave like the legacy UI in this respect?

laxu:
Fixes #103


NejcZdovc:
@laxu this will not be ok. You need to update your master to the latest version and then create branch from this updated master. Here should be only commits that are related to this PR. Now you have everything from macros PR that is already in agent master.

fjozsef:
@laxu @NejcZdovc is right, however you could have just force push your remote branch, so this pull request's commits would have accordingly.

PR #107, opened 2016-09-20
Allow sorting while editing macro action

### LINKED ISSUE #106: Tweak the UX of drag handles when editing macro actions
Apparently, @laxu is on the roll big time, so I keep these coming! :)

laxu:
Fixes #106. Seems to work just fine like that.


mondalaci:
Works like a charm!

Issue #110, opened 2016-09-20
Change the highlight color of the keymap clone icon


mondalaci:
The red highlight color is only supposed to be used for the delete operation (trash can icon). For other icons let's use the same blue highlight color as used for the (yet to be removed) pen icon of macro actions.

NejcZdovc:
Ok, will be changed.

PR #112, opened 2016-09-20
Fixed color

NejcZdovc:
Fixing #110

### DIFF
--- a/src/components/keymap/header/keymap-header.component.scss
+++ b/src/components/keymap/header/keymap-header.component.scss
@@ -1,3 +1,5 @@
+@import '../../../main-app/global-styles';
+
 .keymap {
     &__is-default {
 
@@ -28,7 +30,7 @@
 
         &:hover {
             cursor: pointer;
-            color: #337ab7;
+            color: $icon-hover;
         }
     }
 }
--- a/src/components/keymap/header/keymap-header.component.scss
+++ b/src/components/keymap/header/keymap-header.component.scss
@@ -28,7 +28,7 @@
 
         &:hover {
             cursor: pointer;
-            color: #900;
+            color: #337ab7;
         }
     }
 }
--- a/src/components/macro/macro-action-editor/macro-action-editor.component.scss
+++ b/src/components/macro/macro-action-editor/macro-action-editor.component.scss
@@ -1,3 +1,5 @@
+@import '../../../main-app/global-styles';
+
 .action--editor {
     padding-top: 0;
     padding-bottom: 0;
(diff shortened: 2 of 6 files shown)


NejcZdovc:
@fjozsef Please review

mondalaci:
Looks good, but how about using a global color variable for this purpose?

NejcZdovc:
I am not aware of a way to share global variables between different components. @fjozsef do you know how to achieve this?

NejcZdovc:
How about if we create scss file with all definitions and then import it to the component scss?

fjozsef:
In scss you also can [import](http://vanseodesign.com/css/sass-the-import-directive/) other (scss) files, so yould define the color in a global.scss and import it.

NejcZdovc:
@mondalaci @fjozsef added variable and use it all around the app that blue color was used

mondalaci:
I applaud the introduction of variables.scss but how about using a more descriptive variable name like `$normal-icon-highlight-color`? And then `$delete-icon-highlight-color` could also be introduced for the trash icon.

Also how about using `global-styles.scss`. It might be a more descriptive filename.

NejcZdovc:
I am not a fan of so descriptive naming convention, because if we use this blue color for let's say link or title hover then you have variable for title named `...-icon-highlight-...`. I prefer naming convention like this:

- blue
- blue-dark
- blue-light
- blue-aero
- etc.

This way you can use color blue on multiple places. In my experience you have max 4 different shades of the same color in the project, otherwise you are doing something wrong :smile:

mondalaci:
I see what you mean but it would be better to name variables in a semantic fashion for supreme maintainability. Let's say we want to change icon highlight color from blue to green. Then we'd have to rename loads of occurences versus one. That's why it'd be wise to use a semantic naming convention that is not related to the value of the variables.

NejcZdovc:
Ok I see where are you coming from. How about if we merge both things in something like this:

### Solution 1
Icon -> `$icon`
Duplicate/Edit hover -> `$hover-icon`
Delete hover -> `$hover-icon-strong` (or something like this)
Link hover  -> `$hover-link`
Link -> $link

### Solution 2
Duplicate/Edit hover -> `$icon-hover`
Delete hover -> `$icon-hover-strong` (or something like this)
Link hover  -> `$link-hover`
Link -> `$link`

What do you think?

mondalaci:
I vote for option 2. I'd rather use `$icon-hover-delete` instead of `$icon-hover-strong`. In my mind the more precisely we describe something the better it is.

NejcZdovc:
Variable is now `$icon-hover`.

NejcZdovc:
@fjozsef rebased to the master, please review

PR #116, opened 2016-09-21
Some refactor

NejcZdovc:
- [x] Unification of folder structure
- [x] Unification of header based on key map


src/components/keymap/header/keymap-header.component.scss
```diff
@@ -7,7 +7,7 @@
             cursor: pointer;
 
             &:hover {
-                color: #b9b6b6;
+                color: $icon-hover;
```
fjozsef: Editing this file should not be the part of macro refactoring.

NejcZdovc: I know, but I don't think that this is something that should have a separate PR.

fjozsef: Then rename the PR's title.

NejcZdovc: I don't know to what

PR #117, opened 2016-09-21
Data layer - Macro & UX improvments

NejcZdovc:
- [x] Edit macro name
- [x] Duplicate macro
- [x] Delete macro
- [x] Add macro action
- [x] Save macro action
- [x] Delete macro action
- [x] Reorder macro action

Because of the problems with data layer logic I needed to do some refactoring (because I was already refactoring I also added some UX improvements) :
- [x] Add new action is now separate element
- [x] Improved close other actions, now only currently opened item is closed
- [x] Icon pencil is hidden when you are editing action
- [x] Title is not clickable when you are editing action
- [x] Title have a cursor pointer now when not in editing mode
- [x] Add new item is hidden when you are adding new action
- [x] Switched positions of pencil and trash icons
- [x] Removed `Remember to save your changes!`
- [x] Move capture keystroke to the basic macro page
- [x] Add hover for whole line when not in edit mode

### DIFF
--- a/src/components/macro/item/macro-item.component.scss
+++ b/src/components/macro/item/macro-item.component.scss
@@ -1,3 +1,5 @@
+@import '../../../main-app/global-styles';
+
 :host {
     &.macro-item:first-of-type {
         .list-group-item {
@@ -43,6 +45,22 @@
         &.is-editing {
             background: #f5f5f5;
         }
+
+        &--wrap {
+            justify-content: space-between;
+
+            &.cursor {
+                &:hover {
+                    cursor: pointer;
+                    color: $icon-hover;
+                }
+            }
+        }
+    }
+
+    &--title {
+        display: flex;
+        flex: 1;
     }
 
     &--movable {
@@ -58,8 +76,4 @@
     border-radius: 0;
     border-left: 0;
     border-right: 0;
-}
-
-.cursor {
-    cursor: pointer;
 }
--- a/src/components/macro/item/macro-item.component.scss
+++ b/src/components/macro/item/macro-item.component.scss
@@ -49,7 +49,7 @@
         &--wrap {
             justify-content: space-between;
 
-            &.cursor {
+            &.pointer {
                 &:hover {
                     cursor: pointer;
                     color: $icon-hover;
--- a/src/components/macro/item/macro-item.component.scss
+++ b/src/components/macro/item/macro-item.component.scss
@@ -18,44 +18,48 @@
             background: #f5f5f5;
         }
     }
+}
 
-    .action {
-        &--item {
+.action {
+    &--item {
+        display: flex;
+        flex-shrink: 0;
+        border: 0;
+        border-bottom: 1px solid #ddd;
+
+        icon {
+            margin: 0 5px;
+        }
+
+        > div {
             display: flex;
-            flex-shrink: 0;
-            border: 0;
-            border-bottom: 1px solid #ddd;
-
-            icon {
-                margin: 0 5px;
-            }
-
-            > div {
-                display: flex;
-                flex: 1;
-            }
-
-            &:first-child {
-                border-radius: 0;
-            }
-
-            &.is-editing {
-                background: #f5f5f5;
-            }
+            flex: 1;
         }
 
-        &--movable {
-            &:hover {
-                cursor: move;
-            }
+        &:first-child {
+            border-radius: 0;
+        }
+
+        &.is-editing {
+            background: #f5f5f5;
         }
     }
 
-    .macro-action-editor__container {
-        padding-top: 0;
-        padding-bottom: 0;
-        border-radius: 0;
-        border-left: 0;
-        border-right: 0;
+    &--movable {
+        &:hover {
+            cursor: move;
+        }
     }
+}
+
+.macro-action-editor__container {
+    padding-top: 0;
+    padding-bottom: 0;
+    border-radius: 0;
+    border-left: 0;
+    border-right: 0;
+}
+
+.cursor {
+    cursor: pointer;
 }
--- a/src/components/macro/item/macro-item.component.ts
+++ b/src/components/macro/item/macro-item.component.ts
@@ -29,7 +29,7 @@ export class MacroItemComponent implements OnInit, OnChanges {
     @Output() save = new EventEmitter<MacroAction>();
     @Output() cancel = new EventEmitter<void>();
     @Output() edit = new EventEmitter<void>();
-    @Output() delete = new EventEmitter<MacroAction>();
+    @Output() delete = new EventEmitter<void>();
 
     private title: string;
     private iconName: string;
@@ -72,8 +72,8 @@ export class MacroItemComponent implements OnInit, OnChanges {
         this.cancel.emit();
     }
 
-    deleteAction(macroAction: MacroAction): void {
-        this.delete.emit(macroAction);
+    deleteAction(): void {
+        this.delete.emit();
     }
 
     private updateView(): void {
--- a/src/components/macro/item/macro-item.component.ts
+++ b/src/components/macro/item/macro-item.component.ts
@@ -29,7 +29,7 @@ export class MacroItemComponent implements OnInit, OnChanges {
     @Output() save = new EventEmitter<MacroAction>();
     @Output() cancel = new EventEmitter<void>();
     @Output() edit = new EventEmitter<void>();
-    @Output() delete = new EventEmitter<void>();
+    @Output() delete = new EventEmitter<MacroAction>();
 
     private title: string;
     private iconName: string;
@@ -53,8 +53,6 @@ export class MacroItemComponent implements OnInit, OnChanges {
     }
 
     saveEditedAction(editedAction: MacroAction): void {
-        // @todo save this to keyboard
-        console.log('Saved action', editedAction);
         this.macroAction = editedAction;
         this.editing = false;
         this.updateView();
@@ -74,8 +72,8 @@ export class MacroItemComponent implements OnInit, OnChanges {
         this.cancel.emit();
     }
 
-    deleteAction(): void {
-        this.delete.emit();
+    deleteAction(macroAction: MacroAction): void {
+        this.delete.emit(macroAction);
     }
 
     private updateView(): void {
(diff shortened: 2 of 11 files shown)


mondalaci:
Please let me share with you how I envision the add macro feature.

* Upon pressing the + macro button a new, empty macro gets created.
* The name of the new macro should be "New macro"
* If default macro name already exists then make it "New macro (2)" or increment the number until a non existing macro name is found.
* Select the default name of the macro with the cursor, so that the user can simply type a default name and hit Enter without any additional interactions.

NejcZdovc:
@mondalaci in this PR there will not be add macro, because it's not implemented yet. So in this PR I will implement data layer for all existing functionalities. In new PR I will implement add macro page and add data layer for it. What do you think?

mondalaci:
@NejcZdovc Ah, gotcha! I'm totally fine with that. I can create a dedicated issue for the add macro action and paste the above if you want to.

NejcZdovc:
@mondalaci Great. Yes please create new issue and describe how add macro page should look like and work.

mondalaci:
@NejcZdovc Just created #118. I simply copy-pasted the above for now. You're welcome to let me know if anything is unclear.

src/components/macro/item/macro-item.component.ts
```diff
@@ -29,7 +29,7 @@ export class MacroItemComponent implements OnInit, OnChanges {
     @Output() save = new EventEmitter<MacroAction>();
     @Output() cancel = new EventEmitter<void>();
     @Output() edit = new EventEmitter<void>();
-    @Output() delete = new EventEmitter<void>();
+    @Output() delete = new EventEmitter<MacroAction>();
```
fjozsef: The parent component should know the MacroAction, so emitting void is enough.

src/components/macro/macro.component.ts
```diff
@@ -1,31 +1,41 @@
...
+import 'rxjs/add/operator/publishReplay';
+import 'rxjs/add/operator/switchMap';
+import { Observable } from 'rxjs/Observable';
+import { ConnectableObservable } from 'rxjs/observable/ConnectableObservable';
 
 import { DragulaService } from 'ng2-dragula/ng2-dragula';
 
 import { Macro } from '../../config-serializer/config-items/Macro';
-import { MacroAction } from '../../config-serializer/config-items/macro-action';
+import { MacroAction } from '../../config-serializer/config-items/macro-action/MacroAction';
```
fjozsef: Why have added `/MacroAction` to the end?

src/components/macro/macro.component.ts
```diff
@@ -35,65 +45,83 @@ export class MacroComponent implements OnInit, OnDestroy {
...
 
-    ngOnDestroy() {
-        this.routeSubscription.unsubscribe();
+    deleteAction(macroAction: MacroAction, index: number) {
+        this.store.dispatch(MacroActions.deleteMacroAction(this.currentId, index, macroAction));
+        this.hideActiveEditor();
     }
 
+    private hideActiveEditor() {
+        this.macroItems.toArray().forEach((macroItem: MacroItemComponent, idx: number) => {
```
fjozsef: How about `this.macroItems.toArray()[this.activeEdit].cancelEdit();` ?

src/store/reducers/macro.ts
```diff
@@ -1,16 +1,149 @@
...
+
         default: {
             return state;
         }
     }
 }
+
+export function getMacro(id: number) {
+    // Convert to real number
+    id = +id;
```
fjozsef: What do you mean real number? If the parameter is string, then the caller should convert it to number.

src/components/macro/item/macro-item.component.ts
```diff
@@ -74,8 +72,8 @@ export class MacroItemComponent implements OnInit, OnChanges {
         this.cancel.emit();
     }
 
-    deleteAction(): void {
-        this.delete.emit();
+    deleteAction(macroAction: MacroAction): void {
```
fjozsef: No need for changes here.

src/components/macro/item/macro-item.component.scss
```diff
@@ -18,44 +20,60 @@
...
+            background: #f5f5f5;
         }
 
-        &--movable {
-            &:hover {
-                cursor: move;
+        &--wrap {
+            justify-content: space-between;
+
+            &.cursor {
```
fjozsef: I would rather use the classname `pointer` instead of `cursor`.

src/components/macro/macro.component.ts
```diff
@@ -1,31 +1,41 @@
...
     viewProviders: [DragulaService]
 })
-export class MacroComponent implements OnInit, OnDestroy {
+export class MacroComponent {
     @ViewChildren(MacroItemComponent) macroItems: QueryList<MacroItemComponent>;
-
-    private macro: Macro;
-    private routeSubscription: Subscription;
-    private hasChanges: boolean = false;
+    private macro$: Observable<Macro>;
```
fjozsef: Store the Macro itself, instead of the observable.

src/components/macro/macro.component.ts
```diff
@@ -1,31 +1,41 @@
...
+    private showNew: boolean = false;
+    private newMacro: Macro = undefined;
+    private activeEdit: number = undefined;
+    private currentId: number;
```
fjozsef: If the Macro itself is stored, then you can access to its id, so this will be not needed.

src/components/macro/macro.component.ts
fjozsef: But maybe the if is not needed, since we assume that there is an active.

src/components/macro/item/macro-item.component.scss
NejcZdovc: Done

src/components/macro/item/macro-item.component.scss
```diff
@@ -18,44 +18,48 @@
...
+
+.macro-action-editor__container {
+    padding-top: 0;
+    padding-bottom: 0;
+    border-radius: 0;
+    border-left: 0;
+    border-right: 0;
+}
+
+.cursor {
```
NejcZdovc: Can't find this line.

src/components/macro/item/macro-item.component.ts
NejcZdovc: You are right. Fixed

src/components/macro/item/macro-item.component.ts
NejcZdovc: Done

src/components/macro/macro.component.ts
NejcZdovc: IDEA auto generated it. Fixed

src/components/macro/macro.component.ts
NejcZdovc: I will move this if into this function, so that we can remove it on other places.

src/store/reducers/macro.ts
NejcZdovc: I moved convert operator to macro component. So that this function is clean.

src/components/macro/macro.component.ts
NejcZdovc: What is the benefit of storing whole Macro object? I only need id.

NejcZdovc: So instead of observable and async in template you would like to have whole object saved in the class and then access it in template without async?

fjozsef: Yes

NejcZdovc: Done

src/components/macro/macro.component.ts
NejcZdovc: Done

src/components/macro/macro.component.ts
```diff
@@ -35,65 +40,73 @@ export class MacroComponent implements OnInit, OnDestroy {
...
+                ));
+            }
+        });
+
+        route
+            .params
+            .select<string>('id')
+            .switchMap((id: string) => store.let(getMacro(+id)))
+            .subscribe((macro: Macro) => {
+                if (macro) {
```
fjozsef: Why do we need the if?

NejcZdovc: Done

src/components/macro/macro.component.ts
```diff
@@ -1,31 +1,36 @@
-import { Component, OnDestroy, OnInit, QueryList, ViewChildren } from '@angular/core';
+import { Component, QueryList, ViewChildren } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 
-import { Subscription } from 'rxjs/Subscription';
+import { Store } from '@ngrx/store';
+import 'rxjs/add/operator/let';
+import 'rxjs/add/operator/switchMap';
```
fjozsef: Shouldn't we import operator select here?

src/components/macro/macro.component.ts
```diff
@@ -35,65 +40,71 @@ export class MacroComponent implements OnInit, OnDestroy {
...
+                    +value[4].getAttribute('data-index')
+                ));
+            }
+        });
+
+        route
+            .params
+            .select<string>('id')
+            .switchMap((id: string) => store.let(getMacro(+id)))
+            .subscribe((macro: Macro) => {
```
fjozsef: Shouldn't we store the subscription and unscubscribe in ngOnDestroy?

NejcZdovc: Done

src/components/macro/macro.component.ts
NejcZdovc: I am not aware how to import select operator it's part of observable as far as I know.

fjozsef: See  src/store/reducers/macro.ts 's first line

NejcZdovc: I imported it manually and now it's working. When I did it automatically I was getting error that this operator doesn't exist, even though it's part of the documentation.

src/store/actions/macro.ts
```diff
@@ -1,12 +1,84 @@
 import { Action } from '@ngrx/store';
 
+import { Macro } from '../../config-serializer/config-items/Macro';
+import { MacroAction } from '../../config-serializer/config-items/macro-action/MacroAction';
```
fjozsef: Remove `/MacroAction`

NejcZdovc: Done

Issue #122, opened 2016-09-30
Animate the opening/closing of macro actions


mondalaci:
It'd look really neat if macro actions expanded/shrunk upon edit. Macro action titles would expand vertically to show the macro action editor upon click, and shrink upon cancel/save.

NejcZdovc:
So you would like to have animated toggle. Something like this http://www.w3schools.com/jquery/jquery_slide.asp?

mondalaci:
Yes, I think it'd work great!

fjozsef:
Should the new macro action editor also expand animated upon clicking "Add new macro action"?

mondalaci:
@fjozsef Yep, that'd be sweet!

PR #124, opened 2016-10-17
Macro items toggle

NejcZdovc:
This PR solves #122

### DIFF
--- a/src/components/macro/item/macro-item.component.ts
+++ b/src/components/macro/item/macro-item.component.ts
@@ -1,4 +1,17 @@
-import { Component, EventEmitter, Input, OnChanges, OnInit, Output, SimpleChanges } from '@angular/core';
+import {
+    Component,
+    EventEmitter,
+    Input,
+    OnChanges,
+    OnInit,
+    Output,
+    SimpleChanges,
+    animate,
+    state,
+    style,
+    transition,
+    trigger
+} from '@angular/core';
 
 import { KeyModifiers } from '../../../config-serializer/config-items/KeyModifiers';
 import {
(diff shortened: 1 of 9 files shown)


fjozsef:
LGTM, but I would want to wait [this anwser](https://github.com/UltimateHackingKeyboard/agent/issues/122#issuecomment-254598104), because the new macro action doesn't animate and I don't now if it should.

fjozsef:
Okay, we need them.

mondalaci:
Well, it's up to you guys whether to put the second animation into this PR or to another one.

fjozsef:
It has definitely should be in this one.

NejcZdovc:
It will be in this one

NejcZdovc:
Added sliding effect for new item as well

mondalaci:
This has made the user experience so much smoother!

The only problem I noticed is that now there's a 2 pixel wide separator line between macro actions versus the previous 1 pixel.

NejcZdovc:
This was introduced in PR #87, but it's fixed now.

src/components/macro/macro.component.scss
```diff
@@ -98,7 +101,7 @@ h1 {
 }
 
 .add-new__action-container {
-    border-top: 1px solid #ddd;
+    overflow: hidden;
```
fjozsef: Could you also add `flex-shrink: 0;` ? 
When a macro is selected for edit the buttons at the bottom disappear for me, but it solves the problem.

src/components/macro/item/macro-item.component.ts
```diff
@@ -25,6 +49,7 @@ export class MacroItemComponent implements OnInit, OnChanges {
     @Input() editable: boolean;
     @Input() deletable: boolean;
     @Input() moveable: boolean;
+    @Input() newItem: boolean;
```
fjozsef: I think we shouldn't add one more input. What if we would assume its a new item if `macroAction` is `null`?

NejcZdovc: Added.

PR #125, opened 2016-10-17
Remove services

NejcZdovc:
Removed unnecessary services and implement store where it was missing.


src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
```diff
@@ -46,8 +50,12 @@ export class SvgKeyboardKeyComponent implements OnInit, OnChanges {
 
     private labelSource: any;
     private labelType: LabelTypes;
+    private macros: Macro[];
 
-    constructor(private mapperService: MapperService, private uhkConfigurationService: UhkConfigurationService) { }
+    constructor(private mapperService: MapperService, private store: Store<AppState>) {
+        store.let(getMacroEntities())
```
NejcZdovc: Done

src/components/popover/tab/macro/macro-tab.component.ts
```diff
@@ -21,15 +24,14 @@ export class MacroTabComponent implements OnInit, Tab {
     private macroOptions: Array<Select2OptionData>;
     private selectedMacroIndex: number;
 
-    constructor(private uhkConfigurationService: UhkConfigurationService) {
-        this.macros = [];
+    constructor(private store: Store<AppState>) {
+        store.let(getMacroEntities())
```
NejcZdovc: Done

src/store/storage/local.ts
```diff
@@ -1,17 +1,17 @@
...
 
     getConfig(): UhkConfiguration {
         let configJson = localStorage.getItem('config');
         let config: UhkConfiguration;
+
         if (configJson) {
             config = new UhkConfiguration().fromJsObject(JSON.parse(configJson));
         } else {
-            config = this.dataProvider.getUHKConfig();
+            config = this.uhkConfiguration;
```
NejcZdovc: I think it should be. If not then we need to implement try catch and add notification about the error. User will need to refresh app to get default back. Maybe we shouldn't inject it into the local / electron, but check if `undefined` in `DataStorage` and there replace `undefined` with `uhkConfiguration`

src/store/storage/index.ts
```diff
@@ -4,33 +4,35 @@ import { Action } from '@ngrx/store';
...
 
-    constructor(private dataProvider: DataProviderService) {
+    constructor() {
+        this.initUHKJson();
         this.detectEnvironment();
     }
 
     initialState(): AppState {
-        let config: UhkConfiguration = this._environment.getConfig() || this.dataProvider.getUHKConfig();
+        let config: UhkConfiguration = this._environment.getConfig() || this.uhkConfiguration;
```
fjozsef: We are checking undefined here.

src/store/storage/local.ts
fjozsef: We check in `DataStorage`, see my comment.

src/components/popover/tab/macro/macro-tab.component.ts
```diff
@@ -1,13 +1,18 @@
 import { Component, Input, OnInit } from '@angular/core';
```
fjozsef: Please add `OnDestroy` interface.

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
```diff
@@ -1,5 +1,9 @@
 import { Component, Input, OnChanges, OnInit, SimpleChange } from '@angular/core';
```
fjozsef: Please add `OnDestroy` interface.

src/store/storage/index.ts
NejcZdovc: But this is called only once. So if a user deletes the local storage while using agent we have a problem, because this `initialState `will not be called again.

src/store/storage/local.ts
NejcZdovc: Commented bellow.

src/components/popover/tab/macro/macro-tab.component.ts
NejcZdovc: Done

src/store/storage/index.ts
fjozsef: You're right. We should add this `undefined` check to `saveState`.

NejcZdovc: Added

Issue #118, opened 2016-09-21
Add new macro


mondalaci:
Please let me share with you how I envision the add macro feature.

* Upon pressing the + macro button a new, empty macro gets created.
* The name of the new macro should be "New macro"
* If default macro name already exists then make it "New macro (2)" or increment the number until a non existing macro name is found.
* Select the default name of the macro with the cursor, so that the user can simply type a default name and hit Enter without any additional interactions.

PR #127, opened 2016-10-21
Add macro page

NejcZdovc:
PR based on #118. For simplicity Add page is basically only automatic add of the empty macro. After addition you are redirected to the edit page.  There is already opened new macro action waiting for you so you can start adding macros. One thing that is kinda bothering me is that when you add macro action, add action is closed. So if you would like to add 10 actions, you need to click on add new 10 times. Should we leave new action item opened until you click cancel? @mondalaci

Also in this PR I moved list of macros to the new component, so that can be used all around without macro component dependency.


fjozsef:
Why should we create a page for it? I think it should be just an event handler for the add macro button.

NejcZdovc:
You are completely correct. I was looking this trough component (re-usability) view, but this way we can loose a lot of unnecessary code. Fix is already done.

NejcZdovc:
As discussed with @mondalaci we will not be opening action item, but focus name of the newly added macro. Already implemented.

src/components/macro/list/macro-list.component.ts
```diff
@@ -37,27 +29,38 @@ import { getMacro } from '../../store/reducers/macro';
                 height: '*'
             })),
             transition('inactive <=> active', animate('500ms ease-out'))
+        ]),
+        trigger('togglerNew', [
+            state('void', style({
+                height: '0px'
+            })),
+            state('active',   style({
```
fjozsef: Remove mutliple spaces

NejcZdovc: Done

Issue #129, opened 2016-10-21
Highlight the currently selected menu item


mondalaci:
It'd be really nice to highlight the currently selected menu item in the side menu in some fashion.

PR #130, opened 2016-10-21
Menu items highlight

NejcZdovc:
PR based on #129

### DIFF
--- a/src/components/side-menu/side-menu.component.html
+++ b/src/components/side-menu/side-menu.component.html
@@ -9,7 +9,7 @@
         </div>
         <ul [@toggler]="animation.keymap">
             <li *ngFor="let keymap of keymaps$ | async" class="sidebar__level-2--item">
-                <div class="sidebar__level-2">
+                <div class="sidebar__level-2" [routerLinkActive]="['active']">
                     <a [routerLink]="['/keymap', keymap.abbreviation]">{{keymap.name}}</a>
                     <i *ngIf="keymap.isDefault" class="fa fa-star sidebar__fav"></i>
                 </div>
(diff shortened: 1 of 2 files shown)


mondalaci:
Wow, this was quick! I noticed 2 issues.

1. When I delete a macro or keymap the menu item that automatically gets selected doesn't get highlighted.
2. The background color of the highlighted menu item is rather low contrast, making it hard for the user to notice it. How about using a darker color? Possibly the color of the right border of the side menu, so that it'd merge together with it visually?

NejcZdovc:
1. This is because no macro is actually selected in menu. You are shown default macro page with first macro displayed from the list and because no ID is in the route, angular can't connect it to the macro. If this is something that is important, we can do work around.

2. Something like this:
image

NejcZdovc:
New colour added for the active item.

mondalaci:
The new menu item color seems to gel well with the UI.

I think it'd be a good idea to use the precise URL of the keymap/macro after deleting the previous one and highlight the relevant menu item accordingly.

I also noticed that when every macro gets deleted the empty state is the page that features the "To be done..." text. It'd make sense to display a helpful empty state like: "You don't have any macros. Try to add one!"

On a related note, I just noticed then if I have no macros then I cannot add a new one. The "To be done..." page is displayed instead of a new, empty macro upon trying to add a macro.

NejcZdovc:
Regarding the text and macros. You need to wait for PR #127 to be merged. After that all problems should disappear 😃

NejcZdovc:
I will create new issue where I will tackle this URL default states for keymap / macro.

mondalaci:
Ah, sorry about my new macro comment. I'm getting lots in your PRs :)

fjozsef:
Highlight is not working for the addon modules.

NejcZdovc:
Added height light for addon modules as well. I wanted to left them out until we implement them fully, but sure not a problem.

PR #153, opened 2016-11-17
Refactoring

NejcZdovc:
- Fixed default keymap selection in popover
- Fixed keymap default image
- Hide popover when you delete or switch between keymaps
- Disable layer switcher when popover is opened
- Renamed keymap component to keymap-edit
- Redirect to newly added keymap
- Redirect to specific URL of the favourite keymap (not just to /keymap) after keymap deletion
- Redirect to specific URL of the first macro (not just to /macro) after macro deletion


NejcZdovc:
I think this is everything that was mentioned in #56. Popover has no fixed width, so I think that this PR can close #56.

src/components/keymap/keymap.component.ts
```diff
@@ -35,4 +36,8 @@ export class KeymapComponent {
         this.keymap$ = keymapConnectable;
         keymapConnectable.connect();
     }
+
+    onKeymapDelete() {
```
fjozsef: Why do you pass the parameter in the html if you omit it here?

NejcZdovc: Removed, I forgot it there

src/store/effects/keymap.ts
```diff
@@ -28,20 +28,21 @@ export class KeymapEffects {
...
         .withLatestFrom(this.store)
         .do((latest) => {
-            let state: AppState = latest[1];
+            const state: AppState = latest[1];
 
             if (state.keymaps.entities.length === 0) {
                 this.router.navigate(['/keymap/add']);
             } else {
-                this.router.navigate(['/keymap']);
+                let favourite: Keymap = state.keymaps.entities.find(keymap => keymap.isDefault);
```
NejcZdovc: Fixed

src/components/keymap/header/keymap-header.component.ts
```diff
@@ -14,10 +14,9 @@ import { KeymapActions } from '../../../store/actions';
 })
 export class KeymapHeaderComponent {
     @Input() keymap: Keymap;
+    @Output() deleted: EventEmitter<boolean> = new EventEmitter();
```
fjozsef: What is the purpose of emitting a boolean here? As I see its emitted value is irrelevant, because the only place which is listening doesn't care about the value.

fjozsef: As I am more deeply in how it works I think it should be named to `deleteClick` with type `EventEmitter<void>` and move the `AppState `dispatching logic to the `Keymap(Edit)Component`.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -91,6 +91,7 @@ export class SvgKeyboardWrapComponent implements OnChanges {
     @Input() keymap: Keymap;
     @Input() popoverEnabled: boolean = true;
     @Input() tooltipEnabled: boolean = false;
+    @Input() deleted: boolean = false;
```
fjozsef: Unused input.

src/components/keymap/edit/keymap-edit.component.ts
```diff
@@ -10,17 +10,18 @@ import 'rxjs/add/operator/switchMap';
...
 
-import { Keymap } from '../../config-serializer/config-items/Keymap';
-import { AppState } from '../../store';
-import { getKeymap } from '../../store/reducers/keymap';
+import { Keymap } from '../../../config-serializer/config-items/Keymap';
+import { AppState } from '../../../store';
+import { getKeymap } from '../../../store/reducers/keymap';
 
 @Component({
     selector: 'keymap',
```
fjozsef: If you would want to rename the file, then you should rename the selector, too.

src/components/keymap/edit/keymap-edit.component.ts
```diff
@@ -10,17 +10,18 @@ import 'rxjs/add/operator/switchMap';
...
-    template: require('./keymap.component.html'),
-    styles: [require('./keymap.component.scss')]
+    template: require('./keymap-edit.component.html'),
+    styles: [require('./keymap-edit.component.scss')]
 })
 export class KeymapComponent {
```
fjozsef: And the class also.

src/components/svg/wrap/svg-keyboard-wrap.component.ts
NejcZdovc: Fixed

src/components/keymap/edit/keymap-edit.component.ts
NejcZdovc: Fixed

src/components/keymap/edit/keymap-edit.component.ts
NejcZdovc: Fixed

src/components/keymap/header/keymap-header.component.ts
NejcZdovc: Fixed

src/components/keymap/edit/keymap-edit.component.ts
```diff
@@ -10,17 +10,18 @@ import 'rxjs/add/operator/switchMap';
...
-    template: require('./keymap.component.html'),
-    styles: [require('./keymap.component.scss')]
+    selector: 'keymap-edit',
+    template: require('./keymap-edit.component.html'),
+    styles: [require('./keymap-edit.component.scss')]
 })
-export class KeymapComponent {
+export class KeymapEditComponent {
     private keymap$: Observable<Keymap>;
+    private keymapDeleted: boolean = false;
```
fjozsef: What is the purpose of this member? It is just set in the onKeymapDelete, but otherwise it seems to me it is unnecessary.

src/components/keymap/header/keymap-header.component.ts
```diff
@@ -14,10 +14,9 @@ import { KeymapActions } from '../../../store/actions';
 })
 export class KeymapHeaderComponent {
     @Input() keymap: Keymap;
+    @Output() deleteClick = new EventEmitter();
```
fjozsef: If the keymap-edit don't need to set if a delete has already happened or not, then this can be removed, too.

Issue #49, opened 2016-05-17
Popover mouse tab speed action 


csanyiarpad:
Use large icons and show text "accelerate" "decelerate" below them in smaller font.

Always show a help text below the button group saying something:

> Press this key along with mouse movement/scrolling to accelerate/decelerate the speed of the action.
> You can set the multiplier in [link_to_settings].

csanyiarpad:
[-] Decelerate | [+] Accelerate
button group solution

NejcZdovc:
@mondalaci is this still relevant?

mondalaci:
@NejcZdovc Yes, it is. It'd be worth to change the text to:

Press this key along with mouse movement/scrolling to adjust speed.
You can set the speed multiplier in [settings].

Settings should be linked properly.

The Accelerate / Decelerate texts should be added to the buttons.

NejcZdovc:
What would be a link to the settings? You mean http://localhost:8080/#/settings ?

mondalaci:
@NejcZdovc Yes, exactly.

PR #154, opened 2016-11-17
Mouse speed tab

NejcZdovc:
PR based on #49

### DIFF
--- a/src/components/popover/tab/mouse/mouse-tab.component.html
+++ b/src/components/popover/tab/mouse/mouse-tab.component.html
@@ -84,7 +84,7 @@
     <div *ngSwitchCase="3" class="mouse__config mouse__config--speed text-center">
         <div class="help-text--mouse-speed text-left">
             <p>Press this key along with mouse movement/scrolling to accelerate/decelerate the speed of the action.</p>
-            <p>You can set the multiplier in <a href="#" title="link to the setting">link to setting</a>.</p>
+            <p>You can set the multiplier in <a [routerLink]="['/settings']" title="the setting">the setting</a>.</p>
         </div>
         <div class="btn-group btn-group-lg" role="group">
             <div class="btn btn-default"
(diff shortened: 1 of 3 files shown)


NejcZdovc:
@mondalaci What do you think?

mondalaci:
@NejcZdovc Looks very nice! I can think of the following changes:

* Move the "You can set the multiplier in the setting." text below the buttons and change the "the setting" part to "settings".
* It'd be worth to put `cursor:pointer` to the side menu elements: Move, Scroll, Click, Speed

NejcZdovc:
@mondalaci Done.

fjozsef:
@mondalaci Should the 'the' word part of the hyperlink or not? I think it shouldn't. I would also suggest the title to be 'Settings' with capital S.
mouse

mondalaci:
@fjozsef Agreed!

NejcZdovc:
Fixed text

src/components/popover/tab/mouse/mouse-tab.component.html
```diff
@@ -84,15 +84,25 @@
...
                 [class.btn-primary]="mouseActionParam === MouseActionParam.accelerate"
-                (click)="setMouseActionParam(MouseActionParam.accelerate)">+</button>
+                (click)="setMouseActionParam(MouseActionParam.accelerate)"
+            >
+                +
+                <span>Accelerate</span>
+            </div>
+        </div>
+        <div class="help-text--mouse-speed last-help  text-left">
+            <p>You can set the multiplier in the <a [routerLink]="['/settings']" title="settings">Settings</a>.</p>
```
fjozsef: This should be the fix I mentioned: `title="Settings">settings`
Title with capital S, but the text is lower s.

NejcZdovc: Done

Issue #53, opened 2016-05-26
Key title animation when remapped


csanyiarpad:
When a key is being remapped, a flash like animation should play on the key:
1. the whole key goes white quickly (optionally immediately)
2. the white background fades back to the original one (dark grey), revealing the new (white) label on the key.

NejcZdovc:
@mondalaci is this still relevant?

fjozsef:
@NejcZdovc As far as I know, yes.

mondalaci:
Yes, it is, although it will only be truly relevant after positioning the key action popover correctly.

NejcZdovc:
blocked by #56

PR #156, opened 2016-11-17
Key action change animation

NejcZdovc:
PR based on #53

### DIFF
--- a/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
+++ b/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
@@ -1,4 +1,7 @@
-import {Component, Input, OnChanges, OnDestroy, OnInit, SimpleChange } from '@angular/core';
+import {
+    Component, Input, OnChanges, OnDestroy, OnInit, SimpleChange,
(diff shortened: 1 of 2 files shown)


NejcZdovc:
@mondalaci What do you think about the speed of the animation? You can test it with changing keys from Z-M, otherwise popover will block your view.

mondalaci:
@NejcZdovc It think it looks awesome - exactly the way I imagined! Such little changes go a long way regarding user experience.

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
```diff
@@ -44,16 +59,15 @@ export class SvgKeyboardKeyComponent implements OnInit, OnChanges, OnDestroy {
     @Input() height: number;
     @Input() width: number;
     @Input() keyAction: KeyAction;
+    @ViewChild ('rect') rect: ElementRef;
```
fjozsef: It seems to me it is unused.

NejcZdovc: Fixed

Issue #157, opened 2016-11-19
Uncaught TypeError: Cannot read property 'keyActionType' of null


mondalaci:
I've just tried to remove a keymap unsuccesfully (nothing seemed to happen), then reloaded Agent and got a white screen and this trace:

![untitled](https://cloud.githubusercontent.com/assets/300402/20455331/acbbf61a-ae59-11e6-9156-6e0af41fc15e.png)

Before trying to remove the keymap, I shut down webpack-dev-server which might have contributed to the issue.

NejcZdovc:
Can you reproduce this problem without shutting down the server?

NejcZdovc:
If you shut down the server, then you will not get anything, which is normal.

mondalaci:
I can reproduce the issue without shutting down the server. I've just created a new keymap, renamed it, deleted it, and the error was triggered again.

NejcZdovc:
Which keymap are you using?

mondalaci:
Usually, I:

1. add a new keymap (VIM, or Dvorak)
2. change a key action of the new keymap
3. delete the new keymap

Possibly repeat the above several times, and then the error is triggered.

I've been able to reproduce it about a dozen of times, so I'm confident that this bug is there. I will try to provide you a more accurate reproduction soon.

NejcZdovc:
Ok, thank you

mondalaci:
Finally got it!

1. Clear local storage and reload Agent
2. Add a new Dvorak keymap
3. Remap the backspace key of the Mouse layer of the new Dvorak keymap to move mouse upwards
4. Delete the new Dvorak keymap
5. Boom, an error appears in the console

NejcZdovc:
Ok, I found a very simple reproduction. Assign any action to the empty key. This will trigger an error.

NejcZdovc:
![image](https://cloud.githubusercontent.com/assets/9574457/20461434/663b22ec-aefe-11e6-88d1-41424644bbeb.png)

There are some null values in keyActions and this is a reason for the error.

PR #162, opened 2016-11-20
Fixed null problem

NejcZdovc:
PR resolving #157.

Problem is that we are using array and not objects. And for example if you set action for key that is in the middle of the module you would have empty elements in the array before.

For example:

You set action for backspace key. This is the last element in layer one in module one. That means that you would create array with `keyActions[7]= keyAction`. When we are reading and saving this into JSON and array, you have 6 empty elements before. As you can see on the image in #157 JSON contains null for all empry elements before.

What I did is a **temporary solution**. Proper solution would be to remove array and use objects. This way you wouldn't have any problems with this. But this would require quite a lot of work.

**svg-keyboar-wrap.component.ts**

```js
changeKeyAction(keyAction: KeyAction): void {
        let keyId = this.keyEditConfig.keyId;
        this.keyEditConfig.keyActions[keyId] = keyAction;
    }
```

### DIFF
--- a/src/config-serializer/config-items/Module.ts
+++ b/src/config-serializer/config-items/Module.ts
@@ -50,7 +50,11 @@ export class Module extends Serializable<Module> {
         return {
             id: this.id,
             pointerRole: PointerRole[this.pointerRole],
-            keyActions: this.keyActions.map(keyAction => keyAction.toJsObject())
+            keyActions: this.keyActions.map(keyAction => {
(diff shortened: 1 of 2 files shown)


mondalaci:
A temporary solution is definitely a right step forward, but how about adding another issue regarding the long-term solution?

NejcZdovc:
Will create new issue with a task what to do

Issue #160, opened 2016-11-19
The current keymap should not be offered in the switch to keymap action


mondalaci:
Because it does not make sense to switch to the already active keymap.

PR #165, opened 2016-11-20
Removed current keymap from switch

NejcZdovc:
PR based on #160

### DIFF
--- a/src/components/popover/popover.component.ts
+++ b/src/components/popover/popover.component.ts
@@ -35,6 +35,7 @@ enum TabName {
 })
 export class PopoverComponent implements OnInit {
     @Input() defaultKeyAction: KeyAction;
+    @Input() currentKeymap: Keymap;
 
     @Output() cancel = new EventEmitter<any>();
     @Output() remap = new EventEmitter<KeyAction>();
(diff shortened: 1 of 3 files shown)


src/components/popover/popover.component.html
```diff
@@ -46,7 +46,7 @@
         <layer-tab #tab *ngSwitchCase="TabName.Layer" class="popover-content" [defaultKeyAction]="defaultKeyAction"></layer-tab>
         <mouse-tab #tab *ngSwitchCase="TabName.Mouse" class="popover-content" [defaultKeyAction]="defaultKeyAction"></mouse-tab>
         <macro-tab #tab *ngSwitchCase="TabName.Macro" class="popover-content" [defaultKeyAction]="defaultKeyAction"></macro-tab>
-        <keymap-tab #tab *ngSwitchCase="TabName.Keymap" class="popover-content" [defaultKeyAction]="defaultKeyAction" [keymaps]="keymaps$ | async"></keymap-tab>
+        <keymap-tab #tab *ngSwitchCase="TabName.Keymap" class="popover-content" [defaultKeyAction]="defaultKeyAction" [keymaps]="keymaps$ | async" [currentKeymap]="currentKeymap"></keymap-tab>
```
fjozsef: I think the `keymaps$` should be filtered so we wouldn't need to pass the `currentKeymap` to the `keymap-tab`.

NejcZdovc: Done

src/components/popover/popover.component.ts
```diff
@@ -2,6 +2,8 @@ import { Component, EventEmitter, Input, OnInit, Output, ViewChild } from '@angu
 
 import { Store } from '@ngrx/store';
 
+import 'rxjs/add/operator/filter';
```
fjozsef: You are filtering an array, not an observable, so this is unused.

NejcZdovc: It's a left over. Fixed

PR #171, opened 2016-11-20
Fix keyboard animation

### LINKED ISSUE #170: Keyboard animation not working correctly
After updating to the latest stable angular (2.2.1), the keyboard animation stopped working correctly when selecting e.g. the mouse layer from the base layer.

The mouse layer comes in from left however it should come in from right.

fjozsef:
Fixes #170


fjozsef:
I had to update to the latest angular with it I think it is working. When I've upgraded to the latest angular the old animation started to produce errors.

PR #172, opened 2016-11-20
Add default keymap icon hint

### LINKED ISSUE #167: Default keymap icon hint
It would be useful to provide some explanation for the user regarding the star icon.

My idea is that when hovering over the filled star icon of the side menu or the filled star icon next to the default keymap title we could display the following text:

> This is the default keymap which gets activated when powering the keyboard.

When hovering over the unfilled star icon of a non-default keymap the following text should be displayed:

> Makes this keymap the default keymap which gets activated when powering the keyboard.

fjozsef:
Fixes #167

### DIFF
--- a/src/store/reducers/keymap.ts
+++ b/src/store/reducers/keymap.ts
@@ -62,7 +62,12 @@ export default function(state = initialState, action: Action): KeymapState {
 
         case KeymapActions.SET_DEFAULT:
             newState = state.entities.map((keymap: Keymap) => {
-                keymap.isDefault = (keymap.abbreviation === action.payload);
+                if (keymap.abbreviation === action.payload || keymap.isDefault) {
+                    let newKeymap: Keymap = new Keymap();
+                    Object.assign(newKeymap, keymap);
+                    keymap = newKeymap;
+                    keymap.isDefault = keymap.abbreviation === action.payload;
+                }
 
                 return keymap;
             });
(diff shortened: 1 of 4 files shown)


mondalaci:
Looks perfect.

src/store/reducers/keymap.ts
```diff
@@ -62,7 +62,12 @@ export default function(state = initialState, action: Action): KeymapState {
 
         case KeymapActions.SET_DEFAULT:
             newState = state.entities.map((keymap: Keymap) => {
-                keymap.isDefault = (keymap.abbreviation === action.payload);
+                if (keymap.abbreviation === action.payload || keymap.isDefault) {
```
NejcZdovc: Why did you add or operator with `keymap.isDefault`?

src/store/reducers/keymap.ts
```diff
@@ -62,7 +62,12 @@ export default function(state = initialState, action: Action): KeymapState {
...
+                    let newKeymap: Keymap = new Keymap();
+                    Object.assign(newKeymap, keymap);
+                    keymap = newKeymap;
```
NejcZdovc: Why are you setting newKeymap to the old one. Why don't just do something like:

```
Object.assign(newKeymap, keymap);
newKeymap.isDefault = keymap.abbreviation === action.payload;
return newKeymap
```

fjozsef: I wanted to have one return statement., because this way the code more readable for me. 
If you need a new keymap, then you create a new and replace the old one, otherwise it remains the same. I know that your approach does the same, but I think mine is easier to understand.

PR #183, opened 2016-11-21
Tooltip texts for key actions

### LINKED ISSUE #102: Tooltip for add keymap

fjozsef:


mondalaci:
For mouse actions, let's write "Action:" instead of "Type:"

When I hover over the up arrow key of the Mod layer of the QWERTY keymap I get the following error:

untitled

mondalaci:
Now I can only notice a last, minor issue. The tooltip of the menu key contains "Scancode: 118 ()".

Shouldn't this scancode have a name? In case a scancode doesn't have a name the parentheses shouldn't be displayed.

mondalaci:
Well done! You're good to go.

Issue #184, opened 2016-11-21
Alphabetically sort keymap and macro names in the side menu


mondalaci:

fjozsef:
The sorting should respect the local alphabet. [More information](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare)

PR #187, opened 2016-11-22
Sidemenu alphabetic sort

NejcZdovc:
PR based on #184

### DIFF
--- a/src/components/side-menu/side-menu.component.ts
+++ b/src/components/side-menu/side-menu.component.ts
@@ -39,8 +39,15 @@ export class SideMenuComponent {
             addon: 'active'
         };
 
-        this.keymaps$ = store.let(getKeymapEntities());
-        this.macros$ = store.let(getMacroEntities());
+        this.keymaps$ = store.let(getKeymapEntities())
+            .map((keymaps: Keymap[]) => {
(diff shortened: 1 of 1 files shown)


src/components/side-menu/side-menu.component.ts
```diff
@@ -39,8 +39,31 @@ export class SideMenuComponent {
             addon: 'active'
         };
 
-        this.keymaps$ = store.let(getKeymapEntities());
-        this.macros$ = store.let(getMacroEntities());
+        this.keymaps$ = store.let(getKeymapEntities())
+            .map((keymaps: Keymap[]) => {
+                return keymaps.sort((first: Keymap, second: Keymap) => {
+                    if (first.name > second.name) {
```
fjozsef: How about returning `first.name.localCompare(second.name)`?

NejcZdovc: I would stay with current implementation. https://jsperf.com/string-sort-n

fjozsef: I have talked with László about it. We should support the local alphabet, because we want localize the agent in the future, so it is better if we use local compare now.

NejcZdovc:
Done

Issue #185, opened 2016-11-21
Add new keymap: scrollbar not on the side of the window


mondalaci:
For some reason, the scrollbar is not at the right edge of the window on the add new keymap page. There's some extra space there.

![screenshot from 2016-11-22 00-02-35](https://cloud.githubusercontent.com/assets/300402/20504347/2ea21e36-b047-11e6-8d84-93239f829274.png)

NejcZdovc:
Yes this happens, because we scroll inline window (left part of the app) not the whole window.

PR #188, opened 2016-11-22
Fixed scroll position

NejcZdovc:
PR based on #185


fjozsef:
It just seems to me a quick fix, but you don't fix the problem where it is coming from. Please see #189

NejcZdovc:
It was not a quick fix, because other pages have a different scroll behaviour, but ok let's go with your PR.

PR #189, opened 2016-11-23
Fix main-app scrollbar

### LINKED ISSUE #185: Add new keymap: scrollbar not on the side of the window

fjozsef:
Fixes #185

### DIFF
--- a/src/components/macro/edit/macro-edit.component.scss
+++ b/src/components/macro/edit/macro-edit.component.scss
@@ -1,3 +1,9 @@
+:host {
+    width: 100%;
+    height: 100%;
+    display: block;
+}
+
 .not-found {
     margin-top: 30px;
     font-size: 16px;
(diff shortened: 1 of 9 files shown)


src/components/macro/edit/macro-edit.component.scss
```diff
@@ -1,5 +1,11 @@
-.not-found {
-    margin-top: 30px;
-    font-size: 16px;
-    text-align: center;
+:host {
+    width: 100%;
+    height: 100%;
+    display: block;
+
+    .not-found {
```
NejcZdovc: Why did you nest `.not-found` in the host?

fjozsef: Because hierarchically it is used in the host.

fjozsef: Why do we want it to be flat?

NejcZdovc: Because it will reduce time in style step in pixel pipeline. That's why BEM is good, because you have one level only, so page is rendering faster. Nesting is no good 😄

PR #190, opened 2016-11-24
Refactor: Replace NoneAction

### LINKED ISSUE #176: Keyactions
We are currently using array with specific ID's for key actions in keymap. We need to change this to objects and move this specific ID into object of the action.

This way we can manipulate array (add, remove) without any problems.

When this is done we can revert  #162.

### LINKED ISSUE #182: Replace NoneAction with null in JSON representation

fjozsef:
Fixes #182 

DataStorage should be cleared.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,13 +30,13 @@
     "webpack-svgstore-plugin": "^2.2.2"
   },
   "dependencies": {
-    "@angular/common": "2.3.0-beta.1",
-    "@angular/compiler": "2.3.0-beta.1",
-    "@angular/core": "2.3.0-beta.1",
-    "@angular/forms": "2.3.0-beta.1",
-    "@angular/platform-browser": "2.3.0-beta.1",
-    "@angular/platform-browser-dynamic": "2.3.0-beta.1",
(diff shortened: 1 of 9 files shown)


fjozsef:
Should I rename `index` to `id`?

mondalaci:
`id` reads good to me.

NejcZdovc:
image

Why does is say backspace now, before it was just an arrow?

NejcZdovc:
I would leave `index `here, because we are saving in this variable position on the keymap not ID of the keyaction.

mondalaci:
Backspace was intentionally choosen because the left arrow looked ambiguous. It's not clear whether it means "cursor left" or backspace.

fjozsef:
The backspace text was changed in #183. It has nothing to do with this PR.

NejcZdovc:
I didn't see mentioned anywhere, that's why this was something new for me. Ok then there is just a mapping problem.

fjozsef:
I had to downgrade the angular to the latest stable, because there was an error in the animation with the latest beta. (I couldn't find out if the bug is at our side or at their side.)

Issue #168, opened 2016-11-20
Make the width of the key action popover constant


mondalaci:
When various actions or subactions are selected, the width of the key action popover increases which can be a bit disturbing for the user. The width of the popover should always be constant.

These cases are:
* The Mouse action and the Speed subaction is selected
* The Macro action is selected with something other than "Select macro" in select2
* The Keymap action is selected with "Switch to keymap" in select2

Issue #139, opened 2016-10-22
Key action popover positioning


mondalaci:
The popover should be positioned according to the related key.

fjozsef:
Blocked by #56

mondalaci:
I'd like to add an important point regarding the implementation of this issue.

We were thinking about positioning the popover above / below the choosen key depending on the space available above / below the key. I think it would be more intuitive for the user if we always displayed the popover below the choosen key. The reasons being:

* The content of various tabs within the popover have very different height. If the popover was positioned above a key and the user wanted to switch between the tabs then the tab bar would move around vertically every time when the content of the new tab has different height than the content of the previous tab which would be very disturbing.
* The more consistent the behavior of the UI, the easier it is to interact with it. Always displaying the popover below the key is more consistent than displaying it above or below depending on the space available.

If the popover displayed is outside of the current viewport, then we should smooth-scroll downwards until the whole popver fits into the viewport.

PR #192, opened 2016-11-25
Popover rework

NejcZdovc:
PR based on #139, #168 

- [x] Add positioning based on key
- [x] Add overlay for popover
- [x] Click on overlay will close popover
- [x] Add animation to the popover
- [x] Graphical glitch for the content of the popover
- [x] Add space bellow keyboard if there is not enough space
- [x] Popover width should be constant

### DIFF
--- a/src/components/popover/popover.component.scss
+++ b/src/components/popover/popover.component.scss
@@ -4,6 +4,7 @@
     padding: 0;
     max-width: 568px;
     width: 100%;
+    user-select: none;
 
     &.leftArrow {
         .arrowCustom {
--- a/src/components/popover/popover.component.scss
+++ b/src/components/popover/popover.component.scss
@@ -26,6 +26,10 @@
     }
 }
 
+.nav-tabs > li {
+    overflow: hidden;
+}
+
 .arrowCustom {
     position: absolute;
     top: -16px;
--- a/src/components/popover/popover.component.scss
+++ b/src/components/popover/popover.component.scss
@@ -1,11 +1,66 @@
 :host {
     display: flex;
     flex-direction: column;
-    max-width: none;
     padding: 0;
+    max-width: 568px;
+    width: 100%;
+    transform: translateX(-50%);
+
+    > .container-fluid {
+        overflow: hidden;
+    }
+
+    &.leftArrow {
+        transform: none;
+
+        .arrowCustom {
+            transform: none;
+            left: 22px;
+        }
+    }
+
+    &.rightArrow {
+        transform: none;
+
+        .arrowCustom {
+            transform: none;
+            right: 22px;
+            left: auto;
+        }
+    }
+}
+
+.arrowCustom {
+    position: absolute;
+    top: -16px;
     left: 50%;
     transform: translateX(-50%);
-    top: 100px;
+    width: 41px;
+    height: 16px;
+
+    &:before {
+        content: '';
+        width: 0;
+        height: 0;
+        border-left: 21px solid transparent;
+        border-right: 21px solid transparent;
+        border-bottom: 17px solid rgba(0, 0, 0, 0.2);
+        display: block;
+        position: absolute;
+        top: -1px;
+    }
+
+    &:after {
+        content: '';
+        width: 0;
+        height: 0;
+        border-left: 20px solid transparent;
+        border-right: 20px solid transparent;
+        border-bottom: 16px solid #f7f7f7;
+        display: block;
+        position: absolute;
+        top: 0;
+    }
 }
 
 .popover-action {
--- a/src/components/popover/popover.component.scss
+++ b/src/components/popover/popover.component.scss
@@ -4,16 +4,8 @@
     padding: 0;
     max-width: 568px;
     width: 100%;
-    transform: translateX(-50%);
-    visibility: hidden;
-
-    &.display {
-        visibility: visible;
-    }
 
     &.leftArrow {
-        transform: none;
-
         .arrowCustom {
             transform: none;
             left: 22px;
@@ -21,8 +13,6 @@
     }
 
     &.rightArrow {
-        transform: none;
-
         .arrowCustom {
             transform: none;
             right: 22px;
@@ -108,13 +98,17 @@
 .popover-overlay {
     position: fixed;
     width: 100%;
-    height: 100%;
+    height: 0;
     top: 0;
     left: 0;
-    display: none;
+    z-index: 1050;
+    background: rgba(0, 0, 0, 0);
+    transition: background 200ms ease-out, height 0ms 200ms linear;
 
     &.display {
-        display: block;
+        height: 100%;
+        background: rgba(0, 0, 0, 0.2);
+        transition: background 200ms ease-out;
     }
 }
 
--- a/src/components/popover/popover.component.scss
+++ b/src/components/popover/popover.component.scss
@@ -1,13 +1,14 @@
-:host {
+.popover {
     display: flex;
     flex-direction: column;
     padding: 0;
     max-width: 568px;
     width: 100%;
     transform: translateX(-50%);
+    visibility: hidden;
 
-    > .container-fluid {
-        overflow: hidden;
+    &.display {
+        visibility: visible;
     }
 
     &.leftArrow {
@@ -27,6 +28,10 @@
             right: 22px;
             left: auto;
         }
+    }
+
+    > .container-fluid {
+        overflow: hidden;
     }
 }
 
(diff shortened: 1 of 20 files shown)


NejcZdovc:
@mondalaci please check out the animation of popover. Is this something like you wanted?

mondalaci:
@NejcZdovc Close, but I meant moving upwards, then downwards, not the other way around. Can you also animate the color/opacity of the overlay at the same time (getting darker gradually)?

NejcZdovc:
😄 ok try now

mondalaci:
Just noticed that upon closing the popover it switches to None action. The ease out interval is only 200ms but it's still noticable. The popover should not change its content at all until ease out finishes.

mondalaci:
I've also managed to accidentally highlight the text within the popover a fair amount of times. It'd be worth to make all the text inside of the popover non-selectable.

NejcZdovc:
Inside the popover I didn't disable text select

mondalaci:
Visual glitch fixed. You da man! :+1: 

I can also confirm that the text within the popover is not selectable anymore.

mondalaci:
Right now, upon opening the popover extra margin is added to the bottom. I know this is an intentional solution but I believe it'd be a better idea to have a constant margin because right now the margin causes the scroll bar to appear and disappear all the time upon opening and closing the popover. I'd rather have the scroll bar all the time. Eventually, we'll add add-ons to the bottom of keymap pages, so the extra margin won't be so noticable in full HD.

NejcZdovc:
@mondalaci you can try it again

mondalaci:
@NejcZdovc This is perfect the way it is. Great job!

fjozsef:
bug

fjozsef:
In a small size (1123x470) I've clicked on a 6.
bug2

I expected from the popover to point at the key 6, but it doesn't.

NejcZdovc:
Sadly I can't reproduce either scenario that you pasted. Can you try again with the latest version (didn't change anything)? 

@mondalaci Can you reproduce this errors?

mondalaci:
I cannot reproduce either of these issues.

src/components/popover/popover.component.ts
```diff
@@ -31,49 +34,95 @@ enum TabName {
...
+            })),
+            transition('opened => closed', [
+                animate('200ms ease-out', keyframes([
+                    style({ transform: 'translateY(0)', visibility: 'visible', opacity: 1, offset: 0 }),
+                    style({ transform: 'translateY(30px)', visibility: 'hidden', opacity: 0 , offset: 1 })
+                ]))
+            ]),
+            transition('closed => opened', [
+                animate('200ms ease-out', keyframes([
+                    style({ transform: 'translateY(30px)', visibility: 'visible', opacity: 0, offset: 0 }),
```
fjozsef: The visibility should be in the style of the transition, not in the style of the animate, since it is not something we want to be animated.

src/components/popover/popover.component.ts
```diff
@@ -93,4 +142,28 @@ export class PopoverComponent implements OnInit {
     selectTab(tab: TabName): void {
         this.activeTab = tab;
     }
+
+    onOverlay() {
+        this.cancel.emit(undefined);
+    }
+
+    private calculatePosition() {
+        const offsetLeft: number = this.wrapPosition.left + 265;
```
fjozsef: Where is 265 coming from?

src/components/popover/popover.component.ts
```diff
@@ -93,4 +142,28 @@ export class PopoverComponent implements OnInit {
...
+
+        if (this.leftArrow) {
+            newLeft = this.keyPosition.left;
+        } else if (this.rightArrow) {
+            newLeft = this.keyPosition.left - popover.offsetWidth + this.keyPosition.width;
+        } else {
+            newLeft -= popover.offsetWidth / 2;
+        }
+
+        this.topPosition = this.keyPosition.top + this.keyPosition.height + 7 + window.scrollY;
```
fjozsef: Where is 7 coming from?

NejcZdovc: This is a distance that we add so, that we have a little bit a space between the key and popover.

src/components/popover/popover.component.ts
NejcZdovc: This is a width of the side menu with a padding of 15px.

src/components/popover/popover.component.ts
NejcZdovc: Fixed

src/components/popover/popover.component.ts
fjozsef: Can you add a comment for magic numbers?

NejcZdovc: Done

fjozsef:
Is it normal that there is no remap animation?

NejcZdovc:
For me remap animation is working.

fjozsef:
Not for me, not even after fetch & rebase :(

fjozsef:
Sorry, my bad. Old angular version. It is fine.

src/components/popover/popover.component.scss
```diff
@@ -1,11 +1,66 @@
-:host {
+.popover {
     display: flex;
     flex-direction: column;
-    max-width: none;
     padding: 0;
+    max-width: 568px;
```
fjozsef: I don't really like this magic here :/

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
```diff
@@ -68,7 +69,11 @@ export class SvgKeyboardKeyComponent implements OnInit, OnChanges, OnDestroy {
     private subscription: Subscription;
     private animation: string = 'inactive';
 
-    constructor(private mapper: MapperService, private store: Store<AppState>) {
+    @HostListener('click') onClick() {
```
fjozsef: Why do we need this? The parent could get the nativeElement with using a local template variable.

src/components/svg/wrap/svg-keyboard-wrap.component.html
```diff
@@ -4,12 +4,12 @@
...
                       [@layerState]="layer.animation"
                       [moduleConfig]="layer.modules"
-                      (keyClick)="onKeyClick($event.moduleId, $event.keyId)"
+                      (keyClick)="onKeyClick($event.moduleId, $event.keyId, $event.keyTarget)"
                       (keyHover)="onKeyHover($event.moduleId, $event.event, $event.over, $event.keyId)"
         >
         </svg-keyboard>
     </div>
-    <popover *ngIf="popoverShown" [defaultKeyAction]="popoverInitKeyAction" [currentKeymap]="keymap" (cancel)="hidePopover()" (remap)="onRemap($event)"></popover>
+    <popover [visible]="popoverShown" [keyPosition]="keyPosition" [wrapPosition]="wrapPosition" [defaultKeyAction]="popoverInitKeyAction" [currentKeymap]="keymap" (cancel)="hidePopover()" (remap)="onRemap($event)"></popover>
```
NejcZdovc: So that animations are working correctly

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
NejcZdovc: We talked about this on the hangouts. This way is more elegant.

src/components/popover/popover.component.scss
NejcZdovc: What is here not to like? :smile:

fjozsef: The agent will be localized in the future, so it is possible that there is a language which uses longer words for tabs, so the popover should grow, but can't because of this magic limit.

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
fjozsef: I remember something that we talked about this, but is don't find it elegant now. Maybe we should talk about it again and remind me what was reason.

src/components/svg/wrap/svg-keyboard-wrap.component.html
NejcZdovc: Why would you like to destroy popover component every single time? User would use it to configure the whole keyboard for 100+ times?

fjozsef: Okay, I am convinced :)

src/components/popover/popover.component.scss
NejcZdovc: And on that day I would like to still keep it the same, because otherwise popover would go out of the control and we would have a very big popover, where our goal is very compact popover.

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
NejcZdovc: Sure we can talk about it again :slightly_smiling_face:

PR #193, opened 2016-11-26
Enable OnPush ChangeDetection in SvgKeyboardWrapComponent

fjozsef:
I had to downgrade the angular to the latest stable, because there was an error in the animation with the latest beta. (I couldn't find out if the bug is at our side or at their side.)

### DIFF
--- a/src/components/svg/wrap/svg-keyboard-wrap.component.html
+++ b/src/components/svg/wrap/svg-keyboard-wrap.component.html
@@ -1,7 +1,7 @@
 <template ngIf="layers">
     <layers [class.disabled]="popoverShown" (select)="selectLayer($event.oldIndex, $event.index)" [current]="currentLayer"></layers>
     <div class="keyboard-slider" (mouseout)="hideTooltip($event)">
-        <svg-keyboard *ngFor="let layer of layers"
+        <svg-keyboard *ngFor="let layer of layers; trackBy: trackKeyboard"
                       [@layerState]="layer.animation"
                       [moduleConfig]="layer.modules"
                       (keyClick)="onKeyClick($event.moduleId, $event.keyId)"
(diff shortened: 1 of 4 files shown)


src/components/svg/wrap/svg-keyboard-wrap.component.html
```diff
@@ -1,7 +1,7 @@
 <template ngIf="layers">
     <layers [class.disabled]="popoverShown" (select)="selectLayer($event.oldIndex, $event.index)" [current]="currentLayer"></layers>
     <div class="keyboard-slider" (mouseout)="hideTooltip($event)">
-        <svg-keyboard *ngFor="let layer of layers"
+        <svg-keyboard *ngFor="let layer of layers; trackBy: trackKeyboard"
```
NejcZdovc: Why would be this used here?

fjozsef: Otherwise the angular would instantiate a new svg-keyboard, now it calls OnChanges,

NejcZdovc: Ok

src/store/reducers/keymap.ts
```diff
@@ -99,11 +101,29 @@ export default function(state = initialState, action: Action): KeymapState {
...
+            changedKeymap.layers[layerIndex] = changedLayer;
+
+            const moduleIndex: number = action.payload.module;
+            const module: Module = changedLayer.modules[moduleIndex];
+            const changedModule: Module = new Module();
+            Object.assign(changedModule, module);
+            changedLayer.modules[moduleIndex] = changedModule;
+
+            const keyIndex: number = action.payload.key;
+            changedModule.keyActions[keyIndex] = action.payload.keyAction;
```
NejcZdovc: I think that here should be a new constructor copy

fjozsef: Fixed

PR #197, opened 2016-11-27
Gray out trash at last keymap instead of not rendering

### LINKED ISSUE #196: Tweaking the prevention of last keymap deletion
I [commented](https://github.com/UltimateHackingKeyboard/agent/issues/178#issuecomment-261991683) on the related issue that went unnoticed:

> As for the implementation if this issue, it may be a better idea to always display the trash icon, but gray it out and add a label on hover that says: "The last keymap cannot be deleted."

I'm confident that it'd be a less confusing implementation for the user. Instead of making the delete icon disappear the user could see why deletion is not possible.

fjozsef:
Fixes #196

### DIFF
--- a/src/components/keymap/header/keymap-header.component.html
+++ b/src/components/keymap/header/keymap-header.component.html
@@ -21,12 +21,12 @@
            [title]="starTitle"
            (click)="setDefault()"
         ></i>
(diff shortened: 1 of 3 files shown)


mondalaci:
The UI looks great. Well done!

PR #202, opened 2016-11-29
Tooltip UX fixes

NejcZdovc:
PR based on #102 

- [x] Let's place the tooltip below the key, just like for the key action popover.
- [x] Ideally, the tip of the tooltip should point right to the edge of the key, not inside of it.
- [x] Right now, if I move the pointer from the key to the tooltip quickly, the tooltip stays there instead of showing a new tooltip for the key at the pointer.
- [x] The current tooltip also has the tendency of staying in place even if I scroll a long way

@mondalaci Please check it out


mondalaci:
All issues fixed. Well done!

PR #210, opened 2016-12-03
Keymap switch animation

### LINKED ISSUE #207: Keymap switch animation
Right now, when clicking on a different keymap than the current one in the side menu, some keys animate as if they're being remapped. According to Jozsi this is a side-effect of the current component. We should not do this because it's misleading.

Even though we shouldn't animate keys, we could use some animation. I'm thinking about quickly fading out the previous keyboard and fading in the new keyboard when switching to another keymap. I mean the whole keyboard including its case and keys, not the keys individually.

fjozsef:
Fixes #207 

@NejcZdovc 
My solution is very hacky. I am using `ngFor` to reinstantiate the slider, so one can fade out, while the other fades in. Any ideas how could it be avoided are welcome,


NejcZdovc:
Like you say this approach is very hacky. What if you send information about repaint into `svg-keyboard-key.component.ts` via inputs from `svg-keyboard-wrap.component.ts`?

fjozsef:
It won't solve the problem. We have to have two keyboard parallel (in time), one that fades in and one that fades out. I have been wokring on trying to solve this with creating a KeyboardSliderComponent and then dinamically create a new one (basically this is what ngFor does) and destroy the old one, so hopefully we could imperatively control this, but it would increase the complexity.

mondalaci:
The new keymap should fade in after the old keymap fades out. According to my understanding, it's not parallel but sequential. Maybe I'm missing something regarding the implementation details, but I though I mention it.

fjozsef:
In this implementation the fade in starts when the fade out, so they are parallel. What should be the expected behavior?

fjozsef:
@mondalaci As I have reread your comment then using sequential animation (which seems to me the expected then) would solve our problem.

mondalaci:
@fjozsef Excellent! Sequential animations for the win! :)

fjozsef:
@NejcZdovc Please review
@mondalaci Please review the UI :)

fjozsef:
Transition duration has been reduced from 500ms to 200ms.

NejcZdovc:
Personally I don't like this change. Now when you are switching between keymaps, then are irritating blinking. I preferred the old way.

My suggestion:
- go with a parallel animation
or
- remove animation for the keymaps and just disable change effect for keys when switching between keymaps

mondalaci:
I'm personally fine with the current animation. What's your take on this, @fjozsef?

On the other hand, I don't insist to the animation, so I'm fine with removing it. I don't have a clue how the parallel animation would look like. Alternatively, we could fade out, and fade in the text on the keycaps instead of the whole keyboard.

NejcZdovc:
Personally I would go with a removal and only fixing the key animation when switching. Less is sometimes better :smiley:

fjozsef:
I like both parallel (first version) and sequential (current version) animation, from it the switching seems smoother.  I suggest to talk about the issue related things in #207. Here should only be implementation specific things regarding to the specified issue.

Issue #143, opened 2016-10-22
Capture keystroke


NejcZdovc:
If user clicks on the key with a middle mouse button, capture keystroke is activated. If the user clicks outside of the current active key, capture keystroke is canceled. When capture keystroke is active key is blinking.

NejcZdovc:
Blocked by #228

PR #229, opened 2016-12-19
Added capture keystroke to the popover

### LINKED ISSUE #228: Capture keystroke popover

NejcZdovc:
Resolves #228

### DIFF
--- a/src/components/popover/tab/keypress/keypress-tab.component.ts
+++ b/src/components/popover/tab/keypress/keypress-tab.component.ts
@@ -69,6 +69,8 @@ export class KeypressTabComponent implements OnChanges, Tab {
     onKeysCapture(event: {code: number, left: boolean[], right: boolean[]}) {
         if (event.code) {
             this.scanCode = event.code;
+        } else {
+            this.scanCode = 0;
         }
 
         this.leftModifierSelects = event.left;
--- a/src/components/popover/tab/keypress/keypress-tab.component.ts
+++ b/src/components/popover/tab/keypress/keypress-tab.component.ts
@@ -71,13 +71,8 @@ export class KeypressTabComponent implements OnChanges, Tab {
             this.scanCode = event.code;
         }
 
-        event.left.forEach((value: boolean, index: number) => {
-            this.leftModifierSelects[index] = value;
-        });
-
-        event.right.forEach((value: boolean, index: number) => {
-            this.rightModifierSelects[index] = value;
-        });
+        this.leftModifierSelects = event.left;
+        this.rightModifierSelects = event.right;
     }
 
     fromKeyAction(keyAction: KeyAction): boolean {
--- a/src/components/popover/tab/keypress/keypress-tab.component.ts
+++ b/src/components/popover/tab/keypress/keypress-tab.component.ts
@@ -66,6 +66,20 @@ export class KeypressTabComponent implements OnChanges, Tab {
         return keystrokeAction.scancode > 0 || keystrokeAction.modifierMask > 0;
     }
 
+    onKeysCapture(event: {code: number, left: boolean[], right: boolean[]}) {
+        if (event.code) {
+            this.scanCode = event.code;
+        }
+
+        event.left.forEach((value: boolean, index: number) => {
+            this.leftModifierSelects[index] = value;
+        });
+
+        event.right.forEach((value: boolean, index: number) => {
+            this.rightModifierSelects[index] = value;
+        });
+    }
+
     fromKeyAction(keyAction: KeyAction): boolean {
         if (!(keyAction instanceof KeystrokeAction)) {
             return false;
@@ -139,17 +153,11 @@ export class KeypressTabComponent implements OnChanges, Tab {
         modifierSelects[index] = !modifierSelects[index];
     }
 
-    // TODO: change to the correct type when the wrapper has added it.
-    /* tslint:disable:no-unused-variable: It is used in the template. */
-    private onLongpressChange(event: any) {
-        /* tslint:enable:no-unused-variable: */
+    onLongpressChange(event: {value: string}) {
         this.selectedLongPressIndex = +event.value;
     }
 
-    // TODO: change to the correct type when the wrapper has added it.
-    /* tslint:disable:no-unused-variable: It is used in the template. */
-    private onScancodeChange(event: any) {
-        /* tslint:enable:no-unused-variable */
+    onScancodeChange(event: {value: string}) {
         this.scanCode = +event.value;
     }
 
--- a/src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
+++ b/src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
@@ -1,4 +1,5 @@
 import { Component, EventEmitter, HostListener, Output } from '@angular/core';
+import { CaptureService } from '../../../../services/capture.service';
 
 @Component({
     selector: 'capture-keystroke-button',
@@ -9,19 +10,12 @@ export class CaptureKeystrokeButtonComponent {
     @Output() capture = new EventEmitter<any>();
 
     private record: boolean;
-    private first: boolean;
-    private leftModifiers: Map<number, boolean>;
-    private rightModifiers: Map<number, boolean>;
-    private mapping: Map<number, number>;
+    private first: boolean; // enable usage of Enter to start capturing
 
-    constructor() {
-        this.leftModifiers = new Map<number, boolean>();
-        this.rightModifiers = new Map<number, boolean>();
-        this.mapping = new Map<number, number>();
+    constructor(private captureService: CaptureService) {
         this.record = false;
-
-        this.reset();
-        this.populateMapping();
+        this.captureService.initModifiers();
+        this.captureService.populateMapping();
     }
 
     @HostListener('keyup')
@@ -34,23 +28,18 @@ export class CaptureKeystrokeButtonComponent {
     @HostListener('keydown', ['$event'])
     onKeyDown(e: KeyboardEvent) {
         const code: number = e.keyCode;
+        const enter = 13;
 
         if (this.record) {
+            e.preventDefault();
             this.first = false;
 
-            if (this.mapping.has(code)) {
-                this.saveScanCode(this.mapping.get(code));
+            if (this.captureService.hasMap(code)) {
+                this.saveScanCode(this.captureService.getMap(code));
             } else {
-                // Left
-                if (e.location === 1) {
-                    this.leftModifiers.set(code, true);
-                }
-                // Right
-                else {
-                    this.rightModifiers.set(code, true);
-                }
+                this.captureService.setModifier((e.location === 1), code);
             }
-        } else if (code === 13) {
+        } else if (code === enter) {
             this.record = true;
             this.first = true;
         }
@@ -63,15 +52,13 @@ export class CaptureKeystrokeButtonComponent {
     }
 
     start(): void {
-        if (!this.record) {
-            this.record = true;
-        }
+        this.record = true;
     }
 
     private saveScanCode(code?: number) {
         this.record = false;
-        const left: boolean[] = this.reMap(this.leftModifiers);
-        const right: boolean[] = this.reMap(this.rightModifiers);
+        const left: boolean[] = this.captureService.getModifiers(true);
+        const right: boolean[] = this.captureService.getModifiers(false);
 
         this.capture.emit({
             code,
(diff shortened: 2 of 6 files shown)


src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,184 @@
...
+import { Component, EventEmitter, HostListener, Output } from '@angular/core';
 
 @Component({
     selector: 'capture-keystroke-button',
     template: require('./capture-keystroke-button.component.html'),
     styles: [require('./capture-keystroke-button.component.scss')]
 })
-export class CaptureKeystrokeButtonComponent implements OnInit {
+export class CaptureKeystrokeButtonComponent {
+    @Output() keys = new EventEmitter<any>();
```
fjozsef: Rename event to `capture`

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,184 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, EventEmitter, HostListener, Output } from '@angular/core';
 
 @Component({
     selector: 'capture-keystroke-button',
     template: require('./capture-keystroke-button.component.html'),
     styles: [require('./capture-keystroke-button.component.scss')]
 })
-export class CaptureKeystrokeButtonComponent implements OnInit {
+export class CaptureKeystrokeButtonComponent {
```
fjozsef: Rename class to `KeystrokeCapturerComponent`.
It is not only a button anymore if it does the capturing.

src/components/popover/tab/keypress/keypress-tab.component.ts
```diff
@@ -66,6 +66,20 @@ export class KeypressTabComponent implements OnChanges, Tab {
         return keystrokeAction.scancode > 0 || keystrokeAction.modifierMask > 0;
     }
 
+    onKeysCapture(event: {code: number, left: boolean[], right: boolean[]}) {
+        if (event.code) {
+            this.scanCode = event.code;
+        }
+
+        event.left.forEach((value: boolean, index: number) => {
```
fjozsef: Why not simply `this.leftModifierSelects = event.left;`?

NejcZdovc:
Ok, this is not ready for a review :smile: I only uploaded it so that @mondalaci can test it out if I covered all cases. A lot of this code will be moved to the service so that can be used in #143 as well.

fjozsef:
Oh, okay. I just couldn't resist :D I will wait then

NejcZdovc:
I love your enthusiasm :+1:

mondalaci:
This seems to work exactly as it's supposed to. Well done! The only minor issue I noticed is that when I trigger the button with Enter, the capture immediately stops and nothing gets captured. If it's easy to fix then you can do it, but it doesn't worth a lot of time.

NejcZdovc:
One think that I would like to change is scope of listener. Would it be ok to just listen to this capture component. What this means is that when you click on button listener will be triggered, but button needs to be focused. If you click outside of the button listener will not be triggered, you would need to click on button again. Is this still acceptable? This way we don't need to listen to the whole window. @mondalaci

mondalaci:
@NejcZdovc Your suggestion does make sense to me, as I've just realized that if I start capturing, close the popover, reopen it again then capturing still stays active which is weird.

NejcZdovc:
@mondalaci this two things are not related, but glad that you like the idea :smile: I just pushed new version with new scope and enter, that you requested.

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
NejcZdovc: Changed

src/components/popover/tab/keypress/keypress-tab.component.ts
NejcZdovc: True

mondalaci:
@NejcZdovc I've been trying hard to find any bugs and I couldn't. Your PR is officially approved. :)

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
NejcZdovc: I would leave it as it is

src/components/popover/tab/keypress/keypress-tab.component.ts
```diff
@@ -66,6 +66,15 @@ export class KeypressTabComponent implements OnChanges, Tab {
         return keystrokeAction.scancode > 0 || keystrokeAction.modifierMask > 0;
     }
 
+    onKeysCapture(event: {code: number, left: boolean[], right: boolean[]}) {
+        if (event.code) {
```
fjozsef: Is it intentional that you don't clear the scancode when the event's code doesn' exists?

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,77 @@
...
+    @Output() capture = new EventEmitter<any>();
+
     private record: boolean;
+    private first: boolean;
```
fjozsef: Please comment what is the purpose of this member.

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,77 @@
...
-    ngOnInit() { }
+        if (this.record) {
+            this.first = false;
+
+            if (this.captureService.hasMap(code)) {
+                this.saveScanCode(this.captureService.getMap(code));
+            } else {
+                this.captureService.setModifier((e.location === 1), code);
+            }
+        } else if (code === 13) {
```
fjozsef: Create a local const for 13 or comment it.

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,77 @@
...
+            this.record = true;
+            this.first = true;
+        }
+    }
```
fjozsef: Call prevent default to work in cases like ctrl + d

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,77 @@
...
+            right
+        });
+
+        this.reset();
     }
 
+    private reset() {
+        this.first = false;
+        this.captureService.initModifiers();
+
```
fjozsef: Remove empty line

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,77 @@
...
+    @HostListener('focusout')
+    onFocusOut() {
+        this.record = false;
+        this.reset();
+    }
 
     start(): void {
-        this.record = true;
+        if (!this.record) {
```
fjozsef: Is it faster to check if it false and only assign true in that case?

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
NejcZdovc: What do you mean by that?

src/components/popover/tab/keypress/keypress-tab.component.ts
NejcZdovc: No it's not. Fixed

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
fjozsef: `e.preventDefault();` otherwise if you want to capture ctrl + d chrome wants to save the page in the bookmark.

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
NejcZdovc: We check here for false, so that we know that you are not already capturing.

fjozsef: But it doesn't matter the capturing has already started or not. At the end of this function `record` will be true, no matter what is its value at the beginning.  (If it was true we don't touch it, if it was false we set it to true. It is the same is set to true, no matter the initial value.)

NejcZdovc: Not true. You can trigger keydown effect with scan code. In function `SaceScanCode` you set value to `false`. If you would click on a button between `keydown` and `keyup` event, you would trigger `saveScanCode` in `keyup` event.

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
NejcZdovc: Done

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
fjozsef: I think you misunderstood me. I don't say that it's value is a constant `true` or anything like that.
I say it doesn't matter if you check for false or not. It would have the same behavior if you would leave this function as it was.

NejcZdovc: I understood you correctly. We need to check if value is `false` and only then set it to true. Otherwise user could click on button Capture (while capturing) and if he clicks on this button between keyup event which saves `SaveScanCode` (this function set `record` value to false), he would set value to `true` and this would tigger `saveScanCode` in `keyup` event, which would set it to `none`. Current implementation is correct, otherwise you would introduce a bug, it's hard to trigger it, but it would be there.

fjozsef: Maybe we should talk this about in hangout, but I am strongly believe this function would behave the same as it was. 
It doesn't matter if there are other functions. I am talking about basic stuffs. 
Function has an input parameter this. 
if this.record is true then !this.record is false, so at the end of the function this.record is true.
if this.record is false then !this.record is true, then we set this.record to true, so at the end of the function this.record is true.
It doesn't matter who, when, why calls the start function after calling it this.record will be true.
So I would keep it as it was before.

NejcZdovc: Ok. Fixed

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,76 @@
...
     selector: 'capture-keystroke-button',
     template: require('./capture-keystroke-button.component.html'),
     styles: [require('./capture-keystroke-button.component.scss')]
 })
-export class CaptureKeystrokeButtonComponent implements OnInit {
+export class CaptureKeystrokeButtonComponent {
+    @Output() capture = new EventEmitter<any>();
+
     private record: boolean;
+    private first: boolean; // enable usage of enter
```
fjozsef: It is still not clear me how does it enables the enter. Why do we have to treat it individually?

NejcZdovc: > The only minor issue I noticed is that when I trigger the button with Enter, the capture immediately stops and nothing gets captured. If it's easy to fix then you can do it, but it doesn't worth a lot of time.

This is for triggering capturing with enter when you are tabbing around the popover

NejcZdovc: I added some more explanation to the comment

fjozsef: Have you considered calling `start` on focusIn? Maybe it would solve that problem, too.

NejcZdovc: I wouldn't to that, because you would trigger capturing unintentionally. For example if you are tabbing on popover and you would like to go from scan code to the modifiers, you would trigger capture in the process.

fjozsef: Okay, you're right.

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -1,23 +1,76 @@
...
+    @HostListener('keyup')
+    onKeyUp() {
+        if (this.record && !this.first) {
+            this.saveScanCode();
+        }
+    }
+
+    @HostListener('keydown', ['$event'])
+    onKeyDown(e: KeyboardEvent) {
+        e.preventDefault();
```
NejcZdovc: Fixed

fjozsef:
@NejcZdovc Should I squash the commits?

NejcZdovc:
@fjozsef Yes

PR #231, opened 2016-12-24
Added capture to keys

NejcZdovc:
Resolves #143

PR #233, opened 2016-12-24
Make Esc close popover

### LINKED ISSUE #232: Make Esc close the key action popover
Except when capturing is active in which case it should capture Esc.

fjozsef:
Fixes #232

### DIFF
--- a/src/components/popover/popover.component.ts
+++ b/src/components/popover/popover.component.ts
@@ -1,6 +1,6 @@
 import {
-    Component, ElementRef, EventEmitter, Input, OnChanges, Output, SimpleChanges, ViewChild, animate, keyframes,
-    state, style, transition, trigger
+    Component, ElementRef, EventEmitter, HostListener, Input, OnChanges, Output, SimpleChanges, ViewChild,
+    animate, keyframes, state, style, transition, trigger
 } from '@angular/core';
 
 import { Store } from '@ngrx/store';
@@ -49,7 +49,7 @@ enum TabName {
             transition('opened => closed', [
                 animate('200ms ease-out', keyframes([
                     style({ transform: 'translateY(0)', visibility: 'visible', opacity: 1, offset: 0 }),
-                    style({ transform: 'translateY(30px)', visibility: 'hidden', opacity: 0 , offset: 1 })
+                    style({ transform: 'translateY(30px)', visibility: 'hidden', opacity: 0, offset: 1 })
                 ]))
             ]),
             transition('closed => opened', [
@@ -143,6 +143,10 @@ export class PopoverComponent implements OnChanges {
         }
     }
 
+    @HostListener('keydown.escape') onEscape(): void {
+        this.cancel.emit();
+    }
+
     selectTab(tab: TabName): void {
         this.activeTab = tab;
     }
@@ -156,7 +160,7 @@ export class PopoverComponent implements OnChanges {
         const popover: HTMLElement = this.popoverHost.nativeElement;
         let newLeft: number = this.keyPosition.left + (this.keyPosition.width / 2);
 
-        this.leftArrow = newLeft <  offsetLeft;
+        this.leftArrow = newLeft < offsetLeft;
         this.rightArrow = (newLeft + popover.offsetWidth) > offsetLeft + this.wrapPosition.width;
 
         if (this.leftArrow) {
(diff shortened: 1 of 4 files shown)


src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.html
```diff
@@ -1,6 +1,7 @@
 <button type="button" class="btn btn-sm btn--capture-keystroke"
             [ngClass]="{'btn-default': !record, 'btn-info': record}"
-            (click)="start()"
+            (keydown.enter)="onButtonEnterDown($event)"
```
NejcZdovc: Why was this changed?

fjozsef: With this the pressing of the enter could be intercepted at the button level, so the component would only get the click when capturing is in progress.

NejcZdovc: I am not sure that I understand it, but what is the benefit of this that justify additional event listeners and complexity?

fjozsef: It is necessary since the current implementation is hacky. The recording is starting before it should (with enter) and it has to be filtered with `first`.

NejcZdovc: Recording does not start before. Event for keyup is added same here as it was added before so it's the same. I check if we should start recording in the event keydown, you are checking it elsewhere. I don't agree with it that it's hacky, your approach is different and more complex. One other thing that I don't understand is how is this related to the popover escape, which is the name of this PR?

fjozsef: There are more than improvements ones are needed others don't. If you could suggest a better name for the pr than it is welcome. 
My original issue why I've started the modification of these things is because the saveScanCode is called on both keydown and up and for me it is not clear why should be called in both places. With the escape it caused me to set to escape at down and clear it on up. If you would like then maybe I could reproduce it again.

fjozsef: There were also problems with the interception, with the recording the popover also closed.
(It was propagated however it shouldn't be when it was recorded.).
My solution fixes those and this was easiest solution for me to fix those bugs.

NejcZdovc: `saveScanCode ` was called in both places so that you could save modifiers as well, without needing to press scan code. Your current implementation does not implement this, so you can't save `ctrl` for example or combination of `alt` and `ctrl`.

I think that this is not correct approach and we should be focused on fixing bugs (I didn't find any with current implementation) and implementing new things, not implementing new ones. If you don't understand my code please ask me, I am more then happy to explain it to you. Sometimes easier is not necessary better.

I think this PR should focus on implementing esc to close popover. I think that this can be done without radical change.

fjozsef: It shouldn't be called twice. One capture should emit only one event. Thanks for pointing out that single modifiers don't work. I am about to investigate it more, but I am not sure that is good idea to store the modifiers in the service.

Like I said, the bug occurs in the popover. I can capture the escape of keydown/keyup (I don't remember now which) event which should be intercepted by the capture button at the recording.

I don't understand what do you mean by "Sometimes easier is not necessary better." I believe the code should be always as simple as possible. So if more simple == easier then it is better. But I don't understand how is it connecting to this topic.

Some changes definitly should be done else the closing could not work with escape. If it needs radical changes, then that should be. (In my opinion my changes aren't radical.) I don't see how else could it be solved in a minor way. If you'll come up with a less complex one then please send a PR for it or provide suggestions for how could I reduce the complexity with providing a working solution.

NejcZdovc: With `Sometimes easier is not necessary better.` I was referring to your statement `this was easiest solution for me to fix those bugs`. 

I would suggest that we go trough this code on a hangout. One capture does not emit two events.

I will create PR in the next days for this popover esc problem. I think that service is ok, I created PR #231 with same service and everything work correctly.

fjozsef: `One capture does not emit two events.` I've investigated this more and you're right. 
Okay, let's hangout about it sometime.

fjozsef:
@NejcZdovc Please review again.

PR #239, opened 2016-12-29
Use ids in key actions instead of references

fjozsef:
Partially reverts e48fdea

@NejcZdovc Review at least the tooltip part.

### DIFF
--- a/src/components/svg/wrap/svg-keyboard-wrap.component.ts
+++ b/src/components/svg/wrap/svg-keyboard-wrap.component.ts
@@ -38,6 +38,11 @@ import { AppState } from '../../../store';
 import { KeymapActions } from '../../../store/actions';
 import { PopoverComponent } from '../../popover';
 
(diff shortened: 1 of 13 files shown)


src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -194,13 +202,33 @@ export class SvgKeyboardWrapComponent implements OnInit, OnChanges {
...
+
+    hidePopover(): void {
+        this.popoverShown = false;
+    }
 
+    selectLayer(index: number): void {
+        this.currentLayer = index;
+    }
+
+    private getKeyActionContent(keyAction: KeyAction): typeof SvgKeyboardWrapComponent.prototype.tooltipData.content {
```
NejcZdovc: I would create interface for it or paste definition from the top `Observable<{ name: string, value: string }[]>`

PR #242, opened 2016-12-30
Removed fast capturing for add keymap page

### LINKED ISSUE #238: Fast capturing should only work on keymap edit page
It shouldn't work on add keymap page neither on keymap-tab

NejcZdovc:
Fixes #238

### DIFF
--- a/src/components/keyboard/slider/keyboard-slider.component.ts
+++ b/src/components/keyboard/slider/keyboard-slider.component.ts
@@ -79,6 +79,7 @@ export class KeyboardSliderComponent implements OnChanges {
     @Input() layers: Layer[];
     @Input() currentLayer: number;
     @Input() keybindAnimationEnabled: boolean;
+    @Input() capturingEnabled: boolean;
     @Output() keyClick = new EventEmitter();
     @Output() keyHover = new EventEmitter();
     @Output() capture = new EventEmitter();
--- a/src/components/svg/keyboard/svg-keyboard.component.ts
+++ b/src/components/svg/keyboard/svg-keyboard.component.ts
@@ -11,6 +11,7 @@ import { SvgModule } from '../module';
 export class SvgKeyboardComponent implements OnInit {
     @Input() moduleConfig: Module[];
     @Input() keybindAnimationEnabled: boolean;
+    @Input() capturingEnabled: boolean;
     @Output() keyClick = new EventEmitter();
     @Output() keyHover = new EventEmitter();
     @Output() capture = new EventEmitter();
--- a/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
+++ b/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
@@ -70,6 +70,7 @@ export class SvgKeyboardKeyComponent implements OnInit, OnChanges, OnDestroy {
     @Input() width: number;
     @Input() keyAction: KeyAction;
     @Input() keybindAnimationEnabled: boolean;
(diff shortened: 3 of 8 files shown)


src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
```diff
@@ -70,6 +70,7 @@ export class SvgKeyboardKeyComponent implements OnInit, OnChanges, OnDestroy {
     @Input() width: number;
     @Input() keyAction: KeyAction;
     @Input() keybindAnimationEnabled: boolean;
+    @Input() popoverEnabled: boolean;
```
fjozsef: Rename this to `capturingEnabled`

src/components/keyboard/slider/keyboard-slider.component.ts
```diff
@@ -79,6 +79,7 @@ export class KeyboardSliderComponent implements OnChanges {
     @Input() layers: Layer[];
     @Input() currentLayer: number;
     @Input() keybindAnimationEnabled: boolean;
+    @Input() popoverEnabled: boolean;
```
fjozsef: Rename this to `capturingEnabled`

src/components/svg/keyboard/svg-keyboard.component.ts
```diff
@@ -11,6 +11,7 @@ import { SvgModule } from '../module';
 export class SvgKeyboardComponent implements OnInit {
     @Input() moduleConfig: Module[];
     @Input() keybindAnimationEnabled: boolean;
+    @Input() popoverEnabled: boolean;
```
fjozsef: Rename this to `capturingEnabled`

src/components/svg/module/svg-module.component.ts
```diff
@@ -14,6 +14,7 @@ export class SvgModuleComponent {
     @Input() keyboardKeys: SvgKeyboardKey[];
     @Input() keyActions: KeyAction[];
     @Input() keybindAnimationEnabled: boolean;
+    @Input() popoverEnabled: boolean;
```
fjozsef: Rename this to `capturingEnabled`

src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
NejcZdovc: We are passing data for`popoverEnabled` trough the inputs to the key and based on this value we decide if we would like to enable capture or not. I think that we should keep original naming of the variable, because we don't send data for capturing, but for popover and then at the end we decide based on popover functionality if we would also like to enable capturing or not.

fjozsef: On a key level (component level) there should be no expectation about popover or any surrounding component, or about parents. 
If you are thinking in components, then a component should work in every use case. e.g. If I want to add a key somewhere e.g. on tutorial page where definitely will not be any popover used, then `popoverEnabled` would be weird property.

NejcZdovc: I am not so attached to this naming, so it's changed now :smile: But your argument can be span around as well.

fjozsef: I am aware that some of the components aren't well designed. Of course those are design mistakes, too. Best would be to create a list and refactor/rename those behaviours/property names. But now all we can do is not introduce more.

PR #244, opened 2016-12-30
Disable button when action is not valid in popover

### LINKED ISSUE #161: Display the Remap Key button as disabled whenever input is invalid
Currently, the key Remap Key button of the key action popover doesn't do anything when the content of the popover is incomplete which is correct. But we should also visually disable the button whenever the input is invalid.

NejcZdovc:
Resolves #161

### DIFF
--- a/src/components/popover/popover.component.ts
+++ b/src/components/popover/popover.component.ts
@@ -79,6 +79,7 @@ export class PopoverComponent implements OnChanges {
     @ViewChild('popover') popoverHost: ElementRef;
 
     public tabName = TabName;
+    public isValidAction: boolean;
     private activeTab: TabName;
     private keymaps$: Observable<Keymap[]>;
     private leftArrow: boolean = false;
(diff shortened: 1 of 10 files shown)


src/components/popover/tab/keypress/keypress-tab.component.ts
```diff
@@ -16,6 +16,8 @@ export class KeypressTabComponent implements OnChanges, Tab {
     @Input() defaultKeyAction: KeyAction;
     @Input() longPressEnabled: boolean;
 
+    @Output() validAction = new EventEmitter();
```
fjozsef: Why have you created a new event? What don't you use `keyActionValid` function of the `Tab` interface? It is even easily accesible from popover.

NejcZdovc: But we need to trigger validator every single time that user do an action, so we need some kind of a event emitter.

src/components/popover/popover.component.ts
```diff
@@ -79,6 +79,7 @@ export class PopoverComponent implements OnChanges {
     @ViewChild('popover') popoverHost: ElementRef;
 
     public tabName = TabName;
+    public isValidAction: boolean;
```
fjozsef: Rename to `keyActionValid`

src/components/popover/popover.component.ts
```diff
@@ -134,20 +135,24 @@ export class PopoverComponent implements OnChanges {
...
     }
 
-    @HostListener('keydown.escape') onEscape(): void {
+    @HostListener('keydown.escape')
+    onEscape(): void {
         this.cancel.emit();
     }
 
     selectTab(tab: TabName): void {
+        this.isValidAction = !(tab === this.tabName.Keypress || tab === this.tabName.Mouse);
```
fjozsef: Remove this line and emit from other tabs as well to set the validation flag correctly.

NejcZdovc: My thinking behind this line was that we add outputs only to tabs that need to emit, so that we don't add unnecessary events, where we know that it will always be false.

fjozsef: I know, but I think it is the tabs responsibility when it find the key action valid, so we shouln't assume anything here.

NejcZdovc: Ok, will change the logic

src/components/popover/popover.component.ts
NejcZdovc: Done

src/components/popover/popover.component.ts
NejcZdovc: Done

fjozsef:
I am totally okay with the current implementation, however I have to ask. Wouldn't it be better if we would create an abstract class from `Tab`, so we could move `@Output() validAction = new EventEmitter();` to that class?

PR #245, opened 2016-12-30
Added menu key

### LINKED ISSUE #243: Capturing a menu key is not working

NejcZdovc:
Resolves #243

### DIFF
--- a/src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
+++ b/src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
@@ -18,9 +18,10 @@ export class CaptureKeystrokeButtonComponent {
         this.captureService.populateMapping();
     }
 
-    @HostListener('keyup')
-    onKeyUp() {
+    @HostListener('keyup', ['$event'])
+    onKeyUp(e: KeyboardEvent) {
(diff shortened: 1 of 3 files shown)


NejcZdovc:
@fjozsef Can you please test it out and see where do we need to add prevent, on mac I can't trigger this menu. Thank you

fjozsef:
@NejcZdovc On keyup

NejcZdovc:
Ok try now if it's ok

src/components/popover/widgets/capture-keystroke/capture-keystroke-button.component.ts
```diff
@@ -18,9 +18,10 @@ export class CaptureKeystrokeButtonComponent {
         this.captureService.populateMapping();
     }
 
-    @HostListener('keyup')
-    onKeyUp() {
+    @HostListener('keyup', ['$event'])
+    onKeyUp(e: KeyboardEvent) {
         if (this.record && !this.first) {
+            e.preventDefault();
```
fjozsef: It will not work here since here only the modifiers are captured / saved.

NejcZdovc: Is it working now?

PR #246, opened 2016-12-30
Use indices instead of ids in JSON and in binary representation

### LINKED ISSUE #212: Remove macro ids from JSON and Binary representation
We should use indices.

fjozsef:
Closes #212 

@mondalaci I had to remove Serializable, because most of the classes need different parameters so common interface couldn't work.


mondalaci:
@fjozsef Shouldn't be a big deal, I believe. According to our earlier conversations, `Serializable` was used for dumping which we don't use anymore, and there are other ways to catch bugs, so it's not critical.

fjozsef:
@mondalaci Great! That is what I was thinking :)

PR #256, opened 2017-02-07
Add the whole user configuration to the AppState

### LINKED ISSUE #219: Free storage saveState relying on environments config
Currently our storage reloads (JSON parsing, model building, etc) the UhkConfiguration when we save the state.
The problems with this approach;

- Performance issues, parsing should be done ideally only once
- LocalStorage could be deleted since the app start, so there will be no config to use
- Blocks #212. More detail there.

The application state should store the whole config, which should be used in the serialization.

### LINKED ISSUE #252: Add the whole user configuration to the store
All of the `UserConfiguration` fields should be accessible from the `Store` via `AppState`.
Note: The `UserConfiguration` itself should be included in the state, since later we will have hardware configuration which also has `dataModelVersion` field.

Blocks #142 and partially #213

fjozsef:
Closes #252
Fixes #219 

@mondalaci Please test if everything works as expected (no new changes should be present from the point of the user), I just want to make sure that I didn't break anything.


mondalaci:
@fjozsef Just tested this PR, although I wasn't really sure what to test. I tried both the web and the electron build, remapped a couple of keys, and everything worked as expected. I have two observations:

1. The build process seems to be taking significantly more time than previously. For example, `npm run dev` takes half a minute, even though I have a beefy workstation.
2. When using the electron build, the configuration gets reset upon restarting the application. This may be normal behavior assuming that electron wipes local storage upon restart in which case this is a non-issue. I guess electron will read the config via USB anyways.

fjozsef:
@mondalaci As I have written in the description, there is no change from the point of the user. I just needed someone, who tests it so I can make sure I didn't break anyhting.

1. The build process time shoulnd't change (at least not significantly), have you compared it with the master? If you did, then I would be pleased if you could tell how much build time have you measured.
2. It has nothing to do with this pr, please stick the actual changes in the pr. I suggest you to open an issue about it.

mondalaci:
Just measured the build time of master, and it seems to be the same as this branch. I'm fairly sure that it has taken less time to build Agent a while back, but I'm not sure which change has introduced the extra delay. In any case, this is not a pressing issue so let's not focus on it. This PR looks all right to me.

PR #271, opened 2017-02-18
UHK Detection

### LINKED ISSUE #258: UHK detection screen
Agent-electron should start up by searching for the UHK. This should involve calling `usb.findByIds()`. Only the following should be displayed in center of the window:

searching-for-uhk

If `usb.findByIds()` returns `undefined`, the following should be displayed:

agent-cannot-find-uhk

and then `usb.findByIds()` should be polled on a 100 ms basis as long as it returns with a non-undefined value, at which point Agent should proceed further.

During the execution of Agent, if the UHK gets unplugged it should be detected and we should reveal the "Cannot find your UHK" screen again.

fjozsef:
Closes #258


mondalaci:
This is working so smothly. :) Pass!

PR #277, opened 2017-03-11
Update to webpack 2.x

### LINKED ISSUE #101: Upgrade to webpack2
It's still in beta so maybe we shouldn't switch right away, but we should periodically check it to know when will be the good time to upgrade.

fjozsef:
Closes #101


fjozsef:
It is a known issue that it gives two warnings at compile time. Those should be fixed by angular.

PR #289, opened 2017-04-28
doc: fix renamed uhk-config.json to user-config.json

attilacsanyi:


mondalaci:
@attilacsanyi Great catch, thanks!

PR #290, opened 2017-05-29
Issue 201 - Mark selected key with blue

attilacsanyi:
Animate the selected keyboard key background to blue on each layer / module in order to visually show the currently edited key to the user.


mondalaci:
This is clearly getting better, but I still found a couple of issues:

1. If I click twice on a key in rapid succession (for example by double clicking on it) while the animation is still in progress then the key stays highlighted.
2. I'd like the animation delay to be shorter. Exactly as short as the animation delay of the popover.
3. When clicking on the "Remap key" button, the key transitions A) from blue to gray and then B) flashes from gray to white to gray very quickly. Let's only use the B) transition in this case and make it as long as before.

Thank you!

mondalaci:
Just added bullet point number 4.

attilacsanyi:
thanks @mondalaci for the quick review, I am going to fix those 4 points as well!

attilacsanyi:
Solutions for points 1.-3. have already been pushed.

mondalaci:
@attilacsanyi This looks perfect from a user standpoint. Well done!

@fjozsef Would you please take a look at the code?

attilacsanyi:
Sorry @fjozsef for the late answer. Check my comments below:

**1.** `tsconfig` were added in order to eliminate VSCode errors around not determined source root for those shared typescript files _(try to remove it and lots of errors going to be displayed)_
[Relevant [commit](https://github.com/UltimateHackingKeyboard/agent/pull/290/commits/690e917d74e583441c1b50d7aae2357234390da0).]

**2.** I have tried the simplest solution after our discussion while passing down only `selected key action`, but in this case clicking on one of the 'empty' key all the other 'empty' keys also turned to blue _(as in this case `null` would have assigned to all the 'empty' keys)_
[Relevant [commit](https://github.com/UltimateHackingKeyboard/agent/pull/290/commits/8a6ff955cb9c33cf1e410420ed789c7371c94019) which were reverted.]

attilacsanyi:
Do I need to change anything more here or can be merged? @fjozsef @mondalaci

shared/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
```diff
@@ -48,6 +48,12 @@ enum LabelTypes {
                 ])
             ])
         ]),
+        trigger('active', [
+            // http://colorblendy.com/#!/multiply/4099e5/cccccc
+            state('1', style({ fill: '#4099e5' })), // Signiture blue color blending
```
mondalaci: Fix typo: Signiture -> Signature

shared/src/components/svg/wrap/svg-keyboard-wrap.component.ts
```diff
@@ -136,7 +137,8 @@ export class SvgKeyboardWrapComponent implements OnInit, OnChanges {
                 moduleId,
                 keyId
             };
-
+            // Save the selected key
```
mondalaci: This comment seems rather redundant. I think it'd be worth to remove it.

shared/src/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+      "core-js",
+      "select2",
+      "file-saver"
+    ]
+  },
+  "exclude": [
+    "node_modules",
+    "config-serializer"
+  ]
+}
```
mondalaci: Please add an ending newline here.

mondalaci:
@attilacsanyi Apart from the minor suggested changes this looks good to me.

What's your opinion, @fjozsef?

shared/src/components/keyboard/slider/keyboard-slider.component.html
```diff
@@ -3,6 +3,8 @@
     [moduleConfig]="layer.modules"
     [keybindAnimationEnabled]="keybindAnimationEnabled"
     [capturingEnabled]="capturingEnabled"
+    [selectedKey]="selectedKey"
+    [selected]="selectedKey && selectedKey.layerId === index"
```
fjozsef: How about merging the selected and selectedKey into this
[selected]="selected?.layerId === index ? selected : undefined" ?
You could use similar logic at the other places.

shared/src/components/svg/wrap/svg-keyboard-wrap.component.ts
attilacsanyi: Fixed, thx.

shared/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
attilacsanyi: Fixed, thx.

shared/src/components/keyboard/slider/keyboard-slider.component.html
attilacsanyi: Not sure we can do that as `this.selectedKey = { layerId: this.currentLayer, moduleId, keyId };` has the info of which is the selected key on each level and I have to pass down this info to determine the `selected` flag for each level _(level can be layer, module or key)_.

PR #291, opened 2017-05-29
move travis to trusty

stephengroat:
avoids need for sudo


mondalaci:
@stephengroat Thank you so much! Simplifications like this are always welcome.

PR #292, opened 2017-05-31
Add appveyor CI integration

### LINKED ISSUE #250: Auto update
We should implement cross-platform automatic updates eventually. I'm not sure about the UI, and whether we should explicitly ask the user or automatically update in the background. We should discuss this later.

ert78gb:
This PR is the first step to resolve #250 issue


.editorconfig
```diff
@@ -9,3 +9,6 @@ indent_style = space
 endofline = lf
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*.json]
+indent_size = 2
```
mondalaci: Is this your personal preference? I'd rather use 4 spaces.

ert78gb: Does not matter for me. My editor setting is reformat the edited file when save. I edited package.json file and the editor replaced all 2 space to 4. 
I thought better if add this setting instead of reformat the package.json file.

I could delete this setting and save the json file in the correct format

mondalaci: Thanks for explaining. Please delete this setting. Other than this, your PR looks good to me.

mondalaci:
@ert78gb Thank you! Let's celebrate your first PR! :champagne:

PR #293, opened 2017-06-01
Add electron-builder packager 

ert78gb:
The PR is the 2nd step to achieve #250 issue. The PR a little bit large and sorry for it, but I could not separate better because the problems comes under the tests.

Content:
- I added the standard-version package to help to manage releases. I think it is better than semantic-versioning because we could manage when create a new release. It is help to create the changelog.md if the contributors use the correct commit formats https://github.com/conventional-changelog/standard-version

- I added electron builder. It will create an installer for Mac, Linux, Win and upload the artifacts into the GitHub release.

- I refactored the tsconfig.json files, because the app builded successfully after the second build. I created 1 tsconfig in the root and build specific versions extends it. 

Possible the PR will not green and not work :), because you have to do the following steps:
- create a github personal token. Set the following scopes: repo(full), admin:org(read), admin:repo_hook(full), user(email). Save the key a secure place, you will never see it again :)
- encrypt the key https://ci.appveyor.com/tools/encrypt
- Replace the encripted value in the appveyor.yml GH_TOKEN ....
- **Important** if the first build will success set the TEST_BUILD variable to false and uncomment line 30 and change the reponame === '....' conditions in the scripts/release.js file

Create a new release:
- `git checkout master; git pull origin master`
- `npm run standard-version`
- `git push --follow-tags origin master`
- wait until all build server create the artifacts
- (pre)publish the release on the github release board

I will send the auto update feature soon


.editorconfig
```diff
@@ -9,3 +9,6 @@ indent_style = space
 endofline = lf
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+[*.json]
+indent_size = 2
```
mondalaci: Here we go again :)

How about setting this explicitly to 4, so that your editor won't try to overwrite it again?

ert78gb: I don't understand why I rebased the branch to the master before I created the PR.

PR #297, opened 2017-06-07
style(editorconfig): Add json rule to editor config

ert78gb:
We aggreed to use 2 indent size in json files. The reason is npm use 2 indent in package.json and this is the defacto

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,5 +9,10 @@
     "removeComments": false,
     "noImplicitAny": true,
     "suppressImplicitAnyIndexErrors": true
-  }
+  },
+  "exclude": [
+    "./node_modules",
+    ".electron/dist",
+    "./dist"
+  ]
 }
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -9,5 +9,10 @@
     "removeComments": false,
     "noImplicitAny": true,
     "suppressImplicitAnyIndexErrors": true
-  }
+  },
+  "exclude": [
```
attilacsanyi: [tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)
_"The "exclude" property defaults to excluding the node_modules, bower_components, jspm_packages and <outDir> directories when not specified."_

ert78gb: Yes, but not really works. The AngularLanguageService read the whole node_modules :( I try to find the good solution. I think the best if we convert the shared folder into angular module.
more than 1 min while the IDEs (intellij, vscode) load the whole project.

@mondalaci @fjozsef @attilacsanyi What do you think about it?

PR #298, opened 2017-06-07
refactor(store): Refactor reducer initialisation

ert78gb:
Refactored the ngrx/store reducer initialisation, because hard to extend the original solution.
Bad practise the object initialisation inside the components / services. The new solution use angular DI everywhere.
Separated the web and electron configuration store.

PR #299, opened 2017-06-07
Add 'New update available' dialog to the electron version

ert78gb:
Suggest to accept this PR after #298


electron/src/store/actions/app-update.action.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const PREFIX = '[app-update] ';
+
+// tslint:disable-next-line:variable-name
+export const ActionTypes = {
+    UPDATE_AVAILABLE: type(PREFIX + 'update available'),
+    UPDATE_APP: type(PREFIX + 'update app'),
+    DO_NOT_UPDATE_APP: type(PREFIX + 'do not update app'),
+    UPDATE_DOWNLOADED: type(PREFIX + 'update downloaded'),
+    UPDATEING: type(PREFIX + 'updating')
```
attilacsanyi: typo ;-) `UPDATING`

ert78gb: thx, fixed

ert78gb:
Pls wait for the review. I am working on the update settings form I will it add during the day

ert78gb:
https://github.com/electron-userland/electron-builder/issues/1709 block me

ert78gb:
The guys solved the problem with electron builder so you could accept this PR.

The usb dependency library will be install as node module. Does need to copy it directly into the vendor folder? I think no, but I don't like to delete until you will not approve.

Issue #261, opened 2017-02-11
Windows set up permission button 


mondalaci:
I'm not sure how much work it takes to set up Windows USB permissions with a click of a button. This should be researched. Worst case scenario, we'll just display the following for the first try:

In order to setup UHK permissions, download [Zadig](http://zadig.akeo.ie/) and use it to install the WinUsb driver for "Ultimate Hacking Keyboard (Interface 0)". (The rest of the interfaces should default to HidUsb.)

fjozsef:
https://msdn.microsoft.com/en-us/library/windows/hardware/ff540174.aspx

mondalaci:
For the first implementation, let's implement a regular setup without the special Windows descriptors. We have to deal with this case anyways, and right now I don't have time to implement the descriptors. Should the descriptors make the USB setup process completely automatic? If so, I assume you won't have to deal with them, and I should be able to test them as I complete them.

fjozsef:
@mondalaci I don't know, I just found this feature, which could be helpful for us. But I don't know how the descriptors work, you should read the article yourself.

mondalaci:
Ok then. Let's not deal with WinUsb descriptors in this issue. I'll implement them in the firmware, test them myself, and let you know if there's anything to do regarding Agent.

mondalaci:
This isn't an issue anymore because hidapi doesn't require privilege escalation on Windows.

Issue #262, opened 2017-02-11
OSX set up permission button


mondalaci:
I'm absolutely clueless about whether it's necessary, and if so, how to set up UHK USB permissions on OSX, so we'll be relying on the ultimate wisdom of Mr. @NejcZdovc.

Related issues are #259, #260, #261.

ert78gb:
#314

ert78gb:
To fix #314 issue we have to solve it.

mondalaci:
hidapi has just been integrated in #300 which should resolve this issue. No privilege escalation is required for Agent on OSX.

PR #300, opened 2017-06-10
Windows set up permission button #261

ert78gb:
It is solve the #261 request. But at the end of the installation I got a warning maybe the driver not installed correctly. But install correctly.
Win 7 32 bit


electron/src/components/privilege-checker/privilege-checker.component.ts
```diff
@@ -81,4 +90,32 @@ export class PrivilegeCheckerComponent {
...
+        };
+        /**
+         * The parameters:
+             - vid: vendor ID
+             - pid: product ID
+             - iface: interface ID
+             - usealldevices: if the device has installed USB driver than overwrite it
+             - noprompt: return at the end of the installation and not waiting for any user command
+         */
+        const command = `"${scriptPath}" --vid 0x1D50 --pid 0x6122 --iface 0 --usealldevices --noprompt`;
```
mondalaci: The VID and PID values are also used elsewhere in the codebase. Please extract them and reference them, so they won't be duplicated.

mondalaci:
Any objections, @fjozsef?

fjozsef:
For me it stays on the: 
"Cannot find your UHK
Please plug it in!" screen and doesn't do anything.

mondalaci:
Your prototype has the old VID and PID @fjozsef, that's the reason why. We'll meet soon and I'll give you a new one.

mondalaci:
I'm not sure if we should pursue this one guys, given that we plan to move from libusb to [node-hid](https://github.com/node-hid/node-hid) based on [HIDAPI](https://github.com/signal11/hidapi) which should work out of the box on Windows.

ert78gb:
ready to PR.
I tested on mac and Win 32 bit

ert78gb:
This PR close: #314, #262,

electron/src/components/privilege-checker/privilege-checker.component.ts
```diff
@@ -23,7 +24,19 @@ import { UhkDeviceService } from '../../services/uhk-device.service';
...
 
-    constructor(private router: Router, private uhkDevice: UhkDeviceService) {
+    private rootDir: string;
+
+    constructor(private router: Router,
+                private uhkDevice: UhkDeviceService,
+                @Inject(LOG_SERVICE) private logService: ILogService) {
+        if (isDev) {
+            this.rootDir = path.resolve(path.join(remote.process.cwd(), remote.process.argv[1]), '..');
+        }
```
mondalaci: `}` should be in the same line as the `else` keyword according to our coding standards.

electron/src/components/privilege-checker/privilege-checker.component.ts
```diff
@@ -82,4 +101,33 @@ export class PrivilegeCheckerComponent {
         return subject.asObservable();
     }
 
+    private setUpPermissionsOnWin(): Observable<void> {
+        const subject = new ReplaySubject<void>();
+        /**
```
mondalaci: Let's rather use a single line comment here, please.

electron/src/services/uhk-device-provider.ts
```diff
@@ -0,0 +1,18 @@
+import { Provider } from '@angular/core';
+
+import { UhkDeviceService } from './uhk-device.service';
+import { UhkHidApiService } from './uhk-hid-api.service';
+
+export function uhkDeviceProvider(): Provider {
+    // HID API officially support MAC, WIN and linux x64 platform
+    // https://github.com/node-hid/node-hid#platform-support
+    if (process.platform === 'darwin' ||
```
mondalaci: Let's rather use hidapi exclusively. It should support every platform we need. This way, we'll be able to test hidapi on every platform and ditch libusb eventually if everything works.

https://github.com/node-hid/node-hid#platform-support : 

> ia32, ARM, MIPSel and PPC64 platforms are known to work but are not currently part of our test or build matrix. ARM v4 and v5 was dropped from Node.js after Node v0.10.

electron/src/components/privilege-checker/privilege-checker.component.ts
```diff
@@ -50,14 +63,19 @@ export class PrivilegeCheckerComponent {
             case 'linux':
                 permissionSetter = this.setUpPermissionsOnLinux();
                 break;
+            case 'win32':
```
mondalaci: Please comment out this branch, and add a comment explaining that hidapi shouldn't need privilege escalation on Windows.

mondalaci:
@fjozsef Any further objections?

electron/src/services/uhk-device-provider.ts
ert78gb: Should I do anything until we will finish the test on all platform?
Currently the app use UhkHidApiService on all platform.
If we drop UhkLibUsbApiService from the code, then I like to remove the uhkDeviceProvider and use directly UhkHidApiService every where.

electron/src/components/privilege-checker/privilege-checker.component.ts
ert78gb: done

electron/src/services/uhk-device-provider.ts
mondalaci: Let's not drop `UhkHidApiService` for the time being, and do this refactoring after a fair amount of testing in a separate PR.

PR #306, opened 2017-06-11
Fix keystroke selection when additional field is given but no scancode

### LINKED ISSUE #302: Unmappable scancodes
The Remap Key of the key action popover gets disabled when I select one of the following scancodes:

* Enter
* Every Navigation key except Home and End
* Every Numpad key except NumLock and Enter
* Delete

fjozsef:
Fixes #302


mondalaci:
Works great, thanks!

PR #307, opened 2017-06-12
Additional media keys with icons

### LINKED ISSUE #303: Broken media scancodes
The media scancodes are broken in various ways:
* When I remap a key to Volume Up or Volume Down, it works, but the new action doesn't show up on the key.
* When clicking on any other media key with the exception of Mute in select2, an exception gets thrown which makes further remappings impossible.

fjozsef:
Fixes #303

### DIFF
--- a/shared/src/components/popover/tab/keypress/scancodes.json
+++ b/shared/src/components/popover/tab/keypress/scancodes.json
@@ -567,31 +567,91 @@
             },
             {
                 "id": "128",
-                "text": "Volume Up"
+                "text": "Volume Up",
+                "additional": {
+                    "type": "media",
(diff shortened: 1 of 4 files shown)


mondalaci:
@fjozsef Just fixed a couple of media scancodes and there's one remaining issue left. When I select Play or Pause, then Remap key, and reopen the key action popover then 00 or 000 gets selected respectively. I noticed that in the code these scancodes reside in the `basicScanCodeTextMap` and `mediaScanCodeTextMap` variables both with 176 and 177 scancodes. There must be some bug in the code that transforms media keystroke {176, 177} to basic keystroke {176, 177}.

Issue #137, opened 2016-10-22
Animate merge/split keyboard action


mondalaci:
The UHK hardware is able to detect the split/merged state of keyboard halves. We're about to blow the socks of people off by animating the split/merge actions realtime. The USB protocol is already [in place](https://github.com/UltimateHackingKeyboard/agent/blob/1063cb2e6eaee9c4a8c8b9abcd2e1f509c34d367/usb/read-merge-sensor.js).

dns2utf8:
Can you show the split line all the time?

mondalaci:
@dns2utf8 Yes, we can, although I'm not sure whether it'd be useful. The vast majority of users will use their UHK in a split configuration and Agent will render the UHK as such. The ones who will use their UHK in a merged configuration won't care where the split line is.

mondalaci:
Superseded by #418 and #450.

PR #311, opened 2017-06-17
Keyboard split/merge animation

fjozsef:
Animation for #137 

It doesn't detect if the uhk is splitted/merged, just animates. Testing with pressing `alt + s`


mondalaci:
Awesome work, thanks @fjozsef!

PR #313, opened 2017-06-17
fix(keymap): Save keymap user configuration after changed

### LINKED ISSUE #308: Keymap modifications aren't preserved after page refresh
1. Change a key action
2. Refresh the page

The previously set key action resets, however it should be restored from the local storage.
Introduced by #298

ert78gb:
fix #308 issue

### DIFF
--- a/shared/src/store/effects/user-config.ts
+++ b/shared/src/store/effects/user-config.ts
@@ -1,16 +1,26 @@
-import { Injectable, Inject } from '@angular/core';
-import { Effect, Actions } from '@ngrx/effects';
+import { Inject, Injectable } from '@angular/core';
+import { Actions, Effect } from '@ngrx/effects';
 import { Observable } from 'rxjs/Observable';
-import { Action } from '@ngrx/store';
+import { Action, Store } from '@ngrx/store';
 
 import 'rxjs/add/operator/switchMap';
 import 'rxjs/add/operator/startWith';
+import 'rxjs/add/operator/withLatestFrom';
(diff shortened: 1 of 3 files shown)


shared/src/store/effects/user-config.ts
```diff
@@ -26,7 +36,20 @@ export class UserConfigEffects {
             return Observable.of(new LoadUserConfigSuccessAction(config));
         });
 
+    @Effect() saveUserConfig$: Observable<Action> = this.actions$
+        .ofType(KeymapActions.ADD, KeymapActions.DUPLICATE, KeymapActions.EDIT_NAME, KeymapActions.EDIT_ABBR,
+            KeymapActions.SET_DEFAULT, KeymapActions.REMOVE, KeymapActions.SAVE_KEY, KeymapActions.CHECK_MACRO,
+            MacroActions.ADD, MacroActions.DUPLICATE, MacroActions.EDIT_NAME, MacroActions.REMOVE, MacroActions.ADD_ACTION,
+            MacroActions.SAVE_ACTION, MacroActions.DELETE_ACTION, MacroActions.REORDER_ACTION)
+        .withLatestFrom(this.store.select(getUserConfiguration))
+        .switchMap(([action, config]) => {
```
fjozsef: You could simply map here.

shared/src/store/index.ts
```diff
@@ -1,3 +1,4 @@
+import { createSelector } from 'reselect';
```
fjozsef: Seems to be unused.

Issue #304, opened 2017-06-11
Add system scancodes


mondalaci:
Please create a new "System" category at the end of the scancode selector of the key action popover and add the following *system* scancodes in it:
* Power down: 0x81
* Sleep: 0x82
* Wake up: 0x83

attilacsanyi:
Is it ok to store those new _system_ scancodes in decimal as others? so
`0x81 = 129, 0x82 = 130, 0x83 = 131` ? @mondalaci

fjozsef:
@attilacsanyi It must be. You can't store hexadecimal numbers in JSON.

attilacsanyi:
Do we have any icons for those system keys? @mondalaci @fjozsef
`Power down` and `Wake up` cannot fit into one key. If we use **Power**, **Sleep** or **Wake** abbreviations then they can fit.

Also due to special logic in `findScancodeOptionByScancode` _(keypress-tab.component.ts)_, we have to define those 3 new system keys before `media` keys in `scancode.json` otherwise, different media keys going to be loaded due to `id === scancode` condition after remap.

fjozsef:
**Sleep**: images/icons/kbd__fn--sleep.svg, just need to be registered,
I can't find an icon for the other two.

attilacsanyi:
Thank you guys, icons were added.
Let me know if I have to use a different one from the [icomoon](https://icomoon.io/#preview-ultimate) pack.

PR #316, opened 2017-06-19
Add new system scan codes

attilacsanyi:


fjozsef:
@attilacsanyi 
Answering to https://github.com/UltimateHackingKeyboard/agent/issues/304#issuecomment-309502674 here, since it is more related to the PR.
The order in scancodes.json shouldn't matter. The type should be `system` and everything should work, if the id is unique. Could you clarify a bit more why is it not working for you?
I would like to keep the ascending order of the ids in the `scancodes.json` if it is possible.

fjozsef:
@attilacsanyi I've fixed the bug in https://github.com/UltimateHackingKeyboard/agent/commit/b7b56ca49196afe3d99d2f0ef2c3ea876f7c6f6c

mondalaci:
@attilacsanyi For me, the system keys get displayed above the media keys. Can you place them below the media keys?

attilacsanyi:
System keys are now below media keys @mondalaci.
Let me know if I need to change any system icon from [icomoon](https://icomoon.io/#preview-ultimate).

mondalaci:
@attilacsanyi Please use the 0844-switch icon for power and the 0346-bell3 icon for wake.

attilacsanyi:
Icon change is pushed @mondalaci

mondalaci:
@attilacsanyi Thanks, looks fine and dandy to me.

Any last words, @fjozsef?

PR #317, opened 2017-06-21
use bootstrap button group instead of separate buttons

attilacsanyi:
Based on [bootstrap 3 button group guideline](http://getbootstrap.com/components/#btn-groups).


mondalaci:
Cool, thanks!

PR #320, opened 2017-06-26
renaming config-serializer upper camel case file names to dash case

attilacsanyi:


mondalaci:
@attilacsanyi Looks like Travis doesn't like your PR. Would you please look into the error?

Issue #142, opened 2016-10-22
Transfer the configuration to/from the UHK


mondalaci:
This is the Agent part of UltimateHackingKeyboard/firmware#3

mondalaci:
This issue is overly general. Will create a more specific issue later.

PR #338, opened 2017-07-06
feat(notification): Add undoable notification, close #318

### LINKED ISSUE #318: Undo deletion box
Let's use the following box upon deleting keymaps and macros:

undo-delete

* The box should slide in from the right and slide back, out of screen upon undoing the deletion or pressing X.
* The "keymap" part of the text should be "macro" for macros.
* When navigating to another page and coming back the box shouldn't be visible anymore.
* The style of the box largely utilizes the [dismissible bootstrap warning alert](http://getbootstrap.com/components/#alerts) with some tweaks. Please make it look like on the picture.
* Edge case: When deleting the last macro the empty macro page will be shown. The box should be visible in this case, too.
* Unlike [displaying errors](#319), the point here is not to cover any UI elements and to seamlessly disappear when not acted upon and moving to another page.

Please check whether we can use use anything from #126. If neither this issue nor #319 can utilize #126 then close #126 and #74.

ert78gb:
PR close #318


mondalaci:
The UI is absolutely perfect, but the semantics is not. I'd expect the undo notification to only appear when a keymap or macro gets deleted.

ert78gb:
Only delete? Do nothing on modification or new event?

mondalaci:
No need to do anything on modifications. Only show the undo notification when deleting keymaps or macros. The idea is that it's very easy to undo the remapping of a key manually, but hard to recreate keymaps or macros.

ert78gb:
It is ready to review.
@fjozsef Please check is this the optimal solution to store the previous state of the user configuration.

@mondalaci Relation of the navigation and the undo dialog visibility is a little bit different than you wrote in the original issue, but I hope you will like to you.

mondalaci:
The issues I noticed:

* "Keymap has been deleted" is displayed for macros.
* When a keymap/macro deletion gets undone, please set the route of the restored keymap/macro.
* When the user switches to another route, please make the undo notification disappear.

ert78gb:
The 3rd point will hard to implement

ert78gb:
I solved the all issues. I hope you thought the navigation same than I implemented

mondalaci:
It's almost good now. Now the problem is that when I undo, not the previous state gets resored but the initial state that was active when I started Agent.

ert78gb:
Could you write, how do you test? I tested with the following steps and worked correctly.
- Delete the user-config in application settings in the browser
- reload browser page to load default config
- navigate to "Bla Bla Bla" macro
- click on delete icon
- [by system ]  Show undo notification and activate "My Address" macro
- navigate to "VIM" keymap
- Delete "VIM" keymap
- [by system ]  Show undo notification and activate "QWERTY" keymap
- click on undo 
- [by system ]  Reload and activate "VIM"

The "Bla Bla Bla" macro is not exists, not reloaded.
 
If I reload the app "Bla Bla Bla" macro is not exists

mondalaci:
The web build seems to work fine, but the electron build always restores the initial state.

ert78gb:
Yes, How is working the app.

Load the data from storage in web app and use default settings in electron, because we not store the change configuration in electron currently, you would like to load it from the uhh directly.

When the delete event occurred the system read the current config from the storage or from the keyboard, but keyboard give back the original storage. It is will work correctly when #142 will implement.

Originally I planned to store the prev state in the app memory, but to much work to refactor user-config reducer to store current and previous config.

ert78gb:
@mondalaci It is working as you like

shared/src/components/undoable-notifier/undoable-notifier.component.ts
```diff
@@ -0,0 +1,52 @@
...
+})
+export class UndoableNotifierComponent implements OnChanges {
+    text: string;
+    undoable: boolean;
+    @Input() notification: Notification;
+    @Output() close = new EventEmitter();
+    @Output() undo = new EventEmitter();
+
+    get slideInOut() {
+        return !this.notification ? 'out' : 'in';
```
fjozsef: just swap in and out

shared/src/store/reducers/app.reducer.ts
```diff
@@ -22,9 +30,46 @@ export function reducer(state = initialState, action: Action) {
...
+            if (newState.navigationCountAfterNotification > 1) {
+                newState.undoableNotification = null;
+            }
+
+            return newState;
+        }
+
+        case ActionTypes.UNDO_LAST_SUCCESS:
+        case ActionTypes.DISMISS_UNDO_NOTIFICATION: {
+            return Object.assign({ ...state }, { undoableNotification: null });
```
fjozsef: Why don't you just use object destructuring without Object.assign?

web/src/app.module.ts
```diff
@@ -87,6 +87,7 @@ import { reducer } from './shared/store/reducers/index';
 import { LogService } from './shared/services/logger.service';
 import { AutoUpdateSettings } from './shared/components/auto-update-settings/auto-update-settings';
 import { angularNotifierConfig } from './shared/models/angular-notifier-config';
+import { UndoableNotifierComponent } from './shared/components/undoable-notifier/undoable-notifier.component';
```
fjozsef: Use index.ts

PR #340, opened 2017-07-08
Clear key actions without mutating the original layers

### LINKED ISSUE #339: Agent throws errors when removing all macros
If I delete every macro, I get the following error:

error1

Then if I hit ctrl+o to transfer the whole configuration to the UHK, I get the following error:

error2

This one is quite critical and a blocker because Eric, our firmware development cannot proceed due to this bug, so please fix it ASAP.

fjozsef:
Fixes #339

### DIFF
--- a/shared/src/store/reducers/user-configuration.ts
+++ b/shared/src/store/reducers/user-configuration.ts
@@ -310,22 +310,45 @@ function generateMacroId(macros: Macro[]) {
     return newId + 1;
 }
 
-function checkExistence(layers: Layer[], property: string, value: any) {
-    const newLayers = layers.map(layer => {
-        const newLayer = new Layer(layer);
-
-        newLayer.modules = layer.modules.map((module: Module) => {
-            module.keyActions.forEach((action: KeyAction, index: number) => {
+function checkExistence(layers: Layer[], property: string, value: any): Layer[] {
+    const keyActionsToClear: {
+        layerIdx: number,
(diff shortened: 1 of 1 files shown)


shared/src/store/reducers/user-configuration.ts
```diff
@@ -310,22 +310,45 @@ function generateMacroId(macros: Macro[]) {
...
-
-        newLayer.modules = layer.modules.map((module: Module) => {
-            module.keyActions.forEach((action: KeyAction, index: number) => {
+function checkExistence(layers: Layer[], property: string, value: any): Layer[] {
+    const keyActionsToClear: {
+        layer: number,
+        module: number,
+        keyAction: number
+    }[] = [];
+    for (let i = 0; i < layers.length; ++i) {
```
mondalaci: Please rename `i`, `j`, and `k` to `layerId`, `moduleId`, and `keyActionId` respectively.

fjozsef: Those are indices, not ids.

mondalaci: Then possibly `layerIdx`, `moduleIdx`, and `keyActionIdx` would be better names. Normally, I support simple cycle variable names but in this case there are 3 of them and I really feel strongly that this would improve readability.

mondalaci:
@fjozsef Thanks so much for fixing this so quickly!

PR #342, opened 2017-07-08
Macro editing improvements

fjozsef:
Cleanup at macros, which removes the unneccesary EditableMacroAction.

It also fixes two smaller macro related bugs:
1. Saving right super to a key action.
2. Opening a hold mouse action doesn't open the hold button sub tab.


mondalaci:
Wasn't even aware about these bugs. So glad to see them fixed! Everything works well, and nicely refactored!

PR #345, opened 2017-07-10
Custom bootstrap tooltip with separate styles and directive 

attilacsanyi:
The following [bootstrap tooltip](http://getbootstrap.com/javascript/#tooltips) properties are supported:

- placement
- title
- custom template with header and [glyphicons](http://glyphicons.com/) icon


ert78gb:
why use jquery in the app it is a generic question. I know the js framework behind the bootstrap is jquery but there are many native angular implementation for bootstrap e.g. https://github.com/ng-bootstrap/ng-bootstrap

attilacsanyi:
Thanks for raising this angular lib topic again @ert78gb.
I mentioned this to @mondalaci sooner, but as we just use the tooltip itself and nothing else from bootstrap, we tried to not add this 3rd party angular lib to further increase the bundle size.

Anyway I also a big fan of great angular libs like [material2](https://material.angular.io/components), [bootstrap](https://ng-bootstrap.github.io/#/components) or [primeng](https://www.primefaces.org/primeng/#/) _(this is the richest lib ever)_.

So the reason behing jQuery is just because of agreement and we already using it ;-)

ert78gb:
I don't think the bundle size or the nr of libraries used should manage the development. If we use big lib and it is well structured the rollup help to hold on low the bundle size

attilacsanyi:
I agree @ert78gb. We should handle adding the most suitable 3rd party lib in different issue imho.

I am really open to use libs which support angular. It up to @mondalaci to define the agent roadmap, but we can really simplify the UI layout and consistency with any of the libs above.
_Most of the used components like tabs, accordion or popups are already among any 3rd party lib components. ;-)_

electron/src/app.module.ts
```diff
@@ -159,7 +159,7 @@ import { angularNotifierConfig } from '../../shared/src/models/angular-notifier-
         MissingDeviceComponent,
         PrivilegeCheckerComponent,
         UhkMessageComponent,
-        CancelableDirective,
+        CancelableDirective, TooltipDirective,
```
mondalaci: Please align this as two lines to be consistent with the rest of the code.

web/src/app.module.ts
```diff
@@ -136,7 +136,7 @@ import { angularNotifierConfig } from '../../shared/src/models/angular-notifier-
         AddOnComponent,
         SettingsComponent,
         KeyboardSliderComponent,
-        CancelableDirective,
+        CancelableDirective, TooltipDirective,
```
mondalaci: Please align this as two lines to be consistent with the rest of the code.

mondalaci:
@attilacsanyi I didn't mean anything fancy, such as the info icon at the top of the tooltip. Please remove it. I'd only like to apply some light custom styles upon the standard bootstrap tooltip.

tooltip

Also, I'd like to be able to specify HTML in the tooltip by using the markup alone. Please see [html=true](http://getbootstrap.com/javascript/#tooltips-options) attribute of of bootstrap tooltip and make sure it works.

mondalaci:
As for getting rid of jQuery, feel free to discuss it in #346.

attilacsanyi:
Simplified bootstrap tooltip with html support (you can use html tags in the `title`).
_Implemented review comments @mondalaci_

sample:
`icon name="question-circle" data-toggle="tooltip" title="This action activates <br> when <b>another key gets pressed</b> while holding this key."></icon>`

mondalaci:
Why did package-lock.json change so substantially, and how can we avoid such huge changes in the future?

attilacsanyi:
I assume if we would have used fixed versions in package.json since we upgraded to npm 5, then we could avoid lock file changes after new npm install in case of adding new lib.

mondalaci:
@attilacsanyi That may be the case. What's your opinion, @fjozsef, @ert78gb?

I'm merging this PR, but let's discuss this last issue regardless.

ert78gb:
I think everybody need to use the same npm version. I suggest use the .nvmrc file manage what is the nodejs version that we use. Most of us use mac or linux to develop agent. [NVM](https://github.com/creationix/nvm) is a great tool to manage node version. And use the default npm version that install by node and not upgrade it separatly.

mondalaci:
@ert78gb Sure thing, feel free to send a PR.

ert78gb:
The file is exists in the repo, just József forgott upgrade the node version from 6.9.4 to 8.x.x I will upgrade soon

PR #355, opened 2017-07-15
Hide UpdateAvailableComponent in AppComponent

### LINKED ISSUE #350: Delete the extra space at the top of the window
There's some extra space at the top of the window which should be deleted. Only Agent-electron is affected.

untitled

fjozsef:
Fixes #350

### DIFF
--- a/electron/src/components/update-available/update-available.component.scss
+++ b/electron/src/components/update-available/update-available.component.scss
@@ -1,4 +1,4 @@
-:host {
+.app-update-available-wrapper {
     display: flex;
     justify-content: center;
     margin: 0.5rem;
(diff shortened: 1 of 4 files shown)


mondalaci:
Works for me.

electron/src/components/update-available/update-available.component.scss
```diff
@@ -1,5 +0,0 @@
-:host {
```
ert78gb: if change :host -> div that solve the problem.

fjozsef: I'd rather not show the whole component, since I couldn't found any reason why we should.

fjozsef: @ert78gb What about the current implementation?

ert78gb: I don't understand the question?

fjozsef: Could you review again? I've made some changes.

PR #356, opened 2017-07-15
Detect uhk without async delay

### LINKED ISSUE #353: Remove initial "Cannot find your UHK" animation
When starting up Agent-electron, I can momentarily see the "Cannot find your UHK" error page which slides out, and then the main UI of Agent slides in. Please let's kill this router animation and make sure that we don't show this error message, only when relevant.

fjozsef:
Fixes #353


mondalaci:
Now I cannot see the "UHK is not connected" message, but it'd be great to also kill the initial sliding animation. It'd make Agent feel quicker to load.

PR #359, opened 2017-07-16
feat(app): Show add-on menu if start app with --addons arg

### LINKED ISSUE #351: Don't display the add-on modules by default
The add-on modules will only be implemented later until which point we shouldn't display them in the menu. Please hide them by default and only show them when the `--addons` command line argument is present.

ert78gb:
Close #351

### DIFF
--- a/electron/src/app.module.ts
+++ b/electron/src/app.module.ts
@@ -107,6 +107,7 @@ import { reducer } from './store';
 import { AutoUpdateSettings } from './shared/components/auto-update-settings/auto-update-settings';
 import { angularNotifierConfig } from './shared/models/angular-notifier-config';
 import { UhkHeader } from './shared/components/uhk-header/uhk-header';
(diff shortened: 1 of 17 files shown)


mondalaci:
Please resolve the conflict and I'll give it a try.

ert78gb:
conflict resolved

mondalaci:
I've tried it and `npm run electron --addons` doesn't reveal the add-ons menu for me. Any ideas why?

ert78gb:
When use npm run you have to use -- to add command line argument.

npm run electron -- --addons

mondalaci:
All good! The only reason I'm not merging this is that it's dependent on #338, so let's merge that first.

ert78gb:
You could accept/merge this PR. #338 have to rebase and develop other issues.
I created the dependency linking but after the implementation is not relevant.

ert78gb:
But now see I have to rebase this branch too

PR #360, opened 2017-07-16
feat(keymap): Keymap abbr should be only number or ASCII char

### LINKED ISSUE #341: Enforce further keymap abbreviation rules
1. Currently, it's possible to type any character into the abbreviation field. Going forward, make sure that only English letters and numbers are allowed.
2. Currently, only 3 letter long abbreviation names are allowed. Going forward, make sure that 1-3 character long strings are allowed.

ert78gb:
Close #341


eltang:
We have support for "0" in the firmware as well.

ert78gb:
Ok I will add

PR #362, opened 2017-07-17
fix(keymap): allow 0 in keymap abbr when rename

ert78gb:
in #360 mentioned the generateAbbr method not contains 0 and 1 and it is a bug or not.
I think it is not bug because a function is calculate the next abbreviation and if we allow the 0 and 1 the the next calculated value will be xx0. So I added the 0, but I think I have to delete the 1 from the string and the new abbreviation will xx2 instead if xx0.
@mondalaci What is your opinion?


mondalaci:
Good catch! "0" should definitely be accepted as a valid character.

Issue #364, opened 2017-07-17
Add tooltips to action icons


mondalaci:
Currently, the delete and duplicate icons of keymaps and macros have tooltips. Please also make sure that the tooltips of the download keymap and the default keymap (star) icons look the same.

Note that the star icon has enabled and disabled state and it's also present in the side menu. All of these 3 icons must have tooltips.

Make sure that every tooltip points downwards, not upwards which is currently the case.

I'd also like to get the style of the tooltip changed, so that it has the current dark gray outline but white background and black text.

ert78gb:
What are the text of tooltips?

ert78gb:
Delete
Duplicate
Default ?

mondalaci:
The icons already have Alt text. You can see them if you hover over them. These Alt texts should be put into the respective tooltips.

PR #366, opened 2017-07-18
fix(keymap): Add tooltips to action icons

ert78gb:
Close: #364 

Details:
- Refactored the colour definitions of tooltip. Extracted the colours in global-styles
- Add additional properties to the tooltip-directive.


mondalaci:
Please make the arrows of the tooltips white with dark gray outline. This is how they currently look like:

tooltip-arrow

I'd like to make them look like the arrow of the key action popover:

right-tooltip

Speaking of the key action popover, would you please make its tooltip point downwards instead of upwards?

tooltip-qm

ert78gb:
I solved both of your issue

PR #373, opened 2017-07-25
Fix key capture on Firefox

### LINKED ISSUE #367: Impossible to reassign any key using middle click on FF
That's the behaviour:
When you middle click a key appears on top the record key.
Any key pressed after that do nothing.
When you middle click the key again, the record symbol dissapears.

If the user middle clicks two keys, both will have on top a record key.
redkeys

I've tested the issue on v39 (32bit), but also on the newests versions: v54 (32bit) and v54 (64bit).
Same issue.

laxu:
This fixes #367 by adding tabindex attribute to key svg element so Firefox can properly focus the key. Otherwise focus, focusout events do not get fired and using mouse middle click to change key quickly does not work in Firefox. 

Only Firefox v52+ support focusout event so this will still not work correctly on older FF versions than that.

Also fixes small typo in onKeyUp (was "onKeyUpe").

### DIFF
--- a/shared/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
+++ b/shared/src/components/svg/keys/svg-keyboard-key/svg-keyboard-key.component.ts
@@ -118,7 +118,7 @@ export class SvgKeyboardKeyComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     @HostListener('keyup', ['$event'])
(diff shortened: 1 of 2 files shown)


mondalaci:
Thanks so much for your contribution, @laxu! I'm waiting for the team to review just to be sure, although it'll hardly get any cleaner than your fix.

Issue #280, opened 2017-03-26
Option to use JIT/AOT build modes


fjozsef:
Currently only building in JIT mode is available, we should make it possible to use AOT (and switch between those two transparently) since it has a lot of benefits (e.g. smaller and faster code).

Probably we will use: https://www.npmjs.com/package/@ngtools/webpack

fjozsef:
Blocked since `@ngtools/webpack` is unable to process symbolic links.

attilacsanyi:
Would be great to use [angular-cli](https://github.com/angular/angular-cli/wiki) which is use **AOT by default**.
_Electron part might be an issue in this case._
Angular-cli has [eject](https://github.com/angular/angular-cli/wiki/eject) command, which can be used for comparison to the current agent webpack config.

mondalaci:
@attilacsanyi I'd like to see angular-cli introduced, although chances are AOT won't work right now due to the issue mentioned by @fjozsef.

mondalaci:
How about this one, @ert78gb? According to my knowledge AOT is not used yet due to circular dependencies. How long would it take to make it happen?

ert78gb:
I will check it on next week and I will give a response.

PR #394, opened 2017-08-25
build: Configure AOT

ert78gb:
close: #280


mondalaci:
The build succeeds for me, but the circular dependency warnings are still there. Wondering whether AOT works despite of circular dependencies.

ert78gb:
I didn't know well the circular dependencies not block the AOT. AOT was blocked by angular linked module dependencies

ert78gb:
I have to fix the environment settings for appveyor build. few min and I will fix it

Issue #141, opened 2016-10-22
Validate macro actions


mondalaci:
The Save button of Macro actions must be disabled in the following cases:

* Type text macro action: empty text field
* Key macro action: Scancode None and no modifiers selected
* Mouse action: Move pointer and Scroll: Any of the X, Y fields are empty
* Mouse action: Click and Hold and Release button: None of the buttons are pressed
* Delay: empty or zero delay field

attilacsanyi:
I implemented all review comments. Thanks @ert78gb. Would you mind to check another time @mondalaci ;-)

mondalaci:
Done in #395.

PR #395, opened 2017-08-25
Issue - 141 - Validate macro actions (#141)

attilacsanyi:
Not able to handle auto disable over the save button, just prevent the saving.


mondalaci:
Why can't you disable the save button?

attilacsanyi:
Because the save button is on a top level `macro-action-editor.component` and the changes are in the children components. If I set to disabled then no chance to set back to active again as children `macro` components not emit their state.

_I try to avoid a redesign where all children emit its change event so parent can listen to and update its state. The easy validation functionality was not kept in mind in the beginning._

mondalaci:
I understand, but the user experience is not as great as if disabled was used. Please use disabled and take your time to redesign.

attilacsanyi:
Refactored validation logic @mondalaci, so now the `Save` button immediately reflect the validation state of each macro tab component. I have tested tons of cases and fixed some existing bugs as well. I hope you cannot find any missed validation cases ;-) Thanks for the testing in advance.

mondalaci:
Thanks, this is so much better now! Please also make both the X and Y coordinates of the mouse move and mouse scroll actions mandatory, otherwise the following can happen.

untitled

attilacsanyi:
Thanks for the testing, should work in this case as well by design, will check it again @mondalaci ;-)

attilacsanyi:
Hmm, this is working for me could you please provide me repro steps? @mondalaci 
By design i implemented this `"Any of the X, Y fields are empty"`. Thanks

attilacsanyi:
Should be ok for now, so disabled if `"Any of the X, Y fields are empty"` in case of move or scroll @mondalaci

mondalaci:
Thanks, looks good to me! Waiting for @ert78gb or @fjozsef to review.

packages/uhk-web/src/app/components/macro/action-editor/tab/delay/macro-delay.component.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from '@angular/core';
 
 import { DelayMacroAction } from '../../../../../config-serializer/config-items/macro-action';
+import { MacroBaseComponent } from './../macro-base.component';
```
ert78gb: pls remove ./ from begining of the import

packages/uhk-web/src/app/components/macro/action-editor/tab/mouse/macro-mouse.component.ts
```diff
@@ -7,6 +7,7 @@ import {
     MacroSubAction
 } from '../../../../../config-serializer/config-items/macro-action';
 import { Tab } from '../../../../popover/tab';
+import { MacroBaseComponent } from './../macro-base.component';
```
ert78gb: pls remove ./ from beginning of the import

packages/uhk-web/src/app/components/macro/action-editor/tab/key/macro-key.component.ts
```diff
@@ -71,4 +73,9 @@ export class MacroKeyTabComponent implements OnInit {
         return keyMacroAction;
     }
 
+    isMacroValid = () => {
+        const keyMacroAction = this.getKeyMacroAction();
+        return keyMacroAction.scancode !== undefined || keyMacroAction.modifierMask !== undefined;
```
ert78gb: pls use the following condition `!!keyMacroAction.scancode || !!keyMacroAction.modifierMask;`

@mondalaci If I read well the specification
>Key macro action: Scancode None and no modifiers selected

then the condition is && instead of ||

packages/uhk-web/src/app/components/macro/action-editor/tab/mouse/macro-mouse.component.ts
```diff
@@ -120,4 +126,18 @@ export class MacroMouseTabComponent implements OnInit {
         return TabName.Move;
     }
 
+    isMacroValid = () => {
+        switch (this.macroAction.constructor) {
+            case MoveMouseMacroAction:
+            case ScrollMouseMacroAction:
+                const { x, y } = this.macroAction as MoveMouseMacroAction;
+                return x !== undefined && x !== null && y !== undefined && y !== null;
```
ert78gb: pls use `return !!x && !!y;`

packages/uhk-web/src/app/components/macro/action-editor/tab/mouse/macro-mouse.component.ts
```diff
@@ -120,4 +126,18 @@ export class MacroMouseTabComponent implements OnInit {
...
+            case MouseButtonMacroAction:
+                const { mouseButtonsMask } = this.macroAction as MouseButtonMacroAction;
+                return mouseButtonsMask !== undefined && mouseButtonsMask !== 0;
```
ert78gb: pls use `return !!mouseButtonsMask;`

packages/uhk-web/src/app/components/macro/action-editor/tab/text/macro-text.component.ts
```diff
@@ -9,31 +9,39 @@ import {
...
     ngAfterViewInit() {
         this.renderer.invokeElementMethod(this.input.nativeElement, 'focus');
     }
 
     onTextChange() {
+        this.init();
         this.macroAction.text = this.input.nativeElement.value;
     }
 
+    isMacroValid = () => this.input.nativeElement.value !== undefined && this.input.nativeElement.value.length > 0;
```
ert78gb: pls use `!!this.input.nativeElement.value;`

packages/uhk-web/src/app/components/macro/action-editor/tab/key/macro-key.component.ts
mondalaci: @ert78gb It must be disabled if scancode is none and no modifiers is selected. Thus, it's be valid (not disabled) if scancode is not none or any modifiers are selected. The current implementation is correct.

packages/uhk-web/src/app/components/macro/action-editor/tab/mouse/macro-mouse.component.ts
attilacsanyi: better to keep the longer version as your solution no allow to set `0` to x or y which should be valid. @ert78gb

packages/uhk-web/src/app/components/macro/action-editor/tab/mouse/macro-mouse.component.ts
attilacsanyi: thx.

packages/uhk-web/src/app/components/macro/action-editor/tab/key/macro-key.component.ts
attilacsanyi: thank you guys.

PR #397, opened 2017-08-26
Issue 388 - Device node

attilacsanyi:
Create a new top level device node with provided usb stick icon and define a simple device setting page.


mondalaci:
1. For some reason the USB icon doesn't show for me.
2. Vertical spacing around the Keymaps and Macros menu items has been enlarged. This should not change compared to the previous version.

Before:

untitled2

After:

untitled

attilacsanyi:
Hmm, really weird, icon ok for me. Which browser do you use?
My goal was just to add extra padding for the existing elements, will double-check it again. @mondalaci

mondalaci:
I've only tried the electron build.

attilacsanyi:
Ahh, I just tested `uhk-web` :) So let me test the electron build as well. @mondalaci

ert78gb:
if you run `npm run electron` it is start the electron app. the electron engine is a chrome engine but / maybe a few feature missing.

mondalaci:
The USB icon still doesn't show in either build. Looks like a path issue.

untitled

attilacsanyi:
Will try it, but for electron build under mac I have to test with the keyboard first, which is going to be my first try. I should try with windows first, but that is my other machine :). Let you know once tested. @mondalaci

mondalaci:
No need to build the electron version, because I've just tested the web version and I can confirm the large spacing and the missing icon issues. These should be fixed.

attilacsanyi:
Hmmm weird, rendered correctly for me after merging from master
image

mondalaci:
@ert78gb Would you also test this? Still can't see the USB icon.

ert78gb:
I am checking ...

Issue #377, opened 2017-07-30
Save to keyboard button


mondalaci:
Whenever the user updates hid/her configuration, a button should slide in from the bottom and show up in the right bottom corner:

![untitled](https://user-images.githubusercontent.com/300402/28756817-308c99f2-7576-11e7-9d8e-dce5b95977d4.png)

Upon pressing the button it should trigger the same transfer that is now triggered when pressing ctrl+o and update its content to display "[animated progress spinner] Saving", and when completing the transfer it should slide out downwards.

When changing the content of the button, don't change its size of and always center align its content.

Please unbind ctrl+u, ctrl+i and ctrl+o and clean up related unused code.

I've also observed that when I transfer the config using ctrl+o, then make another keymap default, the configuration automatically transferred. Please make sure that the config never gets automatically transferred.

ert78gb:
> When changing the content of the button, don't change its size

What is the maximum with of the button?

mondalaci:
The maximum width of the button is the width that is necessary to hold the "Save to keyboard" text. The "[spinner] Saving" text will be narrower than that, so the button should keep its width.

PR #402, opened 2017-08-29
feat(device): Add 'Save to keyboard' button

ert78gb:
Created a 'Progress Button' that have 2 state in progress or not.
Able to set different text for different state:
- baseText for normal state
- progressText for in progress state

close: #377


ert78gb:
@mondalaci When saving data to the keyboard is success should show the Success notification on the right top?

mondalaci:
@ert78gb Nope, please only show notifications upon errors.

ert78gb:
@mondalaci removed the success notification

mondalaci:
The first save succeeds and works perfectly.

Unfortunately, the second save throws the following error:

untitled

Minor UI issues to be fixed:
1. Don't make the notification time out and disappear. Make it stay there until the user closes it. Make all notifications behave this way, not just this one.
2. When clicking on the Save to keyboard button, wait until the save is in progress. Then display "Saved!" and only then make the button leave the screen.

PR #409, opened 2017-09-08
Remove excess space above the UHK

attilacsanyi:


mondalaci:
The Electron build displays the following now:

untitled2

attilacsanyi:
Ahh thats why was the width set to 95% which I removed as on the web has no issue. Revert back and should be fine. @mondalaci

mondalaci:
@ert78gb Would you also please take a look at this and report your findings? The bottom half of the keyboard is still cut off for me.

attilacsanyi:
Just realised that it is an issue only on big screens > 1500px wide cases. On my mac it was not visible sorry guys. We need to test several responsive cases from this point as agent is very sensitive @mondalaci

attilacsanyi:
I found it this is because the current keyboard slider height was set to fix `500px` so as you increase the browser width the keyboard cannot fit into this 500px height space. Try to make it flexible.

Issue #412, opened 2017-09-11
Read user configuration


mondalaci:
Right after detecting the UHK, please read its configuration and use it to initialize the state of Agent. Same as #403 but backwards.

Issue #410, opened 2017-09-10
Make configuration serialization responsive


mondalaci:
We'll have to move the serialization of the configuration into a web worker, shared worker, or to the main process in order to make the progress indicator responsive.

Issue #296, opened 2017-06-06
Loading screen


mondalaci:
A second or two passes upon initializing Agent even on my Core i7 machine, and possibly significantly more on slower machines. It'd be nice to show some content in the meantime, so that the user would have confidence that everything is on track.

Please consider the content of the "Cannot find your UHK" Agent screen:

![agent](https://user-images.githubusercontent.com/300402/26851803-3e546f04-4b0c-11e7-820c-09a45b2f62e9.png)

I'd like the same layout for the loading screen with the following changes:

1. Replace the "Cannot find your UHK" text with "Loading Agent..."
2. Replace the "Please plug it in!" text with "Hang tight!"
3. Animate the Agent SVG icon to give a visual indication to the user that loading is in progress. In specific, rotate the Agent icon clockwise for a sec, then wait a sec, then repeat. The rotation should accelerate, then decelerate. All this should be implemented purely in CSS without JavaScript.

mondalaci:
I found a relevant Stack Overflow issue: https://stackoverflow.com/questions/35243443/pre-bootstrap-loading-screen-for-angular2

fjozsef:
Having AOT #280 would significantly reduce the bootstrap time, making this feature unnecessary.

mondalaci:
Excellent point! Let's make this issue depedent on #280 and only proceed with this one if it makes sense.

mondalaci:
Loading Agent feels fairly snappy now, so closing this issue.

PR #413, opened 2017-09-12
feat(device): Read user config from eeprom

ert78gb:
I think we have to implement the loading screen #296 , because reading the whole eeprom is 5-6 sec. Until the end of reading white screen visible for the user.

close: #412, close: #410

### DIFF
--- a/packages/uhk-agent/src/services/device.service.ts
+++ b/packages/uhk-agent/src/services/device.service.ts
@@ -147,7 +147,6 @@ export class DeviceService {
             await this.writeUserConfigToEeprom();
 
             response.success = true;
-            this.logService.info('transferring finished');
         }
         catch (error) {
             this.logService.error('[DeviceService] Transferring error', error);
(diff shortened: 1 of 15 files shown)


ert78gb:
I found a bug. It is not ready to review yet

ert78gb:
@mondalaci ready to test

mondalaci:
I cannot seem to be able to remap any keys now. When I remap a key it's displayed the same way as before.

ert78gb:
Could you write step by step how do you test?

mondalaci:
1. Start up Agent-electron.
2. Remap any key, for example try to remap the "A" key of the base layer of the QWERTY keymap to "B".
3. At this point, I'd expect the "A" letter to change to "B" but it's still shows "A".

(Strangely, when I click on the key and the popover appears, I can see "B" in select2 . So maybe it's changed in the data model, it's just not shown.)

Issue #415, opened 2017-09-15
Read and write the hardware configuration area


mondalaci:
The [hardware configuration area](https://github.com/UltimateHackingKeyboard/agent/blob/d621b1e5e6ee6bf3559ab2f51e4b82dc825dffd9/packages/uhk-web/src/app/config-serializer/config-items/hardware-configuration.ts) (HCA) is a dedicated 64 byte long region of the EEPROM. It's read and written similarly that the user configuration area (UCA) that stores keymaps, macros, and other objects.

The HCA needs to be read and written just like the UCA. The difference is that Agent only needs to read it, and a script should be created to write it. The script will only be used in the factory to initialize the HCA with hardware specific information that won't change in the future.

The [{Read,Write}HardwareConfiguration](https://github.com/UltimateHackingKeyboard/firmware/blob/2f90e40c921201e1f6e9ba50cc27ea0630300e45/right/src/usb_protocol_handler.h#L21-L22) USB commands work just like their *UserConfiguration counterparts, and the size of the HCA can be [queried](https://github.com/UltimateHackingKeyboard/agent/blob/d621b1e5e6ee6bf3559ab2f51e4b82dc825dffd9/packages/usb/read-config.js#L17) just like for the UCA.

Most of the fields of the HCA are constants:

signature = "UHK"
dataModelVersion = 0
hardwareId = 0
uuid = random generated 32 bit uint value
brandId = 0
isIso = see below
hasBacklighting = 0

`usb/write-hca.js` should be created which expects a single argument: `ansi` or `iso`. `isIso` should be true for `iso`, and false for `ansi`. Even though the HCA data structure is smaller than 64 bytes, Agent should fill out the remaining space with zero bytes.

Agent should display the ISO layout if isIso === true, otherwise display the ANSI layout. There are separate SVG (now XML) files for these layouts, so Agent should simply use the relevant one.

PR #423, opened 2017-09-24
feat(config): Read / write hardware configuration area

ert78gb:
Summary:
- created a write-hca.js script that write user configuration to the keyboard
- Moved the config-serializer 'folder' into the uhk-common package
- Created a new uhk-usb package, this package contains the core communications protocol to the UHK Device. 'write-hca.js' use this module. In the future suggested rewrite the whole usb package to use this module instead of uhk.js file.
- When device is connected load the user and hardware configurations from the UHK Device, so renamed the LoadUserConfigurationAction to LoadConfigurationAction
- When load configuration is faild show the generic error notification with the error message.
close: #415


mondalaci:
I'd like the following to be implemented:
* Agent always shows the ANSI layout regardless of the layout set by `write-hca.js`. Agent should show the actual layout.
* I'd like to see "USB[T] ..." comments for every transfer type in the command line, too. This should be part of the shared USB library so that it'd be reused.
* `write-hca.js` should have executable permision.

packages/uhk-web/src/app/components/svg/keyboard/svg-keyboard.component.ts
```diff
@@ -45,13 +47,17 @@ export class SvgKeyboardComponent implements OnInit {
     }
 
     ngOnInit() {
-        this.modules = this.svgModuleProvider.getSvgModules();
+        this.setModuls();
```
mondalaci: Should be `setModules`

PR #427, opened 2017-09-27
Add captured press keystroke for macros (closes #324)

### LINKED ISSUE #324: Add captured keystroke for macros
The "Add captured keystroke" button at the end of macros should behave just like the "Capture keystroke" button of the key action popover, and add a related macro action to the macro.

attilacsanyi:
Try to reuse existing `capture-keystroke` component and render as a bootstrap link under macro list component. @mondalaci


mondalaci:
I'd like to following UI improvements to be implemented:
* Make it work not only when clicking on the link, but on the whole surface of the button.
* Don't highlight the link of the button on hover, just like the button next to it.

attilacsanyi:
You can retest @mondalaci, nice catch.
I implemented this with an advanced angular technic about providing custom template for an existing component.

mondalaci:
Thanks! Can you also make the text of the "Add macro action" button go just as dark as the "Add captured keystroke" button when hovered?

Of all the buttons, only the "Add macro action" button seems to be inconsistent in this respect. Its text color is unchanged upon hover unlike the other buttons that get darker when hovered.

mondalaci:
All good! Waiting for @ert78gb to review.

packages/uhk-web/src/styles/common.scss
```diff
@@ -0,0 +1,10 @@
+%btn-link {
```
ert78gb: I am not css/scss wizard but I like to use the classic cascading, so I prefer to use class selector instead of placeholder. And delete the extending from capture-keystroke-button.component.scss, macro-list.component.scss files

attilacsanyi: @ert78gb thanks for the feedback I chose the placeholder scss feature as we need exactly the same styles in two totally different component.
Any better way of doing this without copy-paste all css class properties on both places?
_I like the class selector approach, but in this case I have to define a global one, is this what you want here?_

Issue #439, opened 2017-10-10
Wait for the EEPROM to be read until reading the configuration


mondalaci:
Simply the getKeyboardState USB call have to be polled before starting to read the hardware and user configuration.

PR #443, opened 2017-10-13
feat(device): Wait for read user/hardware config until device is ready

ert78gb:
close: #439 

I can not test it because from the last firmware update I can not connect to the device


mondalaci:
This seems to behave just like before. Let me try to fix the firmware soon.

Issue #446, opened 2017-10-13
Make "reset user configuration" safer and nicer


mondalaci:
Instead of the current "Reset user configuration" link, please use a red bootstrap button containing the same text.

Write "including all your keymaps and macros" rightwards of the button.

For confirmation purposes, please use the [Bootstrap confirmation plugin](http://bootstrap-confirmation.js.org/). The default confirmation popup will work for us. Make the popup close when the user clicks outside of it.

PR #449, opened 2017-10-14
feat(device): Add confirmation question before reset the configuration

ert78gb:
close: #446


ert78gb:
the used component https://mattlewis92.github.io/angular-confirmation-popover/

PR #468, opened 2017-10-18
feat(agent): remove routing animation from keymaps

### LINKED ISSUE #432: Routing animation fixes
1. When starting up Agent-electron without the UHK plugged in, it first momentarily displays the main application screen with the keyboard, then transitions to the "UHK not found" screen. Please don't show the main application screen in this case, not even for a moment. Always show the relevant screen immediately.

2. Animations are used when switching routes. For example, when switching from a macro to to a keymap the keyboard slides in from the left slide. Let's not use animations in such cases at all.

ert78gb:
close #432


mondalaci:
For some reason, the keyboard is left-aligned when first displayed:

untitled

When another layer gets chosen, it gets center-aligned as expected.

When switching to a non-keymap page, and then back to a keymap page, it gets left-aligned again.

Issue #493, opened 2017-11-09
Secondary role menu item gets stuck


mondalaci:
In the key action popover, when I choose a key of a a secondary role other than none, then close the popover and reopen if for another key of a secondary role of none, I can see the secondary role of the previous key instead of none.

The secondary role should always show the value of the current key.

mondalaci:
Done by #493.

PR #496, opened 2017-11-12
fix(keymap): Secondary role show 'None' value if not set

ert78gb:
close: #493


packages/uhk-web/src/app/components/popover/tab/keypress/keypress-tab.component.ts
```diff
@@ -104,6 +104,9 @@ export class KeypressTabComponent extends Tab implements OnChanges {
         if (keystrokeAction.secondaryRoleAction !== undefined) {
             this.selectedSecondaryRoleIndex = this.mapper.modifierMapper(keystrokeAction.secondaryRoleAction);
         }
+        else {
```
mondalaci: This seems to work well, but would you please write `} else {` in the same line according to our coding standards?

ert78gb: done

Issue #465, opened 2017-10-18
Undefined in URL upon multiple keymap duplication


mondalaci:
When duplicating the QWERTY keymap many times and arriving to "QWERTY (35)", the URL goes "http://localhost/uhk/agent/packages/uhk-web/dist/#/keymap/QWundefined"

ert78gb:
What is the expected behavior? The counter running out from available range. The software use `23456789ABCDEFGHIJKLMNOPQRSTUVWXYZ` values to increment the last character.
If the software run out of the last character range allow to increment the last before character.
After the QWZ the new abbreviation will QX1. Is it acceptable?

mondalaci:
Yes, the suggested behavior would work great.

PR #497, opened 2017-11-12
fix(keymap): Handle more than 35 same keymap abbreviation

ert78gb:
close: #465


ert78gb:
This modification is not handle if the first character step over the character range

PR #501, opened 2017-11-30
Add a number of properties to the user config that are related to mou…

mondalaci:
…se movement, LED brightness, data model versioning, and double tap switch layer timeout. Update firmware related version numbers.


mondalaci:
@ert78gb Now Agent references the newly released 4.0.0 firmware which should work with this PR, but the loading indicator of Agent keeps spinning for some reason.

I'd also like to use the @assertUInt8 annotations (which are now commented out) but they cause a build error.

PR #503, opened 2017-12-02
fix(device): fix IpcRenderer empty data messages

ert78gb:
If IpcRenderer send null / undefined data as argument then IpcMain parse
it as empty array and not a NULL object


packages/uhk-web/src/app/services/device-renderer.service.ts
```diff
@@ -35,7 +35,12 @@ export class DeviceRendererService {
     }
 
     updateFirmware(data?: Array<number>): void {
-        this.ipcRenderer.send(IpcEvents.device.updateFirmware, JSON.stringify(data));
+        if (data) {
+            this.ipcRenderer.send(IpcEvents.device.updateFirmware, JSON.stringify(data));
+        }
+        else {
```
mondalaci: `} else {` should be in the same line according to our coding standards.

Issue #466, opened 2017-10-18
LED brightness options


mondalaci:
Create a new page titled "LED brightness" and put its menu item below "Device -> Mouse speed".

The page should contain the following content:

> LED display icon and layer texts brightness
> [slider of 0-255 values]
> LED display alphanumeric segments brightness
> [slider of 0-255 values]
> Key backlight brightness
> [slider of 0-255 values]

Please use the same slider as in #458.

Use the following properties of UserConfiguration:

* iconsAndLayerTextsBrightness
* alphanumericSegmentsBrightness
* keyBacklightBrightness

mondalaci:
Resolved by #520.

Issue #458, opened 2017-10-15
Add mouse speed page


mondalaci:
Create a new page titled "Mouse speed". Use the fa-slider icon before its title, and redirect to it upon clicking the Device -> Mouse speed menu item.

This page features a lot of sliders. Please use [ng2-nouislider](http://tb.github.io/ng2-nouislider/) for all the sliders. Make the slider tooltip always show. If possible, show the unit of the value (for example px/s²) in the tooltip, for example "8 px/s²". If it's not possible easily, then just display the unit as described below, for example "Acceleration (px/s²): above the individual sliders". Make the sliders occupy the whole width available. Make the range show below every slider.

The page should contain the following content:

> The following options affect the speed of the mouse when being controlled by mouse keys.
>
> Mouse pointer speed
>
> Initial speed (px/s):
> [slider]
> Acceleration (px/s²):
> [slider]
> Decelerated speed (px/s):
> [slider]
> Base speed (px/s):
> [slider]
> Accelerated speed (px/s):
> [slider]
>
> Mouse scroll speed
>
> Initial speed (pulse/s):
> [slider]
> Acceleration (pulse/s²):
> [slider]
> Decelerated speed (pulse/s):
> [slider]
> Base speed (pulse/s):
> [slider]
> Accelerated speed (pulse/s):
> [slider]

The actual values of the user configuration are uint8 types, and each value can be 1-255. (0 is not allowed.)

Use a 25x multiplier for displaying move ranges, so they should show 25 as the minimum and 6375 as the maximum.

Use a 1x multiplier for displaying scroll ranges, so they should show 1 as the minimum and 255 as the maximum.

Use the following properties of UserConfiguration:

* mouseMoveInitialSpeed
* mouseMoveAcceleration
* mouseMoveDeceleratedSpeed
* mouseMoveBaseSpeed
* mouseMoveAcceleratedSpeed

* mouseScrollInitialSpeed
* mouseScrollAcceleration
* mouseScrollDeceleratedSpeed
* mouseScrollBaseSpeed
* mouseScrollAcceleratedSpeed

richrd:
Hi! I'm curious to why `ng2-nouislider` should specifically be used? Not against it but just wondering, since html5 already has input `type="range`.

Another question: are there any issues that I could volunteer to work on?
As I've ordered a UHK and I have Angular 2/4 experience this would be a fun project to contribute to.

Thanks!

mondalaci:
[...]

PR #520, opened 2017-12-21
LED brightness UI

laxu:
Adds LED brightness UI as per #466. I added input boxes that allow showing current value and updating the value by typing it in if desired.

uhk-ledbrightnesspage

This is not yet hooked up to load the values from UserConfiguration as I could use help on the correct way to do that and how to trigger the "Save to keyboard" button if the settings are changed.


mondalaci:
So awesome to see this come to fruition!

I'd like to test this, but hit the following upon running `npm run build`:

ERROR in ng:///home/laci/projects/uhk/agent/packages/uhk-web/src/app/components/device/led-brightness/led-brightness.component.html (18,17): Property 'onChange' does not exist on type 'LEDBrightnessComponent'.
ERROR in ng:///home/laci/projects/uhk/agent/packages/uhk-web/src/app/components/device/led-brightness/led-brightness.component.html (34,17): Property 'onChange' does not exist on type 'LEDBrightnessComponent'.
ERROR in ng:///home/laci/projects/uhk/agent/packages/uhk-web/src/app/components/device/led-brightness/led-brightness.component.html (50,17): Property 'onChange' does not exist on type 'LEDBrightnessComponent'.

laxu:
Yeah forgot an unimplemented onChange there, now it should build.

mondalaci:
Now it builds, but for some reason the sliders don't show up for me. I cloned the repo, switched to your branch, and built it from strach.

screenshot from 2017-12-21 19-51-49

laxu:
Hmm, seems like the nouislider is not correctly installed. Did you run npm install?

laxu:
It seems Electron has issues loading the stylesheet for nouislider.

laxu:
Fixed! The nouislider files were just missing from webpack config.

mondalaci:
Awesome! I can confirm that it's working now. I think it'd be worth to add pips and the current value displayed as hover, both as percentages accurate to 0.1%. The number inputs could be removed, and it'd also be useful to be able to finely adjust the sliders with the left arrow and right arrow keys when focused. I think these changes would result in a cleaner UI.

laxu:
I tried it with tooltips but at least the default option shows the tooltip all the time and it overlaps the label and to me was less usable than the input way. The inputs stay in one place in a row so they are faster to read than slider tooltips that could be all over the place.
 
I added adjusting the sliders with arrow keys.

mondalaci:
Thanks for the arrow key fix!

The problem with the current UI is that it will hurt consistency.

#458 will feature pretty much the same UI, but the big difference is that it'll only be possible to adjust various numbers in larger steps. Imagine a multiplier of 25 for example, only being able to set 25, 50, 75, and so on. Then the numeric input fields won't be intuitive at all. That's why it'd be a good idea to get rid of them for the sake of consistency.

As for the hover, we could make this work by adding some top margin.

mondalaci:
Thanks for the nice screenshot!

While looking at the picture, it seems to me that the tooltips abuse vertical space quite badly. I think that all things considered, it'd be worth to disable them.

How about moving the values not leftwards but rightwards, and displaying them as plain text instead of input boxes?

Would you also make the texts non-bold? It'd be more consistent with the rest of Agent.

mondalaci:
This looks awesome, thanks!

I saw that CI just errored because of a linting issue. Would you please run `npm run lint` and fix it accordingly?

Do you think that you can bind these sliders to the actual values of the model as described in #466?

Lastly, upon changing any value, Agent should make the right bottom "Save to keyboard" button appear. I'm not familiar with the codebase but I believe there are many parts of the code that trigger this event - basically whenever anything changes on the UI that affects the model.

laxu:
Yeah I could use some help in reading and setting the values to the configuration as well as triggering the save button. Pointers to files would help.

In case you are wondering about the stylelint config change, it's there to be able to add styles to the slider elements like the pips.

mondalaci:
@ert78gb Would you give some help to @laxu regarding binding these sliders to the model? If it's super quick, you could even implement it within this PR? Then he'd do the rest for #458 based on your patterns.

laxu:
Found the action for reading the values. Still need to figure out how to save them back in.

ert78gb:
Hi Guys,

I am starting my holiday travels to my parents tomorrow my girlfriend parents. I will travel 1500 km in the next 3 day. If i have free time I add commit to this PR or write introduction what should you do

ert78gb:
@laxu pls ensure you are use npm > 5.6.0. It is handle correct the package-lock.json file on every OS

ert78gb:
@laxu pls create 1 component that compose the slider component, the value indicator and the pips options
If you are use `this.store.select(getUserConfiguration).subscribe(` in the ngOnInit hook than unsubscribe in the OnDetroy hook or much better if you are async pipe

ert78gb:
@laxu pls ensure the new component dispatch change event only after 2-300 ms debounceTime
check my  d3d0905 commit how to save settings. It is a generic action it is able to save every property under the user-configuration route

mondalaci:
This is finally working! Well done, guys!

Thanks for the commit, and the advices, @ert78gb!

@laxu I previously suggested that we could incorporate the mouse speed feature into this PR, it don't think it's a good idea because this one will grow farily large. Let's do that separately later.

laxu:
Yeah I meant as soon as this is merged I could do the mouse speed page as a separate PR. I’ll check this in a few days.

ert78gb:
Before accept the PR pls solve the unsubscribe issue. If you leave the code unmodified that couse memory leak

mondalaci:
@ert78gb Thanks for the heads-up! This PR will only be merged after your final approval.

laxu:
I moved slider to its own wrapper component that handles its visual styling, debounce and passes the relevant values forward to the real slider component. I also added an unsubscribe to the led brightness `ngOnInit` .  

@ert78gb please check if it's fine when you have the time.

packages/uhk-web/src/app/components/slider-wrapper/slider-wrapper.component.ts
```diff
@@ -0,0 +1,75 @@
+import { AfterViewInit, Component, EventEmitter, forwardRef, Input, Output, ViewChild } from '@angular/core';
+import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
+import { NouisliderComponent } from 'ng2-nouislider/src/nouislider';
+import { Observable } from 'rxjs';
```
ert78gb: pls use rxjs/Observable import

packages/uhk-web/src/app/components/slider-wrapper/slider-wrapper.component.ts
```diff
@@ -0,0 +1,75 @@
...
+import { debounceTime } from 'rxjs/operator/debounceTime';
```
ert78gb: import operator import 'rxjs/add/operator/debounceTime';

packages/uhk-web/src/app/components/slider-wrapper/slider-wrapper.component.ts
```diff
@@ -0,0 +1,76 @@
...
+
+@Component({
+    selector: 'slider-wrapper',
+    templateUrl: './slider-wrapper.component.html',
+    styleUrls: ['./slider-wrapper.component.scss'],
+    providers: [
+        { provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => SliderWrapperComponent), multi: true }
+    ]
+})
+export class SliderWrapperComponent implements AfterViewInit, ControlValueAccessor {
```
ert78gb: pls add onDestroy interface and call `this.changeObserver$.complete()`

packages/uhk-web/src/app/components/slider-wrapper/slider-wrapper.component.ts
```diff
@@ -0,0 +1,76 @@
...
+    @ViewChild(NouisliderComponent) slider: NouisliderComponent;
+    @Input() min: number;
+    @Input() max: number;
+    @Input() step: number;
+    @Input() pips: SliderPips;
+    @Input() valueUnit: string;
+    @Output() onChange = new EventEmitter<number>();
+
+    public value: number;
+    private changeObserver$;
```
ert78gb: pls add type `Observer<number>`

mondalaci:
@laxu Regarding your last commit, I'm wondering if `.slider-wrapper-container` should be part of the `slider-component`. I assume we never want the slider handle to go beyond the min position, so this'd make sense.

I also observed a second vertical scrollbar:

screenshot from 2017-12-26 20-19-35

laxu:
I felt it could be outside it to allow for maximum flexibility for the layout so it can be used without any extra spacing if you want to put two next to each other and so on.

mondalaci:
@laxu Thanks for explaining! In light of your explanation, I agree that it should be good the way it is to allow for maximum flexibility.

.stylelintrc.json
```diff
@@ -64,8 +64,10 @@
     "selector-pseudo-class-parentheses-space-inside": "never",
     "selector-pseudo-element-case": "lower",
     "selector-pseudo-element-colon-notation": "single",
-    "selector-pseudo-element-no-unknown": true,
     "selector-type-case": "lower",
-    "selector-max-id": 0
+    "selector-max-id": 0,
+    "selector-pseudo-element-no-unknown": [ true, {
+      "ignorePseudoElements": ["ng-deep"]
```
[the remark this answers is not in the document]
laxu: It is necessary or we can't fix the alignment of the slider pip values as we can't add styles to the underlying component.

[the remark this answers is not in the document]
mondalaci: @ert78gb Any other way to make this happen?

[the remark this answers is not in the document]
ert78gb: just add the `.noUi-value {
    top: 2rem;
}
` to the root style.scss
simple css orverride

.stylelintrc.json
```diff
@@ -64,8 +64,8 @@
     "selector-pseudo-class-parentheses-space-inside": "never",
     "selector-pseudo-element-case": "lower",
     "selector-pseudo-element-colon-notation": "single",
-    "selector-pseudo-element-no-unknown": true,
     "selector-type-case": "lower",
-    "selector-max-id": 0
+    "selector-max-id": 0,
```
ert78gb: pls revert this file and I finished the review :)

mondalaci:
The double scrollbar issue still persists. Other than that, it looks perfect.

laxu:
The double scrollbar issue is caused by the `main-page` element being set to `display: inline-block`. Can you check if removing that rule breaks anything?

.stylelintrc.json
laxu: Done.

mondalaci:
Great job, indeed, @laxu! The suggested CSS fix works. I'll fix it right after the merge. Thank you!

PR #532, opened 2017-12-26
feat(popover): remap key on pressing enter

### LINKED ISSUE #530: Make Enter trigger the Remap Key action of the key action popover
When the key action popover is open, the capture keystroke mode is not active, and the user presses the Enter key, trigger the Remap Key action (button).

fjozsef:
Closes #530


mondalaci:
Just tested this, and now I seriously doubt whether this is a good idea. Usually, a HTML input is selected within the popover, like a select2, or a button. Enter has a well defined, standard behavior on these controls which is now overridden, so I'm confident that this is not the Right Thing to do. Whoever wants to quickly map keys should click with the middle mouse button on them (which is a feature that we should document), then press the wanted key. Sorry for ditching this one @fjozsef, your work is much appreciated nevertheless.

PR #534, opened 2017-12-27
feat(popover): sort keymaps and macros alphabetically

### LINKED ISSUE #512: Sort keymap and macro names alphabetically in select2
Within the key action popover

fjozsef:
Closes #512


mondalaci:
@fjozsef Would you please make the sorting case-insensitive? That way abc wouldn't be after XYZ.

ert78gb:
@fjozsef Why do you not sort in the reducer? I think the reducer and the effects should contain the business logic. Good practice to keep components as simple as possible.

fjozsef:
@mondalaci Of course.
@ert78gb I didn't want to assume that the keymaps and macros stored in order.
We can surely to do sorting in reducers, but in that case some of the logic should be rewritten. Like the navigation after duplicating a keymap.
https://github.com/UltimateHackingKeyboard/agent/blob/5ceca41e0fb56beda304f85966c016b5cd8453cb/packages/uhk-web/src/app/store/effects/keymap.ts#L38

Currently I don't know how could we decide which is the duplicated keymap if they will be sorted.
Can we assume its index is always the increment of the original keymap's index?

Mainting the order in the reducers needs more effort and seemed error prone to me.
Are you sure it should be implemented there?

ert78gb:
@fjozsef You all right need more effort to do the short in the reducer now but I think the app will be more testable in the future.

Action is not modifiable the new ngrx so we can not add new property to the action:)
The solution for the mentioned problem if we will refactor the user-configuration reducer that will store more than just the user-config. Example of the new user-config reducer

`
{
newKeymepAbbr: <new or duplicated keymap abbreviation>,
currentUserConfig,
prevUserConfig: <move from the app reducer>
}
`
I think don't change the code now. Your solution is good for us now

fjozsef:
@ert78gb I have noticed we have already sorted for the side menu, so I decided not to add any duplication and go with your idea, so I've started to move the logic into the reducer. Could you check if I am on the right path? I will fix the navigation bug (after duplication) and move the macro sorting after your approval.

packages/uhk-web/src/app/store/reducers/user-configuration.ts
```diff
@@ -27,7 +32,19 @@ export function reducer(state = initialState, action: Action & { payload?: any }
             newKeymap.name = generateName(state.keymaps, newKeymap.name);
             newKeymap.isDefault = (state.keymaps.length === 0);
 
-            changedUserConfiguration.keymaps = state.keymaps.concat(newKeymap);
+            const existingKeymap: Keymap = action.payload;
+
+            const index = sortedIndexBy(
```
ert78gb: nice solution, but not linear. I think the linear solution is.

```javascript
const keymaps = [];
let added = false;

for (const keymap of state.keymaps) {
    if (!added && keymap.name.localeCompare(newKeymap.name) > 0) {
        keymaps.push(newKeymap);
        added = true;
     }

    keymaps.push(keymap);
}

changedUserConfiguration.keymaps = keymaps;
```

+1 variable, but 1 loop

packages/uhk-web/src/app/store/reducers/user-configuration.ts
```diff
@@ -37,14 +54,39 @@ export function reducer(state = initialState, action: Action & { payload?: any }
...
 
-            changedUserConfiguration.keymaps = state.keymaps.map((keymap: Keymap) => {
-                keymap = Object.assign(new Keymap(), keymap);
+            if (duplicate) {
+                break;
+            }
 
-                if (!duplicate && keymap.abbreviation === action.payload.abbr) {
-                    keymap.name = name;
+            const index = state.keymaps.findIndex(keymap => keymap.abbreviation === action.payload.abbr);
```
ert78gb: I think you could use a same loop as in the ADD case with a little tuning

mondalaci:
@fjozsef Do you know whether you'll be able to resolve this in the foreseeable future? I'd like to know whether this PR is expected to get stuck. If so, maybe it'd be a better idea to hand it over to @ert78gb.

fjozsef:
@mondalaci Probably on the weekend

packages/uhk-web/src/app/store/reducers/user-configuration.ts
fjozsef: My solution is linear, but I can see your point. I am goint to change it.

fjozsef:
@mondalaci  @ert78gb  I've finished with the changes, you can review the PR again.

packages/uhk-web/src/app/store/reducers/user-configuration.ts
```diff
@@ -184,14 +214,28 @@ export function reducer(state = initialState, action: Action & { payload?: any }
...
+            if (duplicate) {
+                break;
+            }
 
-                return macro;
-            });
+            const index = state.macros.findIndex(macro => macro.id === action.payload.id);
+            const macroToRename = state.macros[index];
+            const newMacro = Object.assign(new Macro(), macroToRename, { name });
+            const newMacros: Macro[] = [];
```
ert78gb: why do not use the `addMacro` function here? If I see well it is the same code.

packages/uhk-web/src/app/store/reducers/user-configuration.ts
```diff
@@ -28,7 +33,19 @@ export function reducer(state = initialState, action: Action & { payload?: any }
             newKeymap.name = generateName(state.keymaps, newKeymap.name);
             newKeymap.isDefault = (state.keymaps.length === 0);
 
-            changedUserConfiguration.keymaps = state.keymaps.concat(newKeymap);
+            const newKeymaps: Keymap[] = [];
```
ert78gb: I think you could create a same function as `addMacro` to create the new keymaps.
But the better solution if you create a generic function instead of `addMacro` and use it everywhere.
```typescript
function createNewElements<T extends Macro | Keymap>(array:T[], newElement:T):T[] {
}
```

it is type safe and 1 code

packages/uhk-web/src/app/store/effects/keymap.ts
```diff
@@ -32,10 +34,17 @@ export class KeymapEffects {
 
     @Effect({ dispatch: false }) addOrDuplicate$: any = this.actions$
         .ofType(KeymapActions.ADD, KeymapActions.DUPLICATE)
-        .withLatestFrom(this.store)
-        .map(latest => latest[1].userConfiguration.keymaps)
-        .do(keymaps => {
-            this.router.navigate(['/keymap', keymaps[keymaps.length - 1].abbreviation]);
+        .withLatestFrom(this.store.let(getKeymaps()).pairwise(), (action, latest) => latest)
+        .do(([prevKeymaps, newKeymaps]) => {
+            let newKeymap;
```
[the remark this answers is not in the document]
fjozsef: And where should I add this helper function to?

packages/uhk-web/src/app/store/reducers/user-configuration.ts
fjozsef: Yes, just need to add a function which checks for the old item, which shouldn't be added.

packages/uhk-web/src/app/store/effects/keymap.ts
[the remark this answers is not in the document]
ert78gb: uhk-web/src/app/util/find-element.ts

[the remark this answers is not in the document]
fjozsef: If you don't mind, I'd rather use the word 'item' instead of 'element'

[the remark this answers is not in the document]
ert78gb: feel free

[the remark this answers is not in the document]
ert78gb: item sorter , sorter is better :)

ert78gb:
I think the appveyor CI is out of the sync and will never start again on this PR

mondalaci:
@ert78gb Looks good to me. Would you resolve the conflicts?

ert78gb:
wip :)

Issue #535, opened 2017-12-28
Reverse scrolling on OSX


mondalaci:
OSX users have reported that the upwards and downwards scroll directions are reversed with the UHK and in Agent. In other words, upwards actually goes downwards, and vice-versa.

The reason is that In OSX 10.7 Apple introduced a feature called natural scrolling. This reversed the scrolling direction on scroll wheels and trackpads to make it match better with iOS.

So in reality, Agent shows the correct direction, and works as expected on Linux, Windows, and also on OSX when natural scrolling is disabled, but not when natural scrolling is enabled on OSX.

This issue is fundamentally problematic. Imagine connecting the UHK to a Mac with natural scrolling disabled. It'll work as expected. Then connect it into another Mac with natural scrolling enabled. It'll work in reverse. It's impossible to detect over USB whether natural scrolling is enabled, so the correct direction cannot be guaranteed by any input device.

I plan to solve this issue by:
1. Explaining natural scrolling to OSX users on https://ultimatehackingkeyboard.com/start/mouse
2. Reversing the Scroll Upwards and Scroll Downwards mouse actions on the preloaded Mac keymaps.

spuder:
This is further complicated because `mod + page up` is by default opposite `mouse + page up` on the Mac Profile

The percentage of Mac users who configure OSX to be the same as 'windows/linux' scrolling I expect to be less than 50%. I think using the default inverted scrolling would be the best default option.

mondalaci:
I agree, and I expect that more and more people will use natural scrolling as it became the default on OSX and it's akin to iOS.

laxu:
There's a software called [Scroll Reverser](https://pilotmoon.com/scrollreverser/) that allows reversing scrolling for only the trackpad but not the mouse. I think the easiest way to solve this would be to recommend using it with the UHK. I have previously used this with a mouse on my Macbook Pro and it works fine in tandem with the built-in trackpad or the separate Apple Trackpad.

mondalaci:
@laxu Thanks for the tip! It's good to know about Scroll Reverser, but I think we shouldn't expect Mac users to install it during the onboarding process.

PR #536, opened 2017-12-28
Invert scroll on mac

spuder:
Partially resolves https://github.com/UltimateHackingKeyboard/agent/issues/535

Flips mouse scroll up/down on 

- Colemak for Mac 
- Dvorak for Mac 
- QWERTY for Mac 

Logic to change this in the configurator is still needed.


mondalaci:
Thank you! I'd probably rather not display the keys differently in Agent because they may not apply to another host. I'll rather clarify this in the onboarding process.

PR #537, opened 2017-12-28
Mouse speed UI

laxu:
Implements Mouse speed page UI as per #458

### DIFF
--- a/packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
+++ b/packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
@@ -2,14 +2,6 @@
     <i class="fa fa-sliders"></i>
     <span>Mouse speed</span>
 </h1>
-<button class="btn btn-danger mouse-speed-reset-button"
-mwlConfirmationPopover
-title="Are you sure?"
-placement="bottom"
-confirmText="Yes"
-cancelText="No"
-(confirm)="resetToDefault()">Reset speeds to default
-</button>
 <h3>Mouse pointer speed</h3>
 <div class="row mouse-speed-setting" *ngFor="let prop of moveProps">
     <div class="col-xs-12 col-md-6">
@@ -22,7 +14,14 @@ cancelText="No"
         <ng-container *ngTemplateOutlet="sliderTemplate; context: {prop: prop, sliderSettings: scrollSettings}"></ng-container>
     </div>
 </div>
-
+<button class="btn btn-danger mouse-speed-reset-button"
+        mwlConfirmationPopover
+        title="Are you sure?"
+        placement="bottom"
+        confirmText="Yes"
+        cancelText="No"
+        (confirm)="resetToDefault()">Reset speeds to default
+</button>
 <ng-template #sliderTemplate let-prop="prop" let-sliderSettings="sliderSettings">
     <label>
         <span>{{prop.title}}</span>
--- a/packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
+++ b/packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
@@ -35,7 +35,7 @@
 <button class="btn btn-danger mouse-speed-reset-button"
         mwlConfirmationPopover
         title="Are you sure?"
-        placement="bottom"
+        placement="top"
         confirmText="Yes"
         cancelText="No"
         (confirm)="resetToDefault()">Reset speeds to default
--- a/packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
+++ b/packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
@@ -2,6 +2,36 @@
     <i class="fa fa-sliders"></i>
     <span>Mouse speed</span>
 </h1>
-<p>
-    Coming soon ...
-</p>
+<h3>Mouse pointer speed</h3>
+<div class="row mouse-speed-setting" *ngFor="let prop of moveProps">
+    <div class="col-xs-12 col-md-6">
+        <ng-container *ngTemplateOutlet="sliderTemplate; context: {prop: prop, sliderSettings: moveSettings}"></ng-container>
+    </div>
+</div>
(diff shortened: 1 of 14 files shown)


mondalaci:
Thank you!

1. I can experience the double scrollbar issue here, too.
2. How about placig the reset button at the bottom of the page? I think it'd be a better place, and consistent with the placement of the Configuration page.
3. The step value of the mouse movement speed sliders should be set to 25 to assure that these values can only be multiplies of 25. Currently, when activating a slider and using the left and right keys, the value is always incremented by 1, and when moving these sliders with the mouse the values are not multiplies of 25.

laxu:
Changed the slider step to 25 and moved the reset button to the bottom. I originally put it on top because I thought it might not be found on the bottom but I guess it's ok there.

The double scrollbar is caused by the `.main-page` element being set to `display: inline-block`. I can remove the rule if it doesn't cause any issues on other pages.

mondalaci:
I'm surprised about `inline-block` because I did fix it in master yesterday and when merging it to your branch, nothing seemed to happen. Would you please remove it?

mondalaci:
Everything seems to be perfect on my side, too now that I cloned the repo again. Not sure what was the problem.

Waiting for the review of @ert78gb before merging this. Thanks for the awesome work!

packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.ts
```diff
@@ -8,5 +17,137 @@ import { Component } from '@angular/core';
...
+    }
+
+    onSetPropertyValue(propertyName: string, value: number): void {
+        this.store.dispatch(new SetUserConfigurationValueAction({
+            propertyName,
+            value: propertyName.indexOf('mouseMove') !== -1 ? value / MOUSE_MOVE_VALUE_MULTIPLIER : value
+        }));
+    }
+
+    resetToDefault() {
```
ert78gb: pls create a new device Action `ResetMouseSettingsAction' and implement this functionality in the a DeviceEffect. pls use LoadResetUserConfigurationAction from the effect to load the settings in the reducer

packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.ts
```diff
@@ -8,5 +17,137 @@ import { Component } from '@angular/core';
         'class': 'container-fluid'
     }
 })
-export class MouseSpeedComponent {
+export class MouseSpeedComponent implements OnInit, OnDestroy {
+    public moveProps = [
+        {
+            prop: 'mouseMoveInitialSpeed',
```
ert78gb: The agent currently template driven and I feel if you are create a new component and set property bindings than this part of code should be in the template for the consistency.

packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.ts
```diff
@@ -8,5 +17,137 @@ import { Component } from '@angular/core';
...
+    private userConfig$: Store<UserConfiguration>;
+    private userConfigSubscription: Subscription;
+
+    constructor(private store: Store<AppState>, private defaultUserConfigurationService: DefaultUserConfigurationService) {}
+
+    ngOnInit(): void {
+        this.userConfig$ = this.store.select(getUserConfiguration);
+        this.userConfigSubscription = this.userConfig$.subscribe(config => {
+            this.moveProps.forEach(moveProp => {
+                moveProp.value = config[moveProp.prop] * MOUSE_MOVE_VALUE_MULTIPLIER || 0;
```
ert78gb: @mondalaci I think better place to modify the values in the de/serialization?  If somewhere else we modify this values than we don't have to multiply the values.

mondalaci: It'd be nice to deal with the first-order, non-multiplied values everywhere, and only tell the slider to multiply it if that's what you mean.

laxu: Right now it works so that the slider handles the multiplied value only and the non-multiplied is used everywhere else.

packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.html
```diff
@@ -2,6 +2,43 @@ <h1>
...
+</div>
+<button class="btn btn-danger mouse-speed-reset-button"
+        mwlConfirmationPopover
+        title="Are you sure?"
+        placement="bottom"
+        confirmText="Yes"
+        cancelText="No"
+        (confirm)="resetToDefault()">Reset speeds to default
+</button>
+<ng-template #sliderTemplate let-prop="prop" let-sliderSettings="sliderSettings">
```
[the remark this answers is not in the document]
laxu: Done.

packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.ts
laxu: I'd rather keep them in here. I understand it creates more mess in the component itself but the other alternative is just a ton of repeated code in the template. I think this is a more maintainable solution.

packages/uhk-web/src/app/components/device/mouse-speed/mouse-speed.component.ts
laxu: Done.

mondalaci:
Thanks @laxu!

@ert78gb Would you take a final look?

packages/uhk-web/src/app/store/effects/device.ts
```diff
@@ -18,6 +18,7 @@ import {
     ActionTypes,
     ConnectionStateChangedAction,
     HideSaveToKeyboardButton,
+    ResetMouseSpeedSettingsAction,
```
ert78gb: it is an unused import pls delete it or use it in the `ofType`

ert78gb: Everything else is ok. @laxu great job

Issue #527, opened 2017-12-23
Checking firmware version on keyboard


laxu:
Right now the firmware page does not seem to report what firmware is currently on the keyboard. This would be good as well as some indicator if the firmware bundled with Agent is newer than current one or if your firmware is up to date.

mondalaci:
Yes, this would be very useful. It'd be worth to display the firmware version on the firmware page.

I've just written `packages/usb/get-right-firmware-version.js`. It should be queried when Agent detects the UHK and queries its properties. I'm happy to help with this if you choose to implement it.

Eventually, we'll also query the firmware version of the left half, too.

mondalaci:
Let's display the following below the title of the firmware page:

Firmware a.b.c is running on the left keyboard half.
Firmware d.e.f is running on the right keyboard half.

Please query the above version numbers after connecting to the UHK. See packages/usb/get-{left,right}-firmware-version.js for how to retrieve these values.

ert78gb:
@mondalaci If the left module not connected to the right what is the "left firmware version"?

mondalaci:
@ert78gb It should display "unknown".

mondalaci:
@ert78gb Forget what I said. This is good for now. We'll improve it later. Great job!

PR #540, opened 2017-12-30
Agent menu

### LINKED ISSUE #483: Agent menu
Create a new main menu item on the same level as the UHK menu item and put it below the already existing menu items. Call the menu item "Agent" and use the Agent icon for it.

Create two submenus for the "Agent" menu called "Settings" and "About". Make these level 2 menu items the same size as the level 3 menu items of the toplevel UHK menu.

Remove the existing fixed position Settings menu and make the new "Settings" menu item redirect to the already existing settings page.

Add the following to the About page:
* Agent version (based on the toplevel package.json)
* Agent on GitHub - linked to the Agent GitHub URL

Lastly, remove the Github ribbon.

laxu:
Fixes #483 

Adds app version handling. This works by generating `app-version.ts` on build by reading the main package.json and outputting it as a file that exports `appVersion` constant. Tried a few 3rd party node modules for this but they would not get the version number from the main package.json and opted for the nested uhk-web one instead.


mondalaci:
@laxu You're a force of nature!

1. The obsoleted Settings menu item is still in the left bottom corner. Would you please remove it?
2. The Agent menu item is indented at the same level just like the Keymaps menu item and the rest. This suggests that the Agent menu item is a chlid of "My UHK", in other words, it's device-specific. In reality, this is not the case, so it should be indented at the same level as the toplevel "My UHK" menu item.
3. On the About page, there is the header containing the Agent icon and "About, and there's also another, larger Agent Icon with "Agent" besides it, which feels a bit too much. I appreciate the branding effort, but maybe it'd be better to only leave the About header.

mondalaci:
The UI looks perfect now.

1. I'm wondering if there's any leftover code regarding the ribbon. I could see about 3 lines removed but I suspect there may be more.
2. I can see that you changed an svg and added another one. Why did you change it, and what's the purpose of the second one?
3. Please merge master into this branch. I released 1.0.4 recently to include your awesome mouse settings page.

laxu:
This should be up to date with master according to Github. I added the SVG so that it gets an icon class and the changed one is the one generated by `npm run sprites`.

mondalaci:
Thanks! All looks good to me. Waiting for @ert78gb to review.

laxu:
@ert78gb The `getAgentVersionInfo` action does not work properly. On Electron it reports version as 0.0.0 and in the browser gives an empty object. So unless that is fixed then it's no good.

ert78gb:
It is working only with the release builded version

ert78gb:
Maybe not full developed I will check it tomorrow or after tomorrow

ert78gb:
@laxu `getAgentVersionInfo` is working perfectly with install bundled version of agent. The settings menu was visible only in the electron version of the app and the auto updater also works only with installed version of the agent. `npm run electron` return with 0.0.0. is it true.

@mondalaci would you like to see the agent version number of the web version of the agent? Do you need only the agent version number or do you need the other version numbers `firmwareVersion`, `deviceProtocolVersion`, `userConfigVersion`, `hardwareConfigVersion`?

ert78gb:
I don't like the current solution with the `app-version.ts` because if someone update the main package.json file and do not build the agent then the next build command will generate a  `app-version.ts` file change and  git mark as changed file. 
I see 2 possible solution at this time:
- `app-version.ts` will not add to git. At this time is it ok, but I don't know if we will use other code analyser may it will cause error 
- import the main `package.json` into the agent web

laxu:
I think Git should not mark it as a changed file if its contents do not change. If the package version number changes then the app-version.ts should also be updated. I pretty much followed what was the client-side safe way to get package.json version number into the app. With the software being open source this is not a real concern though.

I think the `getAgentVersionInfo` would be fine if it returned consistent results. The main thing that is useful is probably the Agent version number but on the firmware page it would be nice to be able to show all the other info as well as what firmware version is installed on the hardware itself and use that to show an update indicator as discussed in #527 .

ert78gb:
3rd option. We don't use the main package.json file to manage the agent version.  We create a new versions.json file instead and it will contain all versions of agent related components and import (with require) this json file in the app.

mondalaci:
@ert78gb Yes, I'd like to see Agent version in the web build, too. It'd be very nice to display the other version numbers, too.

ert78gb:
@mondalaci ok for all version
@laxu of for show the keyboard firmware version

I vote for the 3rd option and I also open for extra options

laxu:
Haven't had time to fix this yet. Creating a versions.json file is fine for me. Would you like it to still go thru the `getAgentVersionInfo` action?

ert78gb:
@mondalaci I finished this issue
Summary:
- the version informations moved to the main package.json
  ```json
  "firmwareVersion": "8.0.0",
  "deviceProtocolVersion": "4.0.0",
  "userConfigVersion": "4.0.0",
  "hardwareConfigVersion": "1.0.0",
  ```
- the web module load the main package.json and read the version informations
- created `OpenUrlInNewWindowAction` it is open any URL in new window. Detect the running environment electron/browser and use the correct method
- created a new electron action `openUrl` the electron main service open the externals urls. It is the pair of `OpenUrlInNewWindowAction`

mondalaci:
At the end of the build task I see this which is incorrect:

Start download https://github.com/UltimateHackingKeyboard/firmware/releases/download/undefined/uhk-firmware-undefined.tar.bz2

Issue #477, opened 2017-10-28
Validate device, keymap, and macro names


mondalaci:
Currently, device, keymap, and macro names are not fully validated. Just try to set an empty string, or a string of whitespace characters.

Please make sure that Agent only accepts such names if they contain at least one non-whitespace character, and restore the original name upon hitting Enter if it's not valid. Also trim whitespaces from the left and right of these names.

PR #543, opened 2018-01-02
fix(user-config): Validate device, keymap, and macro names

ert78gb:
close: #477


mondalaci:
The device name doesn't get trimmed. I can add spaces on its left and right sides, and they don't get removed upon applying the name.

ert78gb:
fixed

Issue #455, opened 2017-10-15
Upload device configuration


mondalaci:
On the Device settings page currently the "Upload device configuration" button doesn't do anything. Please make it open a file upload dialog (filter for *.json and *.bin if possible) and overwrite the user configuration with the uploaded file, and handle both JSON and binary configurations properly.

If the specified configuration is invalid, display the usual red box in the top right corner with the message "Invalid configuration specified.".

PR #545, opened 2018-01-05
feat(user-config): Upload user configuration from json/bin file

ert78gb:
close: #455

### DIFF
--- a/packages/uhk-web/src/app/components/device/configuration/device-configuration.component.html
+++ b/packages/uhk-web/src/app/components/device/configuration/device-configuration.component.html
@@ -17,7 +17,6 @@
         <label class="btn btn-default btn-file">
             Upload device configuration
             <input type="file"
-                   accept=".json, .bin"
                    (change)="changeFile($event)">
         </label>
     </li>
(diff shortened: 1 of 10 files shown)


mondalaci:
I found the following issues:
1. I cannot choose a .json file, only a .bin file.
2. The configuration doesn't get automatically applied.
3. When uploading an invalid device configuration (I typed some random characters in the bin file) the red error component appears, but it doesn't contain any text. Please make it contain: "Invalid configuration was specified."

ert78gb:
1. Could you create a screenshot about the open file dialog? In browser the filter working. In electron the filter not applied all files selectable. Maybe elector works differently on linux and mac 
2. What do you mean? If user-config successfully loaded to the agent after automatically call the apply to device function, or just show the button? The code do the same as reset user config -> not show button, not apply user config
3. fixed

mondalaci:
1. I think you're probably right and the open dialog doesn't work consistently across OSes. Here's the Linux screenshot:

screenshot from 2018-01-06 10-36-36

2. Reset applies the configuration, but the upload configuration doesn't apply it. I cannot see the Save button appearing.

ert78gb:
@mondalaci done
- I removed the file filter because it works depend from the OS

PR #547, opened 2018-01-07
Update readme

opie4624:
Fixed a couple typos.
Updated the build instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ It's worth mentioning that Agent has two builds.
 
 The **electron build** is the desktop application which is meant to be used if you have an actual UHK at hand. It starts with an opening screen which detects your UHK. You cannot get past this screen without connecting a UHK via USB.
 
-The **web build** is meant to be used for demonstation purposes, so people who don't yet own a UHK can get a feel of Agent and its capabilities in their browser. Eventually, WebUSB support will be added to the web build, making it able to communicate with the UHK. Given the sandboxed nature of browers, the web build will always lack features that the electron build offers, so this won't make the electron build obsolete.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,23 +13,30 @@ It's worth mentioning that Agent has two builds.
...
 
 ## Building the electron application
 
-First up, make sure that node >=8.1.x and npm >=5.1.x are installed on your system. Next up:
+### Step 1: Build Dependencies
 
-```
-# Execute the following line on Linux. Use relevant package manager and package names on non-Debian based distros.
-apt-get install libusb-dev libudev-dev g++
+You'll need Node (>=8.1.x) and NPM (>=5.1.x). Use your OS package manager to install these. [Check the NodeJS site for more info.](https://nodejs.org/en/download/package-manager/ "Installing Node.js via package manager") Mac OS users can simply `brew install node` to get both.
```
ert78gb: I suggest to use NPM >= 5.6.0. it is handle correct the os specific optional dependencies in package-lock.json file. Or we just write use the NodeJs LTS :)

opie4624: This isn't a content edit, just a reorganization. Those are the same version constraints already present. Perhaps a follow-on edit updating the version constraints?

mondalaci:
@opie4624 The updated readme looks great to me. Thanks for the edit!

@ert78gb Feel free to edit it further.

README.md
ert78gb: I think you could rewrite the content. The first sentence: `You'll need latest NodeJs LTS`. It is 8.9.4 and it is contains npm 5.6.0.
If you would like to add any useful information that help others feel free.

Issue #564, opened 2018-01-19
Bogus key of a specific layout prevents Agent from saving


Hurricaaane:
I have created a layout called _OFF_ where all keys are unassigned, except the Fn key and some layer switching keys. I have recently noticed that for some reason, the left space key still outputs keystrokes when I press it despite appearing as unassigned in the Agent. When I try to edit this key in the Agent (ver 1.1.0), it prevents saving at all; the save button will not appear, and if it is already there, clicking it does nothing.

I have opened up an older version of the Agent source to start the debugger in Electron. This is the output when I try to edit the key:

```
error_reporter.ts:9 Effect "UserConfigEffects.saveUserConfig$" threw an error
error_reporter.ts:12 Source: UserConfigEffects {actions$: Actions, dataStorageRepository: ElectronDataStorageRepositoryService, store: Store, defaultUserConfigurationService: DefaultUserConfigurationService, deviceRendererService: DeviceRendererService…}
webpackJsonp../node_modules/@ngrx/effects/@ngrx/effects.es5.js.ErrorReporter.report @ error_reporter.ts:12
reportErrorThrown @ effect_notification.ts:29
verifyOutput @ effect_notification.ts:22
(anonymous) @ effect_sources.ts:43
webpackJsonp../node_modules/rxjs/operators/map.js.MapSubscriber._next @ map.ts:78
webpackJsonp../node_modules/rxjs/Subscriber.js.Subscriber.next @ Subscriber.ts:95
webpackJsonp../node_modules/rxjs/operators/exhaustMap.js.SwitchFirstMapSubscriber.notifyNext @ exhaustMap.ts:125
webpackJsonp../node_modules/rxjs/InnerSubscriber.js.InnerSubscriber._next @ InnerSubscriber.ts:17
webpackJsonp../node_modules/rxjs/Subscriber.js.Subscriber.next @ Subscriber.ts:95
webpackJsonp../node_modules/rxjs/operators/mergeMap.js.MergeMapSubscriber.notifyNext @ mergeMap.ts:154
webpackJsonp../node_modules/rxjs/InnerSubscriber.js.InnerSubscriber._next @ InnerSubscriber.ts:17
webpackJsonp../node_modules/rxjs/Subscriber.js.Subscriber.next @ Subscriber.ts:95
webpackJsonp../node_modules/rxjs/operators/map.js.MapSubscriber._next @ map.ts:83
[...]

PR #573, opened 2018-02-11
fix(keymap): Fix undefined keymap serialization

ert78gb:
close: #564


mondalaci:
This doesn't fix #564 for me. If I upload that configuration to my keyboard and change the key in question, the Upload to keyboard button still doesn't appear.

ert78gb:
@mondalaci  it is works for me.
I did the same steps in #564 
screen shot 2018-02-17 at 17 56 11

mondalaci:
Now that I checked it again, it works! Don't know why it didn't work for me previously. Thanks for the fix!

Issue #556, opened 2018-01-12
Make Agent able to unbrick bricked modules


mondalaci:
Please reimplement b6bdd1486c7619019e6f241b59bb6c9a960a8aca in Agent. Instead of waiting for the bootloader to be pinged forever, please only wait 30s, then abort.

PR #577, opened 2018-02-11
feat(device): Make Agent able to unbrick bricked modules

ert78gb:
close: #556


mondalaci:
Just retested it and it seems to work now. Well done!

Issue #551, opened 2018-01-09
Enhance firmware page


mondalaci:
Replace the "Flash firmware x.y.z (bundled with Agent) [button]" part with a single button titled "Flash firmware x.y.z (bundled with Agent)".

Replace the "Flash firmware file [file chooser] [button]" part with a single button titled "Choose firmware file and flash it". Pop up a file chooser window, filter "tar.bz2" file extension, and make Agent update the firmware when selecting a file.

When clicking on either of the above buttons, display a confirmation popover first (same as the one used for the "Reset device configuration" button). Make the title of the popover "Firmware update notice" and its content "Please don't disconnect your keyboard or the keyboard halves during the update process!". Name the buttons of the confirmation popover "Flash firmware" (red button) and "Cancel" (white button).

Gray out the menu items of the side menu during the firmware update process.

PR #588, opened 2018-03-14
feat: enhance device firmware page

ert78gb:
close: #551

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3359,22 +3359,14 @@
       "dev": true
     },
     "electron": {
-      "version": "1.7.11",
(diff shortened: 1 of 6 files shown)


ert78gb:
I don't know why but the html input file disregards value of the accept attribute on Mac.

mondalaci:
Thanks! Luckily, it's not a big problem that Mac doesn't accept tar.bz2

I played around with the UI, and I find that the confirmation popovers are rather disturbing, so please remove them. Sorry for changing my mind.

Also, please put the buttons next to each other because there's a lot of space available.

ert78gb:
done

PR #594, opened 2018-04-04
Make saving the configuration more robust

### LINKED ISSUE #467: Make saving the configuration more robust
When uploading the user configuration to the UHK, there is a phase when the lauchEepromTransfer USB command gets called, and then getKeyboardState is getting polled until it returns 0 as the first byte.

This phase is critical because if some for some reason it gets interrupted, the configuration will be left in an inconsistent state.

In order to make Agent more robust, let's do the following:
* When the user clicks on the "Save to keyboard button", save the user configuration to the file system automatically to "${agentDirectory}/${uhkUuid}.json" and only after that execute the already implemented transfer.
* After loading the configuration from the UHK, if the config serializer fails to parse it, display a restoration screen to the user.

The restoration screen should contain the following:

> Title: [fa-exclamation-circle icon] Fix configuration
> Paragraph: Your on-board device configuration is invalid. (1.5 rem above and below spacing just like for the "Device configuration" screen)
> Button: Restore the last valid device configuration (blue button)

Before displaying the above screen, check if there is a previous configuration saved, and also check if it's valid (the configuration parser can successfully parse it). If not, then rather display the following button:

> Button: Reset device configuration (red button, no confirmation)

Also create a menu item for this page as the first child of the Device node. Name the menu item "Fix configuration" and remove it once the button (whichever is displayed) gets clicked and the restoration process finished.

ert78gb:
Summary
- close #467 
- use `node-hid`: 0.5.7 in every module, because somewhere used 0.5.4 
- upgrade `fs-extra` to 5.0.0
- moved `getHardwareConfigFromDeviceResponse` and `getUserConfigFromDeviceResponse` from `UserConfig` effect to `uhk-common`, because it used in `uhk-agent` and `uhk-web` module
- removed some unused references

Implementation of #467:
- when agent start try to load the user and hardware configuration from the device/keyboard
- also try to load the backup of user configuration from the filesystem based on the harder uniqueId
- if user config parsing failed the agent redirect the user to the `/device/restore-user-configuration` page
  - if backup user config exists the system show the `Restore the last valid device configuration` button. If the user clicks on the button then system use the "Load user configuration from file" function to send the backup user configuration to the keyboard
  - if backup user config not exist the system show `Reset device configuration` that bind to "Reset user configuration" function
- when the system save the user configuration at the first step create a copy in the [userData](https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname) folder


mondalaci:
Just hit a bug in Agent-electron/Linux:

```
[21:13:12.342] [error] [DeviceService] Transferring error TypeError: Cannot read property 'map' of undefined
    at UserConfiguration.fromJsonObject (/home/laci/projects/agent/packages/uhk-agent/dist/webpack:/uhk-common/dist/src/config-serializer/config-items/user-configuration.js:46:1)
    at Object.exports.mapObjectToUserConfigBinaryBuffer (/home/laci/projects/agent/packages/uhk-agent/dist/webpack:/uhk-common/dist/src/util/helpers.js:24:1)
    at DeviceService.<anonymous> (/home/laci/projects/agent/packages/uhk-agent/dist/webpack:/src/services/device.service.ts:185:28)
    at Generator.next (<anonymous>)
    at fulfilled (/home/laci/projects/agent/packages/uhk-agent/dist/electron-main.js:53038:58)
    at <anonymous>

[21:13:12.477] [debug] [UhkHidDevice] Device communication closing.
[21:13:12.480] [info] [ '[DeviceRendererService] dispatch action',
  '{"payload":{"error":{"message":"Cannot read property \'map\' of undefined"}},"type":"[device] save configuration reply"}' ]
```

mondalaci:
I forgot to say that I hit the above when trying to save the configuration.

ert78gb:
@mondalaci sry, I fixed

ert78gb:
@mondalaci pls check now

mondalaci:
If the configuration is invalid:
* Only show the Device->Firmware menu item.
* Show the "Fix configuration" menu item and make it active by default. Make it the first item, just under the Device node (and above the "Firmware" menu item.
* When clicking on the "Restore the last valid device configuration" or "Reset configuration" button, jump to the default keymap.
* Make the Device node ("My UHK") non-editable, and fill it based on the last saved configuration if possible, or the default configuration.

ert78gb:
@mondalaci done the side menu modification

Issue #553, opened 2018-01-09
gh-pages deploy task


mondalaci:
I'd like a package.json task named `deploy-gh-pages` that deploys the web build to GitHub pages (into the gh-pages branch of the agent repo). The script should clean up previous files and directories that are not needed anymore.

mondalaci:
> @jos512 wrote: This is a good idea. But perhaps there are ways to do it easier with less time setting up and maintaining.
> 
> I see that #553 talks about the gh-pages branch. If you include the release files in the GitHub pages website, you'll already benefit from a CDN. That's because Github pages are served with GitHub's own CDN (source). A GitHub pages site may be 1GB big (source), so there should not be a problem with the release file taking up too much space. Alternatively you can always choose to only include the last 10 releases in the gh-pages branch, of course.

Thanks for the idea! Haven't thought of using GitHub Pages for this purpose yet. It's an interesting idea, and according to the testing I've just done it could work pretty well. If we'll ever use it, I think we should use another repo for releases because a single OS-specific Agent release occupies about 50M and the Agent repo would grow large very quickly.

GitHub Pages is surprisingly fast, and takes ~2s to download a 100MB file. I'd rather not delete older releases, so I'm not really in peace with the 1GB size constraint. Also, according to my knowledge a single file can't be larger than 100MB which shouldn't be a problem for now but it might be later.

Just for the sake of completeness, I mention that according to my testing GitHub Releases is backed by Amazon S3 whose bandwidth fluctuates quite wildly and it can be very slow at times, so it's out of question. (Also tested Cloudfront and it's not better for large files.)

Google Cloud Storage on the other hand seems to be the king of the hill because it consistently delivers 100MB/s transfer speed so I'd definitely pick it over the others.

PR #614, opened 2018-04-29
chore: add gh-pages deploy task for #553

krokofant:


mondalaci:
Thanks you for your contribution, @krokofant! I'd have never thought that this can be achieved so simply.

krokofant:
@mondalaci Glad I could be of help!

PR #621, opened 2018-05-01
Make long media key macro actions work.

### LINKED ISSUE #612: KeyMacroAction is broken
Currently, there are only 8 bits of space reserved for the scancode in the binary representation of a KeyMacroAction. Therefore, any KeyMacroAction with a scancode wider than 8 bits will not be converted correctly. This issue has not caused any symptoms due to the macro engine in the firmware being not yet operational, but it will need to be fixed before the macro engine is rolled out.

It might be a good idea to rename KeyMacroAction to KeystrokeMacroAction to make it consistent with KeystrokeAction.

mondalaci:
Fixes #612.


ert78gb:
I committed a few changes that solve the problem

Issue #567, opened 2018-01-29
Make firmware updates more robust


travisbhartwell:
I have the UHKAgent v1.1 installed on a 13-in Retina MacBook Pro, Early 2015 model, running macOS Sierra, v. 10.12.6. I tried flashing the included firmware and it fails.

Unfortunately I can't seem to copy the text from the output window, so I have attached a screenshot.

![screen shot 2018-01-29 at 10 49 52 am](https://user-images.githubusercontent.com/10110/35528267-30700d4a-04e2-11e8-81b2-821b81e2f6c1.png)

I have tried doing this with the keyboard connected directly to the laptop and through a USB hub with the same result on both.

mondalaci:
We've just tried Agent 1.1.0 on OSX 10.9.5 and 10.13.2 and couldn't reproduce this issue so far. Would you remove every USB device (including your hub) but the UHK, and retry a couple of times?

travisbhartwell:
That did it. After I unplugged everything, plugged the UHK directly into the MBP, and then restarted the Agent, I was able to flash the firmware. I'm wondering if the Agent got confused because of my Yubikey, which also is a HID device.

mondalaci:
Glad you've made it work! Hard to tell what happened exactly. Did you update because of the freeze bug? If so, please keep me up-to-date on how the new firmware behaves in this respect.

travisbhartwell:
Nope, no problems with the freeze bug. I just saw there was a new firmware update and decided to try it. No issues since doing so. :)

Loving my UHK!

spuder:
I'm getting the same error. So far I have tried

1. Upgrade from 1.1.0 to 1.1.1 of the agent
2. Restart the agent
3. Move the keyboard from my usb docking station to directly to my macbook pro
4. Unplug and replug in the keybaord several times
5. Restart the agent
6. Remove the cable and connect the two halves together
7. Remove logitech wireless receiver

Every time I try and flash the firmware bundled with the agent I get this error.


![screenshot 2018-03-09 11 06 32](https://user-images.githubusercontent.com/242382/37222532-b18f2d18-238a-11e8-8a54-a59a89897d71.png)

I've tried to flash 15 different times. I only had luck 3 of those times, and all 3 of those times I didn't have the logitech receiver plugged in. While not conclusive, it does seem coincidental. 

To anyone else having this issue

1. Remove any other keyboards / logitech wireless receivers
[...]

PR #626, opened 2018-05-02
Fix left half timeout during fw update (#567)

tenteen:
The snooze call is skipped in the try block when the command throws an
exception.  As a result, the command tries maxTry times as fast as
possible.

The fix is to move the snooze call outside of the try block.

PS: #GotMyUHK

### DIFF
--- a/packages/uhk-usb/src/util.ts
+++ b/packages/uhk-usb/src/util.ts
@@ -75,7 +75,6 @@ export async function retry(command: Function, maxTry = 3, logService?: LogServi
         try {
             // logService.debug(`[retry] try to run FUNCTION:\n ${command}, \n retry: ${retryCount}`);
             await command();
-            await snooze(100);
             // logService.debug(`[retry] success FUNCTION:\n ${command}, \n retry: ${retryCount}`);
             return;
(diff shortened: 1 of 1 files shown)


tenteen:
On my system, the left half consistently failed to update to firmware version 8.2.0 from ... whatever version it shipped with.  My UHK ran two different firmware versions for a few days :)

This change allowed what looks like a device rescan to succeed and thus allowed the left half to successfully update.  I now have a happy UHK running 8.2.0 on both halves.

ert78gb:
this info is enough thx.

mondalaci:
@tenteen Sorry for the late follow up! I do appreciate your PR, and I'd like the issue to get fixed, but I also want to make sure that the fix actually works. Your fix may make the update process more reliable for some, but I'm afraid that it might make the process less reliable for others. Currently, I can't reproduce the issue, and I don't understand it deeply yet. I need some more time to delve into this. I'll follow up.

mondalaci:
@tenteen Even though we can't reproduce this issue, your explanation does make sense, and your PR should make things better for everyone. Thank you for contributing to Agent!

Issue #595, opened 2018-04-06
Update the firmware versions after firmware updates


mondalaci:
Currently, the displayed current firmware versions don't get updated right after firmware updates. One has to restart Agent to see the updated firmware versions.

After the firmware update procedure finishes (regardless of whether the update succeeds or fails), please query the firmware versions via USB and update them on the firmware page.

Please also remove the OK button at the bottom of the page. I don't think anybody uses it, and it's confusing.

Lastly, please add a standard looking copy to clipboard UI element, so that users can copy the update log.

ert78gb:
@mondalaci The `OK` button starts the device poller. If I remove the button the app should start the poller after the success/failed firmware update. Is it ok?

mondalaci:
Yes, sounds great.

ert78gb:
I think it is closable. I forgot write the close keyword in the related PR.

PR #649, opened 2018-05-20
feat: update firmware version after update

ert78gb:
#595


mondalaci:
1. The style of the terminal scroll bars is not consistent with the system scroll bars. Please make them consistent.
2. Please don't jump to the default keymap after the update (but still refresh). Stay on the firmware page.

mondalaci:
Thanks! This looks great. I'll look into the "invalid hardware configuration" issue. It's not a problem on Linux.

ert78gb:
still the https://github.com/UltimateHackingKeyboard/firmware/issues/67 issue the "invalid hardware configuration"

PR #662, opened 2018-06-03
feat: make double tap to hold layer optional per key

### LINKED ISSUE #658: Make double tap to hold layer optional per key
Add a new "Lock layer when double tapping this key" option as seen on the following image:

untitled

Only show the option when "Activate" is selected in the first dropdown.

Make the option unchecked by default, but check it if it's enabled in the related key action.

Regarding the data model, so far `SwitchLayerAction.isLayerToggleable` was a boolean property. Please change it to an enum according to the related [firmware type](https://github.com/UltimateHackingKeyboard/firmware/blob/3fc4419f4f9ada0a7df02775213852937aa9e6a0/right/src/key_action.h#L30-L34). Use a string for the JSON representation of this property instead of the boolean, but make Agent able to read old configurations with boolean properties.

ert78gb:
close #658


mondalaci:
`npm run electron --layer-double-tap` doesn't display the relevant checkbox. What am I doing wrong?

ert78gb:
use `npm run electron -- --layer-double-tap=true`

mondalaci:
Almost perfect! I'd like the JSON representation of switchLayerMode to be string, not numeric:

* `"switchLayerMode": 0` -> `"switchLayerMode": "holdAndDoubleTapToggle"`
* `"switchLayerMode": 1` -> `"switchLayerMode": "toggle"`
* `"switchLayerMode": 2` -> `"switchLayerMode": "hold"`

PR #668, opened 2018-06-07
feat: remap key on all keymaps / layers

### LINKED ISSUE #618: Remap key on all keymaps / layers
Please consider the following mockup:

40581561-91e36fa0-615c-11e8-9af4-eb967a7613f6

Add the checkboxes at the bottom of the popover and make them unchecked by default.

* When "remap on all keymaps" is checked and the "Remap key" button is clicked, remap the key on the current layer of all keymaps.
* When "remap on all layers" is checked and the "Remap key" button is clicked, remap the key on all layers of the current keymap.
* When both "remap on all keymaps" and "remap on all layers" is checked and the "Remap key" button is clicked, remap the key on all layers of all keymaps.

Insert the above 3 bullet points as li elements into content of the tooltip.

ert78gb:
close #618 

This PR contains the bug fix of SwitchLayerAction remap delete the non SwitchLayerAction from layers


mondalaci:
When I check "Remap on all keymaps", every keymap gets overwritten with the current keymap (every key on every layer) instead of just overwriting the current key of the current layer on every keymap.

ert78gb:
I fixed

PR #706, opened 2018-06-28
WIP feat: replace ng2-select2 => ngx-select-ex

ert78gb:


ert78gb:
wait for merge optimistex/ngx-select-ex#96

PR #707, opened 2018-06-29
Fix Sleep key for macOS users in Mac keymaps and Fn layers

syl20bnr:
Not sure if this is only my system but the Sleep key on the Fn layer does not work on my Macbook. Changed it to be the shortcut on the official list of shortcut published by Apple: https://support.apple.com/en-us/HT201236 and it works as expected.

image

### DIFF
--- a/packages/uhk-web/src/app/services/user-config.json
+++ b/packages/uhk-web/src/app/services/user-config.json
@@ -689,8 +689,9 @@
                 null,
                 {
                   "keyActionType": "keystroke",
-                  "type": "system",
-                  "scancode": 130
+                  "type": "media",
+                  "scancode": 184,
+                  "modifierMask": 12
                 },
                 null,
                 {
@@ -2653,8 +2654,9 @@
                 null,
                 {
                   "keyActionType": "keystroke",
-                  "type": "system",
-                  "scancode": 130
+                  "type": "media",
+                  "scancode": 184,
+                  "modifierMask": 12
                 },
                 null,
                 {
@@ -4611,8 +4613,9 @@
                 null,
                 {
                   "keyActionType": "keystroke",
-                  "type": "system",
-                  "scancode": 130
+                  "type": "media",
+                  "scancode": 184,
+                  "modifierMask": 12
                 },
                 null,
                 {


eltang:
I can reproduce this issue. However, I have another keyboard whose sleep key works on all OSes, including macOS. I checked its code, and found that the correct scancode is 129, rather than 130, which Agent has been using. The problem now is that it's no longer immediately obvious what the key does, so I propose changing the scancode instead.

mondalaci:
[129 is Power Down, not Sleep.](https://github.com/UltimateHackingKeyboard/agent/blob/e8fe0f8d3ee21d0a5a531a8f9800ab01d0e3b579/packages/uhk-common/src/config-serializer/config-items/scancodes.json#L547-L562) I wouldn't change it just because OSX is weird. Power Down didn't work on my Mac, but Option+Cmd+Eject did. All things considered, I think the current solution is the best one.

eltang:
Too long or too short of a press will not work. The press must be almost exactly one second long.

mondalaci:
Gotcha, but this weirdness only applies to OSX. The right thing happens on the USB level. Maybe it'd be worth to make Agent display these weird cases on a per scancode, per OS basis eventually.

eltang:
It's an intentional behavior to filter out accidental presses, and provide more functionality from one key. If you aren't comfortable changing the scancode of the Sleep key, another option would be to add a new Power Down key with the 130 scancode in accordance with the HID spec.

mondalaci:
According to the [HID usage tables](http://www.usb.org/developers/hidpage/Hut1_12v2.pdf), 129 is power down and 130 is sleep. Agent is totally in line with the HID spec. I consider this issue to b e resolved.

eltang:
Sorry, what I meant is to add a Power Down key with the 129 scancode. Sleep should remain unchanged.

mondalaci:
I still don't get it. Agent already contains a Power Down key featuring the 129 scancode.

eltang:
Oops, I didn't realize that. In that case why not just use it?

mondalaci:
Because it does not work! Nothing works as well as Option+Cmd+Eject. Not even Sleep, Power Down, or Option+Cmd+Power Down. Nothing else.

This thread is taking forever and provides no further value. Please only follow up if you're certain that what you are suggesting actually works better than the current solution.

syl20bnr:
I can confirm what @mondalaci said above. It seems that `Power button` or `Touch ID sensor` are different than `Power Down`. High chance this is an Apple specific thing.

So the following behavior cannot be replicated:

image

> The problem now is that it's no longer immediately obvious what the key does, so I propose changing the scancode instead.

We could imagine that the agent at some point could display the action of the shortcut if known. For instance in a tooltip when hovering the mouse etc... But this is a polish feature maybe for macOS only so very low priority. This also depends on whether the shortcuts listed here https://support.apple.com/en-us/HT201236 are customizable or not, I believe we cannot change them. If it is possible to change them the it would be very hard for the agent to correctly know what a shortcut does.

@mondalaci maybe it's worth it to create a new issue about this even if it is a long term goal ? For instance I have this somewhere: 

image

We could display a camera icon instead of `Shift Command $ 4`. As well as cut/copy/paste icons instead of `Command x` etc... what do you think ?

mondalaci:
Such shortcuts are not only OS-specific, but I bet some of them are specific to various versions of OSes, and some of them might be customizable.

I don't believe it's worth the effort, and it has the potential of causing a lot of support requests long-term in cases where Agent is not in line with OS shortcuts, or when somebody just plugged his UHK to a different machine running a different OS and the shortcut is not available there. I strongly feel that we shouldn't implement this.

syl20bnr:
Makes sense 👍

PR #727, opened 2018-07-16
fix: popover components use OnPush change detection

ert78gb:


mondalaci:
I'm merging this because I have to release a pre-release, but this needs more work.

PR #751, opened 2018-08-14
Add CTRL+Q shortcut for quitting agent. (WIP)

richrd:
I'm mainly opening this for discussion. This is a quite standard shortcut so it would be nice to have as well as the already existing ALT+F4. What do you think?

There are some issues that this might have:
- What happens if the user quits while updating the firmware or config?
- What does the close button and existing ALT+F4 shortcut do in that situation?

### DIFF
--- a/packages/uhk-web/src/app/app.component.ts
+++ b/packages/uhk-web/src/app/app.component.ts
@@ -74,6 +74,14 @@ export class MainAppComponent implements OnDestroy {
             this.enableUsbStackTest();
             event.preventDefault();
         }
+
+        if (runningInElectron &&
+            event.ctrlKey &&
+            event.key === 'q' &&
+            !event.defaultPrevented) {
+            window.close();
+            event.preventDefault();
+        }
     }
 
     updateApp() {


mondalaci:
Now If I right click on a key and hit ctrl+q then Agent quits instead of capturing the key. This must be fixed.

I realize that it's also a problem for ctrl+s to a less extent.

If save is in progress and Agent quits, the configuration won't be corrupted. In this case, the write to EEPROM USB command has either launched or not. If it did, the new config won't be saved. If it didn't, the new config will be saved regardless of closing Agent.

kareltucek:
> What do you think?

My humble opinion is that closing windows is window manager's business, and therefore solving this on application level is surplus. 

One could even argument that teaching non standard shortcuts in case when there are other truly-standard alternatives (i.e., window manager's shortcut) means teaching bad habits. 

On the other hand, implementing redundant shortcuts is really not a big deal...

mondalaci:
I think @kareltucek is right. Case in point, ctrl+q is not a standard shortcut on Mac for this purpose.

richrd:
Thanks for the insights!

> Now If I right click on a key and hit ctrl+q then Agent quits instead of capturing the key.

Very good point, I totally forgot about that!

> My humble opinion is that closing windows is window manager's business, and therefore solving this on application level is surplus.

Yeah I mainly agree. It's a nice convenience though. For some reason almost most software I use react to CTRL+Q even though it's not the standard WM shortcut on my machine.

> Case in point, ctrl+q is not a standard shortcut on Mac for this purpose.

I was not aware of this either. I guess my best approach would actually just be to configure CTRL+Q as an additional shortcut for all windows as it's really simple to do on Kubuntu.

That said I think it's fine to close this unless anyone has any good reasoning for this to be done on the application level.

mondalaci:
Given the above reasons, I'd rather close this PR, and opened #752 as it's been highlighted by this issue. Thank you for your contribution!

PR #767, opened 2018-09-05
feat: secondary role visualisation

### LINKED ISSUE #138: Secondary role visualization
Secondary rule should be visualized as follows:

untitled

If there are modifiers set, this is how it should look like:

untitled

If a key has two symbols on it (for example "@" at the top and "2" at the bottom), and both modifiers and secondary role is set, then instead of displaying 4 rows (which we don't have space for) only display the bottom symbol ("2" in this case instead of "@" and "2").

For mod and space keys that are longer than the others (not sure what should define "longer than the others" but it should be a ratio), display them as such:

untitled

In order to visualize the "2" number inside circle let's use stacked font awesome icons. See [this](https://stackoverflow.com/questions/25534068/using-numbers-with-font-awesome).

ert78gb:
close #138

### DIFF
--- a/packages/uhk-web/src/app/components/svg/constants.ts
+++ b/packages/uhk-web/src/app/components/svg/constants.ts
@@ -1,0 +1,1 @@
+export const SECONDARY_ROLE_BOTTOM_MARGIN = 1;
--- a/packages/uhk-web/src/app/components/svg/keys/index.ts
+++ b/packages/uhk-web/src/app/components/svg/keys/index.ts
@@ -11,3 +11,4 @@ export { SvgSingleIconKeyComponent } from './svg-single-icon-key';
 export { SvgSwitchKeymapKeyComponent } from './svg-switch-keymap-key';
 export { SvgTextIconKeyComponent } from './svg-text-icon-key';
 export { SvgTwoLineTextKeyComponent } from './svg-two-line-text-key';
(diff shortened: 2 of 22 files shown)


mondalaci:
The horizontal position of secondary role texts should be recomputed when changing the secondary role. Let's say I set the secondary role to Fn:



It looks right. Then I change the secondary role to Mouse:

mondalaci:
The recent fix resolved the previous issue. Now I found a new issue: If I set a scancode (let's say "A") and some modifires, and then I add secondary role, then the secondary role is not shown.

mondalaci:
If I set secondary role for the left thumb key, the vertical position is wrong. Interestingly, if I also set modifiers afterwards, it gets corrected.

untitled

mondalaci:
There is a problem with scancodes that have two letters drawn. See the following image. It works with modifiers, but doesn't look right with secondary role.

untitled

mondalaci:
Remap the left case button according to the following:

untitled

It looks right. Then remap it with the same parameters:

untitled2

The more I remap the higher the Space text gets positioned.

mondalaci:
I found an aesthetic issue that affects wide keys:



"A" and "Mod" are bottom-aligned. They'd be much nicer center-aligned:



Also, if the scancode contains 2 letters then it looks like this:



It should contain 2 letters because the modifiers are not in the way and there's enough vertical space for them:



Alignment already works well without modifiers:

PR #817, opened 2018-10-07
fix: stop event propagation when capturing keypress action

### LINKED ISSUE #752: Right click key capturing shouldn't trigger shortcuts
When right clicking on a key, and pressing ctrl+s, not only the shortcut gets captured, but ctrl+s triggers the save to keyboard action. A universal mechanism should be used to not trigger any shortcut when capturing.

(This is not a problem with the capture button of the key action popover.)

ert78gb:
close #752


mondalaci:
Ctrl+s still triggers save while capturing on Linux.

PR #818, opened 2018-10-07
feat: use icon for play/pause keypress action

### LINKED ISSUE #591: Play/Pause action icon
Replace the text of the "Play/Pause" media action with `kbd__play-pause.svg`.

ert78gb:
close #591


mondalaci:
The Play/Pause icon should be as tall as the other icons, but it's not as tall.

untitled

Issue #825, opened 2018-10-10
Fix npm vulnerabilities


xellio:
Running `npm install` after cloning the agent will result in a warning like

> audited 43340 packages in 9.53s
found 46 vulnerabilities (34 low, 1 moderate, 11 high)

At least, fixing those high vulnerabilities is not that hard.
Running the `npm audit fix` command will fix most of those vulnerabilities.
I will create a PR for this one.

xellio:
Not fixing known and obvious (in terms of presentation - by npm) vulnerabilities is reckless not a good idea - even if there is no bandwith for tasks like this. This should have a higher priority than any new feature.

For me - as a non javascript dev - it took less than an hour fixing this problem.
Those 11 high rated, as well as the 34 low rated vulnerabilities are gone.
The single moderate one is still present due to possible breaking changes.

I can not understand that this issue is closed without any comment (except the one on my first pull-request, that there is no bandwith for tasks like this). There is no bandwith for obvious security problems?

I know that you have a lot work to do - but you decided to create a javascript application with all its advantages and disadvantages. If you are not able to handle external dependency updates (due to existing workload), maybe you should rethink this decision.

Does it make sense to have green tests but known vulnerabilities in my code? Do you really want to ship known vulnerabilities to your customers? Don't get me wrong - I have a lot of respect in what you are doing. But closing a reasonable ticket without a comment is not the best way to handle feedback from possible contributors.

mondalaci:
Upgrading npm packages is not always trivial. Sometimes they require a significant time investment with testing included. As long as we have a single part time developer working on Agent, we will not jump on (the very frequent) security notifications of npm, but rather address them in batches from time to time. I respect your opinion, and don't expect you to agree.

Sorry for closing this issue prematurely. I didn't expect that you actually followed up with a fix. Thank you very much for your contribution!

ert78gb:
Hi @xellio you wrote you are not a javascript developer and I think this is the reason why you are worried about the vulnerabilities. We don't like them too :)

We always try to keep the dependencies up-to-date, but the Agent is a moderately complex application. It is a real multi-platform app, I mean it is not just a web app. The Electron gives a shell and the web app is able to run on the desktop.
The following areas are totally OS dependent and we have to be very careful when making changes:
- installer package (electron-builder dependency)
- usb communication (node-hid)
- electron
[...]

PR #826, opened 2018-10-10
fixed vulnerabilities;

xellio:
https://github.com/UltimateHackingKeyboard/agent/issues/825


mondalaci:
CI doesn't seem happy with your PR, unfortunately:

> ERR! lerna Unknown argument: scope

We could `npm update` on a several daily basis and fight issues like this all the time. We have no bandwidth for this, so we generally `npm update` much less frequently. If you can't fix this then I'll simply close it. Please let me know.

xellio:
I will have a look at this and reopen a PR as soon as I fixed this issue.
Not fixing known and obvious (in terms of presentation - by npm) vulnerabilities is not a good idea - even if there is no bandwith for tasks like this. This should have a higher priority than any new feature.

PR #842, opened 2018-10-22
feat: show udev rules on missing device screen

### LINKED ISSUE #827: Add instructions to update the udev rule file
The new node-hid version comes with an updated udev file. When using the old udev file, Agent is unable to access the UHK every now and then on Linux, so updating to the new file is necessary for a decent user experience.

When `/etc/udev/rules.d/50-uhk60.rules` is present but its content differs from the new udev rule file, display a modified "Cannot find your UHK" screen based on the following screenshot:

untitled

(Use the same font style. This is just an ugly mockup I created.)

ert78gb:
close #827


ert78gb:
@mondalaci I modified only the list introductions. Would you like to I change the subheader too?

mondalaci:
We can't get the [interface 0 bug](https://github.com/node-hid/node-hid/issues/258) bug fixed, so we have to deal with it.

When Agent is running on Linux and all the interfaces are present except interface 0, then change the "Please plug it in!" sentence of the "Cannot find your UHK" screen to "Please reconnect it!"

PR #845, opened 2018-10-29
feat: allow 8.4.3 firmware features

### LINKED ISSUE #822: Firmware version specific feature display
If the firmware version of the right half is larger or equal to 8.4.3 then:
* Hide the "Please note that macro playback is not implemented yet. You can create macros, but they won't have any effect until firmware support is implemented. We're working on this." message from the macro tab of the key action popover and from the pages of the macros.
* Enable the --layer-double-tap=true feature.

ert78gb:
close #822

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1049,7 +1049,7 @@
         },
         "readable-stream": {
           "version": "2.3.6",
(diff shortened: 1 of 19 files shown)


mondalaci:
Is there any leftover code regarding `--layer-double-tap=true` that can be cleaned up?

ert78gb:
`--layer-double-tap` command line argument option deleted

PR #854, opened 2018-11-12
fix: exclude Steam Controller from the devices list

### LINKED ISSUE #820: Available devices always relisted
With the new node-hid, available devices are always relisted, even when unchanged.

See [firmware.log.txt](https://github.com/UltimateHackingKeyboard/agent/files/2454354/firmware.log.txt)

ert78gb:
Fixes #820.


mondalaci:
Just for the record, this commit is about making the firmware update log shorter, thus easier to read. We love the Steam controller, and as it turned out, it's not the culprit. :)

Issue #872, opened 2018-11-30
Pressing Tab after "update description" corrupts UI


Ardagan:
Agent version 1.2.12

To repro:
1. Open Agent
2. Go to keyboard layout edit
3. Click on description of layout
4. Use tab to go to Update Description button
5. Press tab

mondalaci:
Thank you for reporting this issue! I can clearly reproduce it, and we'll fix it.

PR #878, opened 2018-12-16
fix: Pressing Tab after "update description" corrupts UI (#872)

monutam91:
Fixes: #872

### DIFF
--- a/packages/uhk-web/src/app/components/svg/module/svg-module.component.html
+++ b/packages/uhk-web/src/app/components/svg/module/svg-module.component.html
@@ -7,7 +7,7 @@
             [width]="key.width" [height]="key.height"
             [attr.transform]="'translate(' + key.x + ' ' + key.y + ')'"
             [attr.fill]="key.fill"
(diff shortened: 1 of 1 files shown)


packages/uhk-web/src/app/components/svg/module/svg-module.component.html
```diff
@@ -16,5 +16,6 @@
             (capture)="onCapture(i, $event)"
             (mouseenter)="onKeyHover(i, $event, true)"
             (mouseleave)="onKeyHover(i, $event, false)"
+            [attr.tabindex]="-1"
```
ert78gb: line 10 also contains the tab index settings. pls modify that.

And I think not need the one way data binding because the `-1` is constant value

```suggestion
            tabindex="-1"
```

mondalaci:
@dgyimesi Please take over this, because this has been in the queue for more than a month.

ert78gb:
pls remove the newly added line 19, because it is duplication

PR #879, opened 2018-12-16
fix: device recovery mode

### LINKED ISSUE #864: Fix device recovery mode
The device recovery mode (#607) got broken in Agent 1.2.12 which needs to be fixed.

Don't auto-close the recovery screen after the recovery process started.

Add a white "Close" button rightwards of the "Fix device" button. Make it active only when the update process is not active. Make it close the recovery screen when clicked.

ert78gb:
close #864

### DIFF
--- a/packages/uhk-agent/src/services/device.service.ts
+++ b/packages/uhk-agent/src/services/device.service.ts
@@ -220,10 +220,6 @@ export class DeviceService {
 
             await this.operations.updateRightFirmware();
 
-            await snooze(500);
-
-            this.pollUhkDevice();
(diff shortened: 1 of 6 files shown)


mondalaci:
This fix doesn't work for me. I triggered the bootloader by [shorting the RESET pin to GND](https://ultimatehackingkeyboard.com/modding-and-repair/unbricking-the-right-keyboard-half), and the recovery screen didn't appear. Please try it with your UHK.

mondalaci:
On further tries, this does work.

PR #895, opened 2019-01-17
fix: Don't change the tab immediately when closing the key action pop…

### LINKED ISSUE #886: Don't change the tab immediately when closing the key action popover
When closing the key action popover, the animation gets triggered which makes the popover move downwards and fade. Unfortunately, the tab also changes to None. This is a little bit hard to notice, but it's noticable, and it's slightly disturbing. The tab should stay the same until the animation is finished.

dgyimesi:
Fix #886.


mondalaci:
Looks good to me. CI is happy, so feel free to review, @ert78gb.

Issue #798, opened 2018-09-27
Make Right click, Esc set the key action to none


kareltucek:
Currently, there is number of ways how to map a key quickly via right mouse, however I do not see any way to unmap a key quickly. I personally need to unmap keys slightly more often than I need to map them.

Ctrl + click seems unused at the moment so that might be used?

mondalaci:
Ctrl and Alt are already utilized, even for right click. See the help screen of recent Agent versions. Other modifiers can't be used in a cross-platform manner (we tried). I can't see a way to make this happen, so I'm closing this issue, but feel free to follow up.

mondalaci:
I can't remember the details, but the bottom line is that Super and Ctrl can't be used in a cross-platform manner in Electron.

I think your suggested behavior is the most intuitive one. We'll make right-click, then Esc set the relevant key action to None.

luteijn:
Although I've been remapping escape a lot the last few days, trying out some alternatives, in general I would agree. Middle click might be an alternative, although currently it seems to do the same as right click? 
Related to this, maybe at the same time of addding the escape as a way to clear assignment, make it also possible to unmap a key by selecting none as the scan code, no modifiers and no secondary role?

mondalaci:
@luteijn Middle click has similar cross-platform issues as modifiers. Can't trigger it on Mac.

I don't feel strongly about your second suggestion, but I find it conceptually cleaner to set None actions only one way.

ert78gb:
IMHO breaking the generic logic is not straight forward;

kareltucek:
(@mondalaci  @dgyimesi @ert78gb As was suggested in #848, there are at least another two options how to realize the feature request without voiding generality - either "<shortcut> + <click key>" or "<shortcut> + <mouseover>". I strongly support these, since they provide sufficient number of shortcut combination for future extensions of similar nature.)

mondalaci:
@kareltucek I don't like the suggested solutions. Accidental keypresses can lead to unintended actions.

kareltucek:
I mean, lets choose a shortcut - lets say Ctrl+N (as None), which will clear any key which is currently under the cursor.

(I can't see how a user can possibly type shortcut like that by accident.)

mondalaci:
I understand what you're suggesting, but the idea of shortcuts that are dependent on the hovered item don't feel right to me.

mondalaci:
[...]

PR #897, opened 2019-01-17
feat: Make Right click, Esc set the key action to none

dgyimesi:
Do #798.


ert78gb:
The `ts-keycode-enum` very sexy but it adds 2,6 KB extra data to the bundled code. The current tree shaker not able to drop the unused enum values.
I support to use enum or constants instead of numbers for the better code reading, but I think this lib is overkill in our case because we use max 20 key code in the app.

and pls also commit the `package-lock.json` file too.

mondalaci:
@dgyimesi Please use `ts-keycode-enum`. Its codesize is negligible compared to the entire codebase of Agent, and I'd much rather use it than adding keycode constants every now and then. In my mind, easy, long-term maintainability trumps minor optimizations.

Your fix doesn't work for me, neither with the web build, nor with the desktop build. I right-clicked on a key, pressed Esc, and the key stayed the same.

dgyimesi:
@mondalaci I included the enum into the codebase as `const enum`. In this way no code will be added to the final bundle as tsc inlines referenced values.

mondalaci:
@dgyimesi I understand, but as I told you above, I'm more concerned about maintainability than code size, so please add `ts-keycode-enum` to the project.

mondalaci:
@dgyimesi Your code doesn't achieve the intended outcome. Esc should unmap the key, not cancel the mapping process. Unmapping the key means that its key action should be set to NoneAction.

On further thought however, I agree with @ert78gb in that this would break the generic logic, and it's not straightforward from the user's perspective, so let's not implement it.

@dgyimesi What I would like to do is to use constants instead of numeric key codes. So this issue will end up being a refactoring task.

Issue #689, opened 2018-06-19
Redesign the About page


mondalaci:
Please consider the following mockup:

![untitled](https://user-images.githubusercontent.com/300402/41625854-9b8f2e4c-741a-11e8-9b13-e033ca141540.png)

* Use packages/uhk-web/src/assets/images/agent-logo-with-text.svg as the top image.
* Use the standard bootstrap blue color for links.
* Get the list of contributors via the http://api.github.com/repos/UltimateHackingKeyboard/agent/contributors GitHub API endpoint, but only load it when the About page gets visited.

PR #899, opened 2019-01-20
feat: Redesign About page.

dgyimesi:
Do #689.


packages/uhk-web/src/app/components/agent/about/contributor-badge/contributor-badge.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Component({
+    selector: 'contributor-badge',
+    templateUrl: './contributor-badge.component.html',
+    styleUrls: ['./contributor-badge.component.scss']
+})
+export class ContributorBadgeComponent {
+    @Input() contributor: UHKContributor;
+
+    get name(): string {
```
ert78gb: I just would like to ask what is the benefit of using property getters instead of `{{ object.property }}` in the template?

dgyimesi: It's just clearer. GitHub API property names don't communicate their purpose clearly, plus they break lowerCamelCase policy.

packages/uhk-web/src/app/components/agent/about/about.component.ts
```diff
@@ -11,7 +12,15 @@ import { getVersions } from '../../../util';
         'class': 'container-fluid'
     }
 })
-export class AboutComponent {
+export class AboutComponent implements OnInit {
```
ert78gb: Maybe I overengineering the app, but I like to separate the responsibility of the elements.
In my solution, we create a reducer and actions to load and store the contributors and we handle the loading and error state too.

It is up to you to follow these suggestions or not, but I believe automatic testing will be easier for the app.

packages/uhk-web/src/app/components/agent/about/about.component.ts
```diff
@@ -11,7 +12,15 @@ import { getVersions } from '../../../util';
...
     version: string = getVersions().version;
     agentGithubUrl = Constants.AGENT_GITHUB_URL;
+    contributors$: any;
```
ert78gb: You created the `UHKContributor` interface, pls use it here too.

packages/uhk-web/src/app/components/agent/about/contributor-badge/contributor-badge.component.ts
ert78gb: ok, thx

ert78gb:
@dgyimesi You can use the `OpenUrlInNewWindowAction` to open link. This action handle the agent is running in browser or in electron

packages/uhk-web/src/app/components/agent/about/about.component.ts
```diff
@@ -1,8 +1,16 @@
-import { Component } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
+import { Store } from '@ngrx/store';
+import { Observable } from 'rxjs/Observable';
+
 import { Constants } from 'uhk-common';
 
 import { getVersions } from '../../../util';
 
+import { AppState, contributors } from './../../../store';
```
ert78gb: the `./` does not need in the imports. pls remove it every where

packages/uhk-web/src/app/store/effects/contributors.effect.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    @Effect() fetchContributors$: Observable<Action> = this.actions$
+        .ofType<FetchAgentContributorsAction>(ActionTypes.FETCH_AGENT_CONTRIBUTORS)
+        .pipe(
+            mergeMap(() => this.http.get<UHKContributor[]>(Constants.AGENT_CONTRIBUTORS_GITHUB_API_URL)),
+            switchMap((response: UHKContributor[]) => {
+                return from(response).pipe(
+                    mergeMap(
+                        (contributor: UHKContributor) => {
+                            return this.http.get(contributor.avatar_url, { responseType: 'blob' });
```
ert78gb: Why do you load the avatar images? The electron renderer process can load the images.

packages/uhk-web/src/app/components/agent/about/about.component.ts
dgyimesi: Vscode auto imports. Gonna create a rule for this...

packages/uhk-web/src/app/store/effects/contributors.effect.ts
dgyimesi: Because it renders awfully. Images show up randomly for contributors, so I decided to wait until each is available before applying the template.

ert78gb: Ok. I checked, ~0.5kb the size of an image, it is not big

dgyimesi:
@mondalaci the list is actually ordered by the number of contributions by default. Strangely the display order doesn't mimic the ordering of the response. I added an extra sort in the effect. Commit's coming soon.

Issue #347, opened 2017-07-13
Add prettier


ert78gb:
What do you think if I add [prettier](https://github.com/prettier/prettier) to our Dev process?

mondalaci:
I welcome everything that makes our development experience better. Any objections, @fjozsef, @attilacsanyi?

fjozsef:
Personally I don't see why would it be necessary. What features do we need from it which isn't handled by tslint or auto format from a modern IDE?

ert78gb:
It is not necessary. It is just an option. It is not just warning if the code is not match some rule, but it automaticaly fix it. And we are use different IDE, with different support of tslint and autoformat. It is a common cli to that could integrate as pre-hook commit and automaticaly fix some error.

mondalaci:
I feel a bit uneasy about getting our code automatically formatted. Consider the following code:

```
enum Direction {
    Up   = 1,
    Down = 2,
    Left = 3,
    Righ = 4
}
```

I assume that prettier would kill the above, intentional alignment. This may not be a strong case against prettier, but it's the only one on my mind.

mondalaci:
@ert78gb I've just seen the introductory video of prettier, and I'm confident that it's the way forward and it'll save us a lot of time on the long run. Please make Agent run on prettier! :)

owlstead:
> 
> 
> I feel a bit uneasy about getting our code automatically formatted. Consider the following code:
> 
> ```
> enum Direction {
>     Up   = 1,
>     Down = 2,
>     Left = 3,
>     Righ = 4
> }
> ```
> 
> I assume that prettier would kill the above, intentional alignment. This may not be a strong case against prettier, but it's the only one on my mind.

Intentional alignment should not be used anyway. It's a maintainability nightmare. This is the day that development environments allow for renaming and other higher level functions to be used. For instance, if somebody noticed that `Righ` should be `Right` then you suddenly would have `4` being displayed one space to the right. Developers have other things to do than to insert indentation and then to maintain them indefinitely. End-of-line comments are sometimes even worse.

attilacsanyi:
You can use prettier ignore.

I have these for all enums:

// prettier-ignoreexport enum MyActionTypes { Val1 = '[Feature] Value', ValXx2 = '[Feature] Longer value'}
I hope this helps.

Attila    On Thursday, 6 December 2018, 15:21:58 CET, owlstead <notifications@github.com> wrote:  
 
 

[...]

PR #908, opened 2019-01-27
chore: add Prettier

dgyimesi:
Add Prettier rules and apply autofix on entire repo.


mondalaci:
The team has ultimately agreed that prettier is overly opinionated, and it'd cause more problems than it solves. Closing this PR along with #347.

PR #921, opened 2019-02-24
feat: convert UdevRulesInfo to string enum

ert78gb:
For easier log reading better to use string enum.

### DIFF
--- a/packages/uhk-common/src/models/udev-rules-info.ts
+++ b/packages/uhk-common/src/models/udev-rules-info.ts
@@ -3,14 +3,14 @@
  * Only on Linux need extra udev rules.
  */
 export enum UdevRulesInfo {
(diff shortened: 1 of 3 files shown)


packages/uhk-common/src/models/udev-rules-info.ts
```diff
@@ -3,14 +3,14 @@
  * Only on Linux need extra udev rules.
  */
 export enum UdevRulesInfo {
-    Unkonwn,
-    Ok,
+    Unkonwn = 'Unkonwn',
```
mondalaci: This is misspelled. It should be `Unknown`.

ert78gb: fixed

PR #923, opened 2019-02-26
Update lerna dependency from 3.2.0 to 3.13.0

sky1e:
Tested on linux by running `npm test` and by making sure it ran. This fixes 3 of the high-severity vulnerabilities reported by `npm audit`.


ert78gb:
@sky1e thx for the PR. I would like to ask you what is the OS that you use because `packages/uhk-web/package-lock.json` contains `optional=true` additional entries. I am just curious these additional entries are the side effect of the npm/npm#17722

ert78gb:
>Tested on linux

my question was stupid

PR #925, opened 2019-02-28
chore: upgrade angular => 7.x

ert78gb:
It is a big PR and contains many package upgrades.

The following option removes the white spaces from the HTML content so space may be removed between HTML element. 
`
"angularCompilerOptions": {
    "preserveWhitespaces": false
  }
`
screenshot 2019-02-28 at 18 59 09

screenshot 2019-02-28 at 18 59 37

I tested the app and fixed where I found this error.

I switched back to the official version of `ngx-select-ex` because it Angular 7 compatible. I reviewed their modifications and they did the same performance improvement that I did earlier.


mondalaci:
There some major issues compared to the previous release:

1. Agent takes about twice as much time to start up now. Previously, the Agent icon spun once. Now it spins twice.
2. Switch from the Mouse speed page to a keymap. It takes about 1.5s now. It used to take about 0.3s.
3. There is no space around icons anymore. This can be seen everywhere.
4. There are other space issues all across the UI, usually around select boxes.

packages/uhk-web/src/app/store/effects/device.ts
```diff
@@ -286,8 +286,8 @@ export class DeviceEffects {
 
     private sendUserConfigToKeyboard(userConfiguration: UserConfiguration, hardwareConfig: HardwareConfiguration): void {
         this.deviceRendererService.saveUserConfiguration({
-            uniqueId: hardwareConfig && hardwareConfig.uniqueId,
-            configuration: userConfiguration.toJsonObject()
-        });
+                                                             uniqueId: hardwareConfig && hardwareConfig.uniqueId,
```
mondalaci: Why did such indentations change? I preferred the old style.

ert78gb: I upgraded my editor and it was its default settings.

PR #928, opened 2019-03-08
chore: upgrade bootstrap to 3.4.1

ert78gb:


mondalaci:
The width of the tooltips have changed. The following is not good:



Other tooltips have changed, too. Please fix them according to the following screenshots:



(On the above image, the tooltip is transparent. Please make it opaque like the other tooltips.)




Please fix the following, too:

ert78gb:
I fixed the width problem of the tooltips. I can not reproduce the opaque problem.

packages/uhk-web/src/app/directives/tooltip/tooltip.directive.ts
```diff
@@ -56,6 +60,12 @@ export class TooltipDirective implements AfterContentInit, OnChanges {
             innerStyle = `style="max-width: ${this.maxWidth}px;"`;
         }
 
+        console.log({
```
mondalaci: Why do you want to log this?

ert78gb: I just used it for ugly debugging. Now deleted

Issue #834, opened 2018-10-14
Support more mouse buttons


Dagon:
My mouse has 7 buttons - it would be nice to be able to send buttons higher than 3 from the mouse layer on the keyboard.

mondalaci:
We'll make this happen. I slightly edited the subject of this issue because mouse actions can be assigned to any layers, not just the mouse layer.

ert78gb:
@mondalaci I fixed the multi-button selection and added the Key Action texts.
The question is how to visualize the action on the keyboard?
The left-click visualization
<img width="90" alt="Screenshot 2019-08-21 at 20 40 50" src="https://user-images.githubusercontent.com/496775/63459485-e7fe2400-c454-11e9-80d6-fb8540a7edab.png">
 
Now I just added the number as the text.
<img width="64" alt="Screenshot 2019-08-21 at 20 47 15" src="https://user-images.githubusercontent.com/496775/63459558-09f7a680-c455-11e9-97de-afb3d85180cd.png">

PR #936, opened 2019-04-14
feat: support more mouse buttons

dgyimesi:
close: #834


mondalaci:
@dgyimesi When I click on the left button, button 4-8 get toggled.

Untitled

When I click on button 4-8, none of them get toggled.

mondalaci:
@krokofant We can definitely use some help on this. If you want to contribute please [sign the CLA](https://github.com/UltimateHackingKeyboard/agent/blob/master/CONTRIBUTING.md) first. Thanks so much!

ert78gb:
@mondalaci I fixed the multi-button selection and added the Key Action texts.
The question is how to visualize the action on the keyboard?
The left-click visualization
Screenshot 2019-08-21 at 20 40 50
 
Now I just added the number as the text.
Screenshot 2019-08-21 at 20 47 15

krokofant:
Maybe make the button number formated similar to the weight of these mouse actions
image
font-weight:bold and font-size around 30px

Does the mouse buttons 4-8 really exists? I'm not sure what documentation to look at but the Windows docs for [virtual key codes](https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes) lists left, right, middle, XBUTTON1 and XBUTTON2. I'm wondering what the relationship is between the left, middle, right, XBUTTON1, XBUTTON2 and the ones targeted in this PR.

Would the 4:th button actually be what MS refers to as XBUTTON1? In that case the naming maybe should be `Left`, `Middle`, `Right`, `X1`, `X2`, `X3`, `X4`, `X5`.

mondalaci:
@ert78gb Please only display Button 4-8 in the popover if the user config is greater or equal than 4.1.1. (I'll release a new firmware today.)

@krokofant The buttons are numbered based on their USB representation. See the https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf document, although I can't find the relevant page right now. I'm open to alternative namings as long as they're consistent with what OSes display.

mondalaci:
@krokofant On second thought, I agree that button 4-8 are not intuitive names. Let's leave them the way they are for now, though. Once this PR is merged, we'll be able to trigger these buttons with the UHK and test them with various key capture applications to find more intuitive names.

ert78gb:
@mondalaci I added the 
>  Please only display Button 4-8 in the popover if the user config is greater or equal than 4.1.1.

Issue #877, opened 2018-12-16
Clicking on a link with Mouse 3 inside the agent opens a new window


monutam91:
Agent version: 1.2.12

Repro steps:

1. Open agent
2. Go to About
3. Click on link 'Agent on Github' with Mouse 3 (Mouse wheel)
4. It opens a new UHK Agent window with the page opened inside of it

mondalaci:
Good catch! This should clearly open in the browser.

cokert:
I've been looking into this a bit and found this link.  Not sure if you guys have looked into this much or are aware, but this seems to be a common issue.  https://stackoverflow.com/questions/44578022/electron-prevent-middle-click

cokert:
I'm working on this and have the fix pretty much done (essentially, you just to add an auxclick handler in addition to the onclick handler as mentioned in the link).  I'm seeing two different ways of handling links in the code, though.  One decorates the a tag with externalLink, and one has a click handler in the .ts file.  What's the "preferred" way of handling links?  I can change all the links to one VS the other as part of fixing this.  It looks like they were implemented at two different times in two totally different ways.  IMO, the externalLink decorator seems cleaner.

Also, the help page has an embedded URL (it's the link for entering accented characters), where the other URLs are in a Constants file.  Do you guys want me to move the accented characters URL to a constant?

mondalaci:
@cokert Thank you very much for the fix! I think it's as clean as possible, and extracting the URL into its own constant seems fine.

mondalaci:
Fixed by #937.

Issue #931, opened 2019-03-25
Make CLA signing mandatory for contributors


mondalaci:
Set up https://cla-assistant.io/ for the Agent repo to make signing https://gist.github.com/mondalaci/3f4b8106e4683e6e510011ad976e5fcc mandatory for contributors.

mondalaci:
There is a [major issue](https://github.com/cla-assistant/cla-assistant/issues/336) regarding CLA Assistant. We'll see [whether CLAHub is also affected](https://github.com/clahub/clahub/issues/170). We may end up using [clabot](https://github.com/clabot/clabot).

mondalaci:
@ert78gb This has been done.

PR #937, opened 2019-04-14
fix for middle clicking links

cokert:
This is a fix for #877.  There are some questions on the original thread that I asked, but here's what I've done so far for feedback.

### DIFF
--- a/packages/uhk-common/src/util/constants.ts
+++ b/packages/uhk-common/src/util/constants.ts
@@ -2,6 +2,5 @@ export namespace Constants {
     export const AGENT_GITHUB_URL = 'https://github.com/UltimateHackingKeyboard/agent';
     export const AGENT_CONTRIBUTORS_GITHUB_PAGE_URL = 'https://github.com/UltimateHackingKeyboard/agent/graphs/contributors';
     export const AGENT_CONTRIBUTORS_GITHUB_API_URL = 'https://api.github.com/repos/UltimateHackingKeyboard/agent/contributors';
-    export const AGENT_HELP_ACCENT_CHARS_URL = 'https://uhk.io/blog/2018/06/23/how-can-i-type-accented-characters-with-my-uhk';
     export const FIRMWARE_GITHUB_ISSUE_URL = 'https://github.com/UltimateHackingKeyboard/agent/issues/new';
 }
--- a/packages/uhk-common/src/util/constants.ts
+++ b/packages/uhk-common/src/util/constants.ts
@@ -2,5 +2,6 @@ export namespace Constants {
     export const AGENT_GITHUB_URL = 'https://github.com/UltimateHackingKeyboard/agent';
     export const AGENT_CONTRIBUTORS_GITHUB_PAGE_URL = 'https://github.com/UltimateHackingKeyboard/agent/graphs/contributors';
     export const AGENT_CONTRIBUTORS_GITHUB_API_URL = 'https://api.github.com/repos/UltimateHackingKeyboard/agent/contributors';
+    export const AGENT_HELP_ACCENT_CHARS_URL = 'https://ultimatehackingkeyboard.com/blog/2018/06/23/how-can-i-type-accented-characters-with-my-uhk'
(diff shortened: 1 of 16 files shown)


mondalaci:
Thank you very much for your contribution! CI failed because the linter complained. Would you please fix these stylistic issues?

packages/uhk-common/src/util/constants.ts
```diff
@@ -2,5 +2,6 @@ export namespace Constants {
     export const AGENT_GITHUB_URL = 'https://github.com/UltimateHackingKeyboard/agent';
     export const AGENT_CONTRIBUTORS_GITHUB_PAGE_URL = 'https://github.com/UltimateHackingKeyboard/agent/graphs/contributors';
     export const AGENT_CONTRIBUTORS_GITHUB_API_URL = 'https://api.github.com/repos/UltimateHackingKeyboard/agent/contributors';
+    export const AGENT_HELP_ACCENT_CHARS_URL = 'https://ultimatehackingkeyboard.com/blog/2018/06/23/how-can-i-type-accented-characters-with-my-uhk'
```
ert78gb: This constant is never used. Why did you add it?

cokert: It is in the help page.  It was originally embedded directly in the HTML of the component.  I didn't change it in this original commit/PR, but it's should be used now.  It's discussed in comments of the original issue thread.

packages/uhk-common/src/util/constants.ts
```diff
@@ -2,6 +2,6 @@ export namespace Constants {
     export const AGENT_GITHUB_URL = 'https://github.com/UltimateHackingKeyboard/agent';
     export const AGENT_CONTRIBUTORS_GITHUB_PAGE_URL = 'https://github.com/UltimateHackingKeyboard/agent/graphs/contributors';
     export const AGENT_CONTRIBUTORS_GITHUB_API_URL = 'https://api.github.com/repos/UltimateHackingKeyboard/agent/contributors';
-    export const AGENT_HELP_ACCENT_CHARS_URL = 'https://ultimatehackingkeyboard.com/blog/2018/06/23/how-can-i-type-accented-characters-with-my-uhk'
+    export const AGENT_HELP_ACCENT_CHARS_URL = 'https://uhk.io/blog/2018/06/23/how-can-i-type-accented-characters-with-my-uhk';
```
mondalaci: I'd rather prefer ultimatehackingkeyboard.com. We only use our uhk.io in onboarding URLs. Using the full URL also makes opening such links faster, as no redirection is involved.

cokert:
In this updated commit I've changed all the links to use the externalUrl directive.  I also found a link in the Firmware page that isn't visible by default.

packages/uhk-common/src/util/constants.ts
cokert: That puts the URL over the linter's limit.  I could break this into separate lines?  

But that brings up the question again of do you guys WANT all URLs in the constants file?  It seems kind of weird to have FAQ links embedded in the constants file.  For those, I'd think it better to have them just embedded in the HTML like it was originally.  Especially if you want to add more links in the help page over time.

mondalaci: Sure, you can break overly large strings into separate lines.

I agree that it's simpler to embed such URLs in the HTML. Please do.

cokert:
Ok, the FAQ link is back to being embedded in the HTML.   :)

ert78gb:
@cokert Thx for the PR. Could you create an empty commit to start the travis builds.

cokert:
I signed the CLA last night, but it's still reporting as pending.  Is there something you guys need to do on your side to move this forward?

mondalaci:
@cokert I've just removed CLA assistant because of a [standing issue](https://github.com/UltimateHackingKeyboard/agent/issues/931#issuecomment-483848667). We'll implement CLA signing later, and it shouldn't be needed in the meantime. Can you commit now?

cokert:
Ok, empty commit pushed, CI started.

cokert:
Is there anything else I need to do on this from my end?

mondalaci:
@cokert Just merged this PR. Sorry for the delay, and thank you very much for your contribution!

PR #990, opened 2019-07-26
feat: decrease the size between the keyboard layout and description in split mode 

ert78gb:


mondalaci:
The increase/decrease size animation works great, but the space between the keyboard and keymap name should be larger. Please see the following image of the previous Agent version for reference:

Untitled

PR #1006, opened 2019-07-31
chore(usb): add kboot-firmware-upgrade script

ert78gb:
With this script, we can test the new kboot firmware upgrade without new release

run the following command in the agent root directory:
`$ node -r ts-node/register -r source-map-support/register packages/usb/kboot-firmware-upgrade.ts`


mondalaci:
The script bricks my UHK. See [the log](https://github.com/UltimateHackingKeyboard/agent/files/3450524/update.txt).

PR #1008, opened 2019-08-03
chore(kboot): add more logging

ert78gb:


ert78gb:
@mondalaci Can u test it? I did  few modification and now it is looks like stable on Win 10. I will test on Mac and Linux also on afternoon

PR #1024, opened 2019-08-23
fix: allowed file extension for the firmware file upload

ert78gb:
The `.tar.bz2` value does not allow selecting the firmware file on Mac


mondalaci:
This works well on Linux, too. Please test it on Windows.

ert78gb:
This is works well on Win 10 I merge it after the CI run success

Issue #847, opened 2018-11-01
Macro reference count


kareltucek:
Small enhancement idea - it would be nice if agent showed count of references for each macro. I.e., how many times a macro is referenced from keymaps.

Usecase - I wanted to clean up my macros a bit - delete unused ones etc., but oops... Which ones are actualy unused? 

(of course, this is an extremely-low-priority request)

mondalaci:
This feature makes sense to me, although I'm wondering if we should clutter the UI by displaying the counts. Maybe a "Delete unused macros" button would be a nicer solution.

kareltucek:
I don't think that "Delete unused macros" is a good solution, because it makes a huge-and-ingeneral assumption about what the user actually wants. (In other words: I don't always want to delete all unused macros.)

I think that the count itself is not very important - it is the indication of unreferenced macros that matter. E.g., showing reference count (or some sort of indication) only for unreferenced macros would be totally sufficient.

kareltucek:
(I am not sure which form it should take - these things can be indicated by very subtle means like different font type supplemented by mouseover tooltip or a small icon next to unreferenced macros...)

mondalaci:
I tend to agree about a general solution. I think we should display the counts in the sidebar aligned rightwards within [bootstrap pills](https://getbootstrap.com/docs/3.3/components/#badges).  We should use subtle color that is not distracting to the eye.

kareltucek:
Sounds fine ;-)

cokert:
Is anyone actively working on this?  I have something that should be ready in the next few days that fixes this if not.

mondalaci:
@cokert We haven't started working on this yet. Your contribution is definitely welcome.

cokert:
I have this working, but I'm assuming we want the pills all aligned to the right, correct? 

![image](https://user-images.githubusercontent.com/1710772/57121535-425c6300-6d46-11e9-9d77-759dafccc38a.png)

mondalaci:
Correct.

cokert:
Ok, I have it aligned right (using a similar method to the star for startup keyboard).  This is what it looks like with double digit usage count.  Is this acceptable visually?
![image](https://user-images.githubusercontent.com/1710772/57227456-6ededc00-6fdf-11e9-929d-820ba2908225.png)

mondalaci:
Yes, this looks great!

cokert:
[...]

PR #1030, opened 2019-09-02
feat: show macro usage count

ert78gb:
close: #847


mondalaci:
This works well, but the these badges look rather annoying on the UI.

image

Please only show them if Alt is held.

ert78gb:
Implemented the ALT key visibility

Issue #1046, opened 2019-09-23
Macro keystroke capture sets all modifiers


mondalaci:
When adding a macro action with the "Add captured keystroke" button, every modifier gets set, even if none are pressed.

PR #1054, opened 2019-10-10
fix: Macro keystroke capture sets all modifiers

ert78gb:
close: #1046


mondalaci:
It's better now, but the following button captures forever even if I press a key.

Untitled

Issue #831, opened 2018-10-13
Gray out the current keymap


highend:
Hi,

Fn + 1, 2, 3, etc. switch to the standard keymaps that the UHK comes with

I didn't want to modify the original keymap (QWERTY for PC) but made a copy of it to modify this instead.
The problem I'm facing is that I can't switch to this keymap by remapping Fn + number keys for example.
The "Keymap" tab only lists the default keymaps in the drop-down list for "Switch to keymap" but not the newly created one...

So it would be great if additionally created keymaps are listed in this drop-down so that we can assign this switching to a key combination....

mondalaci:
The list intentionally doesn't show the actual keymap because it doesn't make sense to switch from the current keymap to the same keymap. You can switch to the new keymap from another keymap.

gmarian:
What about adding a note explaining this to the keymap tab?

In retrospect, it's quite obvious. However, when you're new, it isn't obvious that you need to do this mapping on all the other keymaps.

mondalaci:
@gmarian Yeah, I think we should feature every keymap in the list but gray out the current one and add a small message under it within its list item saying "You can only switch to this keymap from other keymaps."

PR #1055, opened 2019-10-10
feat: gray out the current keymap in keymap tab dropdown list

ert78gb:
close: #831


mondalaci:
This works well, but it's not exactly what I asked.

In the menu item of the greyed out keymap, please add a text under the keymap text with smaller font saying "You can only switch to this keymap from other keymaps."

ert78gb:
I added the text

Issue #1039, opened 2019-09-12
flickering texts in macro editor


ert78gb:
![uhk-agent-macro-menu](https://user-images.githubusercontent.com/496775/64812414-79b60880-d59f-11e9-8b6e-7b1185739a52.gif)

PR #1058, opened 2019-10-11
chore: upgrade electron => 6.1.5

ert78gb:
fix: #1039

Electron engine upgrades
- node: 12.4.0
- V8: 7.6.303.22
- Chromium: 76.0.3809.146


ert78gb:
Linux firmware upgrade is not stable with electron 6.

ert78gb:
I upgraded electron from 6.0.12 to 6.1.5 and the Linux firmware upgrade works again 10 success from 10 tries.

It also fix #1039

PR #1072, opened 2019-10-21
feat: macro text info tooltip

### LINKED ISSUE #1067: macro type text wrong output
I tried out the default macro `Silent Bob's address` and it result in the following output:

Silent Bob
711+2880 Nulla St.
Mankato Mississippi 96522
)257= 563+7401

the final line result in wrong characters for 
( => )  
) => =

It seams to me that the macro presses the keys on the keyboard where it thing the characters are, instead of using the characters that the text holds. 

Agent version: 1.2.15

Firmware 8.6.0 is running on the left keyboard half.
Firmware 8.6.0 is running on the right keyboard half.

ert78gb:
close #1067


mondalaci:
@ert78gb Please break the text so that one line is used per sentence.

ert78gb:
@mondalaci I added the line breaks

PR #1075, opened 2019-10-21
feat: add keypress modifiers tooltip

### LINKED ISSUE #933: Add help icon to explain what "Modifiers:" means when remapping a key.
Maybe I'm missing something (high chance of that), but I don't understand the intended effect where one selects an ordinary key, say **A**, then, underneath where it says **Scancode:**, (currently set to **A**, in this case, obviously), there is a row of buttons titled **Modifiers:**. 

Say I select one (or more?) of those **Modifiers**, then press **Remap key**. At this point small icons appear on the graphical depiction of that key, in the agent UI.  I don't know what this is doing. I've tried to guess (I don't own a UHC yet - I'm a potential buyer trying out the software), but no joy.

Other icons have a small **?**, which is helpful, but this one doesn't.  
Looks a wonderful keyboard. Very practical. Thank you for making it.

I note the web build of agent is v. 1.2.8, whereas desktop (which I can't run as I don't own a keyboard yet), is v.1.2.12. (Could you maybe bump the web version so potential buyers can try the latest?).

ert78gb:
close #933

### DIFF
--- a/packages/uhk-web/src/app/components/agent/help-page/help-page.component.html
+++ b/packages/uhk-web/src/app/components/agent/help-page/help-page.component.html
@@ -20,7 +20,6 @@
             <li>Right click on a key: Capture key</li>
             <li>Hold Shift while clicking on a key: Remap on all keymaps</li>
             <li>Hold Alt while clicking on a key: Remap on all layers</li>
-            <li>Hold Alt to see macro reference counts in the side menu</li>
         </ul>
     </div>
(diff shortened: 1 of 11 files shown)


mondalaci:
Please leave more space on the left side of the question mark icon. The same amount as for the secondary role question mark icon.

ert78gb:
@mondalaci more space added

PR #1092, opened 2019-11-04
feat: highlight save to keyboard.

### LINKED ISSUE #782: Highlight the "Save to keyboard" button the first time it appears
I've seen that [users sometimes don't notice](https://www.youtube.com/watch?v=VoBgLJGRKiM&t=5m03s) the "Save to keyboard" button, so let's show an animation that highlights it the first time it appears.

Use an SVG arrow that is easy to see and has the same color as the "Save to keyboard" button. Make the angle of the arrow so that it points exactly towards the button. Move the arrow from the center of the Agent window to the "Save the keyboard" button.

Make the arrow disappear when the user clicks in Agent. Make Agent only show this the first time, and don't ever show it again.

dgyimesi:
Closes #782.


mondalaci:
@ert78gb This is out of sync with the master. Please make it usable, and I'll test it.

PR #1123, opened 2020-01-03
feat: sanity-check firmware files before updates

### LINKED ISSUE #608: Sanity-check firmware files before updates
Right now, the user can specify any file, and Agent will try to update the firmware with it. This deletes the  firmware of the right keyboard half, and times out the update process.

Agent should sanity-check the firmware image. The sanity check should be done before running the update process both for user provided firmware images and the bundled firmware image.

The sanity check should do the following:
1. Extract the tarball.
2. Read package.json
3. Check if the files described by `devices[*].source` exist.
4. Check if the files described by `modules[*].source` exist.

If the sanity check fails, pop up the "Invalid firmware file provided." error message.

ert78gb:
close #608

### DIFF
--- a/packages/uhk-agent/package-lock.json
+++ b/packages/uhk-agent/package-lock.json
@@ -596,11 +596,6 @@
 				"commander": "~2.8.1"
 			}
 		},
-		"semver": {
-			"version": "5.6.0",
-			"resolved": "https://registry.npmjs.org/semver/-/semver-5.6.0.tgz",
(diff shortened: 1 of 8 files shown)


mondalaci:
Please handle invalid package.json in a nicer way, just like missing devices and modules directories.

image

ert78gb:
Pls share your firmware compressed file

mondalaci:
[uhk-firmware-8.7.0-invalid-package-json.tar.gz](https://github.com/UltimateHackingKeyboard/agent/files/4022194/uhk-firmware-8.7.0-invalid-package-json.tar.gz)

ert78gb:
added the invalid JSON handling

PR #1134, opened 2020-01-13
feat: show out of space warning

### LINKED ISSUE #278: Out of space warning
If the binary configuration is larger than the available space then it should not be sent to the UHK. In this case, Agent should display a permament warning bar at the bottom, displaying the current and available space.

The available space equals the 32K EEPROM space minus the hardware configuration area (which will probably be about 64 bytes).

ert78gb:
close #278


mondalaci:
Please use thousand separator commas (12,345) and use the [.bg-warning bootstrap color](https://getbootstrap.com/docs/4.0/utilities/colors/) instead of the blue background color.

PR #1173, opened 2020-02-21
fix: set udevRulesInfo = OK when Agent runs in privileged context

ert78gb:
close: #1171


mondalaci:
The permission check is not sufficiently general. https://github.com/sindresorhus/is-root/blob/master/index.js probably gets it right.

PR #1241, opened 2020-05-03
fix: don't crash when backup user config does not exists

ert78gb:


mondalaci:
After restoring the user configuration, please jump to the default keymap.

Issue #1250, opened 2020-05-13
Remove HTML tooltips


mondalaci:
The bulilt-in HTML title tooltip shouldn't appear.

![image](https://user-images.githubusercontent.com/300402/81876420-f06a3200-9582-11ea-9b2c-d20a52841c0f.png)

 This happens to:
* the star icon of the toolbar
* the star icon next to the keymap title
* the delete keymap icon
* the duplicate keymap icon
* possibly others?

PR #1251, opened 2020-05-17
chore: use ng-bootstrap tooltip

ert78gb:
close: #1250


mondalaci:
There are 4 problematic tooltips on the Mouse speed page:

image

ert78gb:
fixed

Issue #678, opened 2018-06-17
Add Neo2 preset keymap


davidak:
Here is a new issue to track progress since https://github.com/UltimateHackingKeyboard/agent/issues/91 was closed without adding it.

You find the layout with all 6 layers here: https://www.neo-layout.org/

mondalaci:
Thanks for the report! This is currently blocked by #617.

tuxlifan:
In the meantime I created a "Reverse" map that can be used to get back the letters a-z like in QWERTY when the OS-keymap is set to Neo2.
i.e. when you switch to Neo2 on OS level

bjoernjacobs:
Having Neo2 would be great. I just ordered my UHK and - to be honest - expected it to not have problems with any layout. Finding out that it is not supported yet (after I ordered) was a little bummer.

mondalaci:
@bjoernjacobs You can already create a Neo2 keymap on the UHK. In case 4 layers aren't sufficient for you, I think [Karel's UHK firmware fork](https://github.com/kareltucek/firmware) provides more. We'll add more layers to the native UHK firmware and Agent eventually.

bjoernjacobs:
@mondalaci Thank you for the fast feedback. If I can use the first 4 layers, I'm happy! (Layer 5 and 6 of Neo2 contain mainly greek letters which I do not need for coding)

mondalaci:
If you guys think that Neo2 can already be implemented reasonably well on the UHK without #617 then please come up with a Neo2 UHK keymap, attach your configuration, and I'll add it to the official UHK configuration.

qux42:
I build up neo2 layout on a English base layout. It works really good. Sadly I'm in holidays, so I can't export it at the moment.
I will do it at soon I'm back

Sam
________________________________

skaldesh:
@qux42 Hi, I did the same as you. Did you get it to work so that after pressing both Mod3 keys, you enter "Capslock" mode? I did not manage to get it work (but I tried it 2 years ago, dont know what changed)

evs-ch:
Just in case someone might find it useful to have something to start with, here's my config with a rather basic neo2 layout I've made ages ago when I got the UHK.
[...]

PR #1254, opened 2020-05-23
fix: import user configuration

ert78gb:
related story #678 

Import Buffer from the 'buffer/' package


mondalaci:
Only the web version was affected.

PR #1255, opened 2020-05-24
feat: Allow more character types on the LED display

### LINKED ISSUE #765: Allow more character types on the LED display
it seems i cannot use characters like `_`, `/`, and `\` on the led display, while it seems the display itself is perfectly capable of showing those characters.

more specifically, this prevents me from putting `\o/` on the display, which i consider a huge issue since anything else will not properly convey how awesome this keyboard is.

ert78gb:
close #765


mondalaci:
Please only allow the new characters if user config >= 4.2.0

You can test it with the newly released firmware 8.10.0

ert78gb:
added the user config check

Issue #1278, opened 2020-06-30
Cosmetic fixes


mondalaci:
1. Don't leave extra space at the end of keymap, macro and device names when editing them.

![image](https://user-images.githubusercontent.com/300402/86160533-2b60fe80-bb0c-11ea-987d-0cb2d7b727c3.png)

2. On the about page, put a spinner before the Loading... text, change the text to "Loading contributor list...", and animate it to expand. Make the animation delay the same as for layer switching (but don't let the layer switching animation checkbox change it.)

PR #1317, opened 2020-08-08
feat: improve the design of the auto grow edit

ert78gb:
1st part of the #1278


mondalaci:
Closing PR because all things considered, the previous implementation was better.

Issue #1319, opened 2020-08-08
Reduce key label if it doesn't fit


mondalaci:
![image](https://user-images.githubusercontent.com/300402/89718563-d6908d80-d9bf-11ea-8f72-2cbd289c563d.png)

PR #1322, opened 2020-08-09
feat: prefix every numpad key primary text with "Np "

ert78gb:
closes: #1320


ert78gb:
I think worth to wait for the #1319 before we merge it.
Screenshot 2020-08-09 at 10 04 17

PR #1323, opened 2020-08-09
feat: decouple Import Configuration from Save To Keyboard

### LINKED ISSUE #761: Decouple Import Configuration from Save To Keyboard
I think Agent should pop up a dialog box to ask whether the user wishes to do this, or better yet allow the user to make changes to it before saving it to the keyboard manually.

ert78gb:
closes #761


ert78gb:
"Are you sure ..." confirmation removed

Issue #1325, opened 2020-08-12
Firmware update failed from 8.9.3 to 8.10.1 on Surface Book 2


robert-blip:
1. Start agent
2. Go to _Firmware_
3. Click on _Flash firmware 8.10.1 (bundled with Agent)_

![image](https://user-images.githubusercontent.com/66017623/90022626-49408800-dca2-11ea-8877-ba712edd0b09.png)

Not sure, but maybe Agent is not waiting long enough. When reenumerating, I can see a Windows message saying Configuring Bootloader. Then, Agent says it is not able to write. While the 'USB found' sound did not yet play.

When updating my Garmin watch, first the USB disconnect sound is played, then a USB connect sound is played after which the connection is finally made. This process takes some time. More than the amount of time that Agent is waiting. Of course this is a very differernt device, but you never know. :-)

Log:
Start flashing firmware
Agent version: 1.5.1
Device right firmware version: 8.9.3
Device left firmware version: 8.9.3
[DeviceService] stop poll UHK Device
[DeviceService] stopped poll UHK Device
New firmware version: 8.10.1
[UhkOperations] Operating system: Windows_NT 10.0.18363 x64
[UhkOperations] Start flashing right firmware
[UhkOperations] Reenumerate bootloader
[UhkHidDevice] Start reenumeration, mode: Bootloader
[DeviceService] updateFirmware error {"message":"Cannot write to HID device","stack":"Error: Cannot write to HID device\n    at t.UhkHidDevice.<anonymous> (C:\\Users\\rober\\AppData\\Local\\Programs\\uhk-agent\\resources\\app.asar\\electron-main.js:100:32424)\n    at Generator.next (<anonymous>)\n    at s (C:\\Users\\rober\\AppData\\Local\\Programs\\uhk-agent\\resources\\app.asar\\electron-main.js:43:1740)"}
[UhkHidDevice] Transfer error:  Error: Cannot write to HID device
    at t.UhkHidDevice.<anonymous> (C:\Users\rober\AppData\Local\Programs\uhk-agent\resources\app.asar\electron-main.js:100:30188)
    at Generator.next (<anonymous>)
    at C:\Users\rober\AppData\Local\Programs\uhk-agent\resources\app.asar\electron-main.js:43:1938
    at new Promise (<anonymous>)
    at Object.l [as __awaiter] (C:\Users\rober\AppData\Local\Programs\uhk-agent\resources\app.asar\electron-main.js:43:1683)
[...]

PR #1330, opened 2020-08-19
feat: increase bootloader timeout

ert78gb:
issue: #1325


ert78gb:
I close it because we solve the problem in a different way

PR #1354, opened 2020-10-08
chore: upgrade node-hid => 2.0.0-0

ert78gb:


ert78gb:
wait until [node-abi](lgeiger/node-abi#90) fixes

PR #1370, opened 2020-11-09
#1360 Dark mode

### LINKED ISSUE #1360: Dark mode
While the Agent is nice it's all white and that makes some visually impaired users very uncomfortable.

There are standard dark modes supported by MacOS and Windows and there are standard CSS selectors which should be  available through electron. 

It would be great if you introduce dark mode support. And it should be relatively easy.

laxu:
Closes #1360 

# What's new

* Upgrades Bootstrap to 4.5.3 version.
* Upgrades various templates and styles to work with new Bootstrap version.
* Adds support for dark theme.
* Uses CSS variables for colors in UHK components.
* `ThemeService` for changing theme based on media queries or programmatically.
* Move `common.scss` to the proper `_common.scss` partial file.
* Move some global style definitions to `_global.scss`.
* Set a minimum width variable for `auto-grow-input` component so it never becomes unusably small if the user scales the screen. Defaults to 100px.
* Set minimum width for sliders in mouse speed etc tabs for the same reason as above.
* Improve macro action editor scaling to work better, mainly showing its buttons in a nicer way in narrow views. Before it would have arrow style buttons that made no sense if the screen was too narrow.

## Why the Bootstrap 4.5 upgrade?

* NG-Bootstrap version used for JS components is already supposed to be using BS 4.5.x.
* Bootstrap 4 supports SCSS so changing and using its variables, mixins etc works with UHK styles. This was essential to get dark theme working right as it relies on redefining BS color variables as well as using them for CSS variables used by UHK components.
* Only required parts of Bootstrap styles can now be loaded instead of the whole thing.
* BS4 upgrade only touches templates and styles so JS code and thus app functionality remains unaffected.

## How themes work

* The supported themes are light and dark but it would now be possible to add more themes with ease if so desired. 
* `_themes.scss` file defines colors for both themes. 
* Light theme is based on the colors that were present in current version of UHK Agent and where custom colors were not used, it uses BS4 defaults. 
* Dark theme redefines some of these colors to work with a dark background. I chose a cyan primary color because it was better visible against the dark background compared to the BS4 default blue. This can be changed from a single variable if needed.
* Separate bundles for each theme are needed because there is no way to override Bootstrap colors on the fly since they are defined as SASS/SCSS variables so they need to be set before styles are built. While BS4.5 defines some colors as CSS variables, refactoring Bootstrap itself to use them is slated for BS version 6 which is probably years away. So the separate bundles get around this issue.
* UHK SVG icons get their colors inverted. This looks fine to me and means we don't need separate icons for dark mode.

#### Requisites:
* Stop injecting the main styles in `angular.json`, instead make separate bundles for light and dark themes. The main difference in these files is a single SCSS variable `$dark-theme` which determines which theme's colors are in use.
* Load light theme as fallback in `index.html` and assign it the ID `uhk-theme` and attribute `data-theme="light"`. This is used to get the element easily and identify which theme stylesheet is loaded.

#### Switching themes:
* On app load singleton service ThemeService is launched.
* ThemeService checks for the `prefers-color-scheme: dark` media query.
* ThemeService attaches a listener for subsequent theme changes. This allows for changing the OS theme setting and having UHK Agent follow by changing its style on the fly!
* ThemeService changes the theme on app launch to dark if user prefers dark color scheme. Otherwise it does nothing as light theme is already loaded.
* If current theme is same as new theme, do nothing. This is determined by what theme you are trying to set vs the `data-theme` attribute in the currently loaded stylesheet element. If more theme support is desired in the future this could be moved to be read from app state instead.
* Theme can also be selected from Settings. If Light or Dark setting is chosen, OS preference is not followed. System default uses whatever the user set in their OS settings.

# Screenshots

image
image

image
image


mondalaci:
@laxu Thanks so much for the huge effort!

1. The new blue color used across Agent pops out way too much to the point that it feels it was designed for visually impaired users. I much prefer the old blue color.
2. Button borders are higher contrast than they used to be. I prefer the old, more subtle contrast, and it's also consistent with Select2 border styles.
3. Header styles, layer buttons, and the numeric inputs of the macro section are considerably larger than previously, and I prefer the old size.
4. There is too much space between side menu items. I prefer the previous spacing.
5. I can't find a Dark Mode checkbox under Settings, so I could only test Light Mode (for lack of a better name).

laxu:
@mondalaci fonts, colors and spacing can be tweaked, I can see what the old BS theme used and put them a bit closer to those. 

I tried to make sure that things adhere to Bootstrap font size variables so that they are easy to configure in the future but I think I have to tweak it more.

Previous side menu had inconsistent element spacing and highlighting so this fixes that too. Its spacing can be of course tightened a bit. Personally I like this spacing because it makes elements easier to click.

To test dark mode, set Windows or MacOS to use dark theme from its settings. Adding a menu option for this should be a followup issue as this is already a big enough PR.

mondalaci:
@laxu Thanks for the quick follow-up!

Being a Linux user, I can only test dark mode if there's a checkbox for it which I believe should be part of this PR.

The lack of a global dark mode setting on Linux makes the implementation of this Agent setting nontrivial. I propose Agent to use the global OS dark mode setting on Windows and OSX and show a settings checkbox on Linux.

Tweaking the styles would be much appreciated.

laxu:
@mondalaci I see, I was not aware that Linux had this sort of deficiency. Adding the menu options would be a good amount of extra work (UI, managing state of theme selection, making sure that it gets overridden correctly compared to the dark mode mediaquery) so I would rather do it as a follow up issue. 

Maybe you can test dark mode in web browser like this: https://stackoverflow.com/questions/56401662/firefox-how-to-test-prefers-color-scheme

packages/uhk-web/src/app/components/popover/tab/mouse/mouse-tab.component.scss
```diff
@@ -36,10 +36,10 @@
                             width: 0;
                             height: 0;
                             top: 0;
-                            right: -4rem;
-                            border-color: transparent transparent transparent $icon-hover;
+                            right: -37px;
```
ert78gb, line 39: Can't we use rem?

laxu: Fixed.

tyler-ham:
> Being a Linux user, I can only test dark mode if there's a checkbox for it which I believe should be part of this PR.
> 
> The lack of a global dark mode setting on Linux makes the implementation of this Agent setting nontrivial. I propose Agent to use the global OS dark mode setting on Windows and OSX and show a settings checkbox on Linux.

I suggest using the same UI across all platforms in the form of a three-option selection:

- Auto
- Light Mode
- Dark Mode

Auto is selected by default and would match the OS setting on Windows and OSX and activate Light mode on Linux. Selecting Light Mode or Dark Mode would set the mode explicitly regardless of the platform settings.

laxu:
@mondalaci Now spacing and font sizing should be better. Turns out I was importing BS variables in incorrect order. With SCSS `!default` variables the override needs to be defined first for it to be used right.

I also tightened the spacing of the side menu.

laxu:
@mondalaci @ert78gb If I were to add a theme setting, where should I store its state? Do we have a reducer that would be a good place for this?

packages/uhk-web/src/styles/_themes.scss
```diff
@@ -0,0 +1,158 @@
+$dark-theme: false !default;
```
ert78gb, line 1: Would not be better to use 2 theme file instead of 1? 

If we will use 2 files I see the following benefits
- the `$dark-theme` variable not needed
- the `@if $dark-theme` condition not needed, less indentation
- we can compare light and dark theme file much easier and we can see detect much easier the different things

laxu: That's a fair point. I was inventing this along as I went but they could definely be separate files at this point.

ert78gb:
Now I understand the scss processor does not support css variable in the internal function like `lighten` and `darken`. It would be better with BS5 until BS5 uses `mix` :(

ert78gb:
@laxu you can use the `packages/uhk-web/src/app/store/reducers/app.reducer.ts` reducer to store the light and dark mode settings. `animationEnabled` is a good example.

mondalaci:
@tyler-ham I also thought about providing these 3 options, but I'm unsure if we should expose this setting to Windows and Mac users. What options do other applications provide on Windows and Mac when it comes to dark mode?

@laxu Thank you, I can see the mentioned spacing improvement. However, I've just noticed that when I hover over the "Macros" side menu item, the "Go to UHK site in browser" macro name gets wrapped into 2 lines.

laxu:
@mondalaci The ones I have seen the most are "System preference / Light / Dark".

packages/uhk-web/src/styles/_themes.scss
laxu: They are now separated in their own themes folder.

tyler-ham:
@mondalaci, I'm also a Linux user, so I can't easily speak about typical Windows and Mac applications. However, I can offer as an example that the Gmail app on Android offers a Theme setting with three options:

- Light
- Dark
- System default

So there is precedent for offering these three options at the application level.

ert78gb:
The apps that developed by the Apple does not support theme changes at application level. All app use the system preferences.

Other apps support the changes, but not all they just use the system settings.
Source tree example
Screenshot 2020-11-09 at 23 00 05
Screenshot 2020-11-09 at 23 00 36

mondalaci:
Thanks, guys! Let's follow the best practices of other applications, then.

"Theme setting" dropdown:
- Light
- Dark
- System default

Given that there is no system default on Linux when it comes to dark mode, I think we shouldn't expose it on Linux. Consequently, the default option should be "System default" on Windows and Mac and "Light" on Linux.

packages/uhk-web/src/styles/_themes.scss
ert78gb: thx

packages/uhk-web/src/app/shared.module.ts
```diff
@@ -251,7 +252,8 @@ import { KeyActionDragAndDropService } from './services/key-action-drag-and-drop
         UhkDeviceUninitializedGuard,
         UhkDeviceLoadingGuard,
         UhkDeviceLoadedGuard,
-        UhkDeviceBootloaderNotActiveGuard
+        UhkDeviceBootloaderNotActiveGuard,
+        ThemeService
```
ert78gb, line 256: The `ThemeService` import is not needed because you use the `{ providedIn: 'root' }` options

packages/uhk-web/src/styles/_bootstrap.scss
```diff
@@ -0,0 +1,44 @@
+/*!
+ * Bootstrap v4.5.3 (https://getbootstrap.com/)
+ * Copyright 2011-2020 The Bootstrap Authors
+ * Copyright 2011-2020 Twitter, Inc.
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/main/LICENSE)
+*/
+
+@import '~bootstrap/scss/functions';
```
ert78gb, line 8: Should we need this file? I don't know the uncommented elements contains any sass variables that break the break the current build.

The tree shaker will remove the unused css classes. 
If we keep this file and BS changes the file structure we have to modify this file too. Now I just see an extra file that we have to maintain.

packages/uhk-web/src/styles-dark.scss
```diff
@@ -0,0 +1,12 @@
...
+@import './styles/common';
+@import '~angular-notifier/styles';
+@import './styles/tooltip';
+@import './styles/uhk-icons/uhk-icon';
+@import './styles/side-menu';
+@import './styles/ngx-select-ex';
+
+@import './styles/bootstrap';
+
+@import './styles/global';
```
ert78gb, line 12: The selected content same in the dark and light mode. Maybe worth move them in separated file and import this file in the light and dark mode css

ert78gb (changes requested):
Great job

packages/uhk-web/src/styles/_bootstrap.scss
laxu: We do need it, it's used by a lot of stuff.

laxu:
@mondalaci I added theme selector to settings. It will follow system preference unless Light or Dark is explicitly selected.

mondalaci:
@laxu I'd like to check it, but `npm run build` is unhappy at my side:

```
ERROR in /home/laci/projects/agent/packages/uhk-agent/src/services/app-update.service.ts
./src/services/app-update.service.ts
[tsl] ERROR in /home/laci/projects/agent/packages/uhk-agent/src/services/app-update.service.ts(149,13)
      TS2741: Property 'appTheme' is missing in type '{ checkForUpdateOnStartUp: true; everAttemptedSavingToKeyboard: false; }' but required in type 'ApplicationSettings'.
```

Based on the screenshot, please display "Theme:" with the dropdown on its right side.

laxu:
It should now run, didn't have time to do any other tweaks today.

mondalaci:
I can't see the dropdown on Linux.

image

laxu:
Added some fixes:

* Use Bootstrap 3 colors for light theme.
* Made theme changes work better, now it applies the new theme first, then removes previous one after 300ms. This prevents a flash of unstyled content.
* Moved theme imports to _main.scss file so they are not duplicated in different styles.
* Fixed elements moving around when hovering side menu.
* Added inverting colors of UHK icons if dark theme is in use so new icon files are not needed.
* Fixed theme selector becoming invisible.
* Put theme selector and label on single line.
* Removed unnecessary ThemeService import from shared module.

mondalaci:
@laxu Very impressed by your development pace! Theme switching works great on Linux now. There are a number of small issues left. Some of them:

- The text of the tooltips is invisible.
- Button border should look the same as Select2 border.
- Please use a Select2 for theme selection for the sake of consistency and to avoid OS-specific dropdown styling.
- The rightmost several pixels of the USB icon (next to the "My UHK" text) is not visible.
- Please make the Cancel button background color of the key action popover and the macro section white because it looks better on the surrounding gray background.
- The corners of slider widgets are missing.
- The smallest text size became too small which is heavily used in the key action popover.

laxu:
@mondalaci Requested fixes implemented. Changing border colors of buttons is a bit hacky because it's generated from the button color in BS4. In any case I added an override for the btn-outline-secondary used in many places so it looks as desired in the light theme.

mondalaci:
@laxu You're on a roll! Some more:

- The "System default" theme option shouldn't be present on Linux.
- The top corners of the key action popover are not rounded.
- These cancel buttons would pop nicely if they had white backgrounds:

image
image

- This tooltip is not visible entirely, and a scrollbar is added when displayed:

image

laxu:
@mondalaci I fixed the first 3 issues. Not sure what happens with the tooltip. I will take a look at it sometime later when I have time.

laxu:
@mondalaci I fixed tooltip issues and a couple other small problems. The tooltip issue existed on current release version of the app too, just was harder to trigger.

mondalaci:
- When dark mode is selected, Agent is white upon application startup and it visibly transitions to dark mode. Not a major issue, but it'd be nicer if Agent was dark from the get go.

Light mode issues:

- Please make the red buttons the same color as in the last Agent version.
- Make the hover color of buttons the same color as in the last Agent version.

Dark mode issues:

- Low contrast:

image
image

- Style issue around Ctrl and keys:

image

laxu:
@mondalaci Fixed the low contrast for the keymap tab, colors of the key icons and tweaked button colors. Also fixed some scaling issues with keypress tab so its fields and capture keystroke button should no longer become too small.

Restored the "default" button type which no longer exists in BS4. It looks about the same as it did previously now. 

Hover colors are generated from the main color of the button and it's a waste of time to try to make them exactly the same as BS3. I think they are just fine as they are now.

packages/uhk-web/src/styles/_bootstrap.scss
laxu: Tried this both ways. With our _bootstrap.scss stylesheets are 161 kB, with importing whole boostrap they are 189 kB. So maintaining our "just what we need" file is better.

mondalaci:
@laxu Thank you! Some further issues I found:

Some buttons don't turn to blue when clicked, including these:

image

Please implement button focus visualization in some other way, like as the previous Bootstrap version as the current implementation looks buggy in the following situation:

image

Modifier buttons should be in the same line as their label:

image

laxu:
@mondalaci The focus overlap seems to be a BS4 bug as it happens with their example pages too. Might even be a rendering bug as it seems to only happen when an element is after the highlighted one. I have removed the focus highlight box shadow for now.

The mouse tab buttons are now blue. This was caused by order specificity of CSS where the last defined element can override those preceding it in CSS when they are of the same level of value (in this case `btn-default` is the same rule value as `btn-primary` but defined later in styles so `btn-primary` did not override its style. The fix is a bit annoying but works.

I also fixed positioning issues in the keypress tab.

I would like to get this PR wrapped up and merged soon. Too many evenings spent on small fixes.

mondalaci:
@laxu I also want to wrap up this PR, but I will only merge it if it's devoid of any UI issues. I'm sure there aren't many left, and I found some more:

- On the mouse speed page, the width of the sliders change when moving the knob across the full range of the slider.
- Please make the tooltip of the modifiers of the key action popover and on the macro page point downward to not cover the modifier buttons.
- Pressing either of the Select2 dropdowns in the Layer section of the key action popover pushes them downward by a pixel.
- In dark mode, the text under Configuration history is very low contrast.

laxu:
@mondalaci I pushed a pile of fixes including the ones on your list. Hopefully that is all the issues.

mondalaci:
@laxu I hit a bug while testing your PR. I've tested a number of pages, then switched to dark mode, then eventually when clicking on a macro I got "There is no macro with this id." or something similar instead of the usual macro page content. Then I tried to switch back to light mode and nothing has happened. In the console, I saw this message:

```
19:18:25.844 › TypeError: Cannot read property 'modules' of undefined
    at Tg (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:816914)
    at userConfiguration (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:826744)
    at file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:1605723
    at file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:1606010
    at computeNextEntry (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:1383692)
    at recomputeStates (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:1384012)
    at file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:1386940
    at o.u.pipe.Object.state [as accumulator] (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:1387915)
    at o._tryNext (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:282837)
    at o._next (file:///home/laci/projects/agent/packages/uhk-agent/dist/renderer/main-es2015.194a43f48e70cf36058c.js:1:282738)
```

As for the cosmetic issues, there's an extra newline at the end of the scancode and secondary role tooltips, and the text of the confirmation popover is low contrast and isn't rendered with dark background in dark mode.

laxu:
@mondalaci Can you post a screenshot of the cosmetic issues? I am not quite sure what you mean or how to get the confirmation popover.

The macro thing sounds like you have stumbled upon some other bug. The error message doesn't really give anything to go on as just searching for "modules" in code gives a lot of stuff.

mondalaci:
image

As for the bug, I'll try to reproduce it, but feel free to disregard it if it's not relevant.

laxu:
@mondalaci Thanks for the screenshot, did not know where to look for that one. It was just the default Bootstrap popover did not have dark theme styles defined for them.  Also thanks for testing, I certainly become blind to a lot of things.

The tooltip issues were just default paragraph tag margins being applied, I gave them a more suitable margin.

mondalaci:
@laxu Thanks for the quick fix! I could only find one minor issue. When switching from a non-macro page to a macro page, it seems as if the "Add captured keystroke" button gets pressed momentarily.

I couldn't reproduce the "macro not found" bug.

@ert78gb When dark mode is active, and Agent starts up, it should be dark from the first moment instead of switching from light to dark. Can you make this work?

ert78gb:
I have multiple idea how to solve it, suddenly I don't know which is the better

laxu:
@ert78gb Just post what you can think of.

ert78gb:
The root cause is the default mode is light. After the angular app started will load the saved/calculated theme, have to move the theme loading into the index.html.

The solution that works in electron and web build:

- load the saved theme setting in the preload script. It may change the behaviour of the `webPreferences.nodeIntegration: true`, so maybe have to pass the IPC renderer to the renderer process too. For the better explanation the name of the javascript variable: `UHK_THEME` the values `LIGHT`, `DARK`, `SYSTEM_DEFAULT` 
- remove the light theme link from the `index.html`
- create an inline script in the `index.html` that based on the `UHK_THEME` value load the theme
- if the `UHK_THEME` does not exist then the user uses the web build so have to check the local storage for the saved settings.
- if the local storage settings do not exist then have to use the default theme
- extra task is the refactoring of the `ThemeService` to eliminate the code duplication of the inline script and `createStylesheetElement`

packages/uhk-web/src/app/components/agent/settings/settings.component.ts
```diff
@@ -22,11 +26,24 @@ import { ToggleAnimationEnabledAction } from '../../../store/actions/app';
...
+        { id: AppTheme.Dark, text: 'Dark' }
+    ];
 
     constructor(private store: Store<AppState>) {
         this.updateSettingsState$ = store.select(appUpdateSettingsState);
         this.animationEnabled$ = store.select(getAnimationEnabled);
+        this.appTheme$ = store.select(getAppTheme);
+
+        this.operatingSystem$ = store.select(getOperatingSystem);
+        this.operatingSystem$.pipe(filter(os => os === OperatingSystem.Linux)).subscribe(() => {
```
ert78gb, line 44: Selectors are the best place to calculate logics and keep the components simple as possible. Please move the `themes` calculation to a selector.
If the refactored version you also use a subscription then please unsubscribe from it on the `ngOnDestroy`

packages/uhk-web/src/app/services/theme.service.ts
```diff
@@ -0,0 +1,98 @@
...
+    darkModeListener$: Observable<MediaQueryListEvent>;
+    darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+
+    constructor(
+        @Inject(DOCUMENT) private document: Document,
+        private store: Store<AppState>,
+        private logService: LogService
+    ) {
+        this.appTheme$ = this.store.select(getAppTheme);
+        this.appTheme$.subscribe(value => this.setTheme(value));
```
ert78gb, line 29: Please implement the `OnDestroy` interface and unsubscribe. I know it is a singleton in the root DI, but I like if we properly release the resources, prevent us from the future memory leak investigations

ert78gb (changes requested):

ert78gb:
@laxu If only the better dark mode loading missing I can implement my idea if you prefer it after @mondalaci merged this PR.

richrd:
This looks really nice, great work! Can't wait to start using this :heart_eyes: .

I just wanted to point out that some Linux distros do have a global light/dark mode setting, such as Ubuntu. Though I'm not sure if Electron detects it but it seems like it probably does https://github.com/electron/electron/issues/23861

image

packages/uhk-web/src/app/components/agent/settings/settings.component.ts
laxu: Done.

packages/uhk-web/src/app/services/theme.service.ts
laxu: Done.

laxu:
> The root cause is the default mode is light. After the angular app started will load the saved/calculated theme, have to move the theme loading into the index.html.
> 
> The solution that works in electron and web build:
> 
> * load the saved theme setting in the preload script. It may change the behaviour of the `webPreferences.nodeIntegration: true`, so maybe have to pass the IPC renderer to the renderer process too. For the better explanation the name of the javascript variable: `UHK_THEME` the values `LIGHT`, `DARK`, `SYSTEM_DEFAULT`
> * remove the light theme link from the `index.html`
> * create an inline script in the `index.html` that based on the `UHK_THEME` value load the theme
> * if the `UHK_THEME` does not exist then the user uses the web build so have to check the local storage for the saved settings.
> * if the local storage settings do not exist then have to use the default theme
> * extra task is the refactoring of the `ThemeService` to eliminate the code duplication of the inline script and `createStylesheetElement`

This sounds good to me. @ert78gb You can implement it if you have a good grasp on how to do it. 

I have changed the `AppTheme` enum and code that uses it to follow Electron's convention of "system", "light", "dark". Seems like `nativeTheme.themeSource` should be used for this. https://www.electronjs.org/docs/api/native-theme#nativethemeshouldusedarkcolors-readonly

I have implemented OnDestroy for ThemeService and moved getting theme select items in Settings to a selector.

@mondalaci Should we restore the System default option on Linux so it can work on Linux versions that happen to support it like @richrd mentioned? I see no real drawback for having the menu option, it will default to light theme if there is no preference or use whatever you select in Settings.

mondalaci:
@laxu Good question! Do you or @richrd have a good understanding of how the dark theme setting works behind the scenes on Linux and which distros support it? It'd make sense to have a system default option on Linux, too, but I'm afraid that it'd confuse many users.

ert78gb (approved):

ert78gb:
> This sounds good to me. @ert78gb You can implement it if you have a good grasp on how to do it.

OK, I will do it in separate PR after it merged

laxu:
@mondalaci I tried googling for the state of dark theme support on Linux but can't find anything conclusive on whether they work with the `prefers-color-scheme` media query or Electron's methods for getting this. ArchLinux and Ubuntu might support it. There does not seem to be a mechanism for determining if the OS has this support or not.

My opinion is that we should just leave the "System default" option available regardless of OS and it is either supported or not. It will just default to light theme if there is no support for it so no harm done. If that does not match what the user prefers then they can simply set it to dark theme themselves.

mondalaci:
@laxu Yes, please add the System default option for Linux, too, and add a tooltip on Linux with the content of:

> Following operating system theme. The "System default" option may not be supported on all Linux distributions.

laxu:
> @laxu Yes, please add the System default option for Linux, too, and add a tooltip on Linux with the content of:
> 
> > Following operating system theme. The "System default" option may not be supported on all Linux distributions.

Should I rename the settings option to "Follow operating system theme"? That would be clearer for all platforms. Then only Linux would show the tooltip telling that it might not be supported.

mondalaci:
@laxu Yes, please do.

laxu:
@mondalaci Implemented tooltip next to theme selection. It is only shown on Linux.

mondalaci:
@laxu Thank you so much for your hard work! I deeply appreciate your perseverance, attention to detail, and rapid feedback. You've made dark mode happen.

laxu:
@mondalaci And thanks for having more patience testing and getting it right than I sometimes did! :D

Looking forward to starting to use dark mode in a future release.

richrd:
I'm on Kubuntu 20.04 and I'm very happy to let you know that verison 1.5.6 of Agent autodetects that I'm using a dark theme on the system and defaults to that :relaxed: So the "Follow operating system theme" works great here!

PR #1402, opened 2020-12-22
chore: upgrade node-hid => 2.1.1

ert78gb:


ert78gb:
@mondalaci can you check the agent starting on your machine?
for this new node-hid version needed gcc 9

PR #1410, opened 2021-01-05
fix: remove the auto update limitation at first start of the Agent

ert78gb:


mondalaci:
Please test it and merge it if it works.

Issue #1428, opened 2021-02-07
Update module firmwares


mondalaci:
On the Firmware page, list the connected modules and their firmware versions below the left and right keyboard halves, such as "Firmware 8.10.4 is running on the key cluster module."

During the firmware update process, query the list of connected modules and update them.

The touchpad is a special case because its firmware update mechanism and protocol differ from the rest of the modules. Display "The touchpad module is present, and its firmware is not updatable yet."

PR #1467, opened 2021-03-15
feat: update module firmwares

ert78gb:
closes: #1428

BREAKING CHANGES:
- the `<moduleSlot>` argument replaced with `<i2cAddress>` of the `update-module-firmware.ts` USB script changed. Hexadecimal and decimal format supported.

### DIFF
--- a/packages/uhk-common/src/models/module-slot-to-i2c-adress.ts
+++ b/packages/uhk-common/src/models/module-slot-to-i2c-adress.ts
@@ -1,0 +1,7 @@
+export enum ModuleSlotToI2cAddress {
+    leftHalf = 0x10,
+    leftModule = 0x20,
+    rightModule = 0x30,
+    leftIqs5xx = 0x4B,
+    rightIqs5xx = 0x6D
+}
--- a/packages/uhk-common/src/models/module-slot-to-i2c-adress.ts
+++ b/packages/uhk-common/src/models/module-slot-to-i2c-adress.ts
@@ -2,6 +2,6 @@ export enum ModuleSlotToI2cAddress {
     leftHalf = 0x10,
     leftModule = 0x20,
     rightModule = 0x30,
-    leftIqs5xx = 0x4B,
-    rightIqs5xx = 0x6D
+    leftTouchpad = 0x4B,
+    rightTouchpad = 0x6D
 }
--- a/packages/uhk-common/src/models/uhk-products.ts
+++ b/packages/uhk-common/src/models/uhk-products.ts
@@ -1,13 +1,7 @@
 import { ModuleSlotToI2cAddress } from './module-slot-to-i2c-adress';
 import { ModuleSlotToId } from './module-slot-id';
 
-export enum UhkProductTypes {
-    Device = 'Device',
-    Module = 'Module'
-}
-
-export interface UhkProduct {
-    type: UhkProductTypes;
+export interface UhkDeviceProduct {
     id: number;
     // TODO: Maybe it is not necessary
     name: string;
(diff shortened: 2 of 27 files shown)


packages/uhk-common/src/models/module-slot-to-i2c-adress.ts
```diff
@@ -0,0 +1,7 @@
+export enum ModuleSlotToI2cAddress {
+    leftHalf = 0x10,
+    leftModule = 0x20,
+    rightModule = 0x30,
+    leftIqs5xx = 0x4B,
```
mondalaci, line 5: Rename to `leftTouchpad` and `rightTouchpad`.

packages/uhk-common/src/models/uhk-products.ts
```diff
@@ -45,3 +48,77 @@ export const UHK_DEVICES: Array<UhkDeviceProduct> = [
...
+
+export const LEFT_KEY_CLUSTER_MODULE: UhkModule = {
+    type: UhkProductTypes.Module,
+    id: 2,
+    name: 'Left Key Cluster',
+    slotId: ModuleSlotToId.leftModule,
+    slotName: 'KeyClusterLeft',
+    i2cAddress: ModuleSlotToI2cAddress.leftModule,
+    firmwareUpgradeSupported: true,
+    bootloaderPingReconnectMsg: 'Cannot ping the bootloader. Please remove the "Left Key Cluster" module, and keep reconnecting it until you see this message.'
```
mondalaci, line 82: Is there an easy way to not repeat this for every module?

packages/uhk-common/src/models/uhk-products.ts
```diff
@@ -45,3 +48,77 @@ export const UHK_DEVICES: Array<UhkDeviceProduct> = [
...
+    slotName: 'LeftKeyboardHalf',
+    i2cAddress: ModuleSlotToI2cAddress.leftHalf,
+    firmwareUpgradeSupported: true,
+    bootloaderPingReconnectMsg: 'Cannot ping the bootloader. Please remove the bridge cable, and keep reconnecting the left keyboard half until you see this message.'
+};
+
+export const LEFT_KEY_CLUSTER_MODULE: UhkModule = {
+    type: UhkProductTypes.Module,
+    id: 2,
+    name: 'Left Key Cluster',
```
mondalaci, line 77: The modules are one-sided, so please don't write left or right. Only capitalize the first word. So this is: `Key cluster`

packages/uhk-web/src/app/components/device/firmware/device-firmware.component.html
```diff
@@ -8,8 +8,12 @@ <h1>
             </h1>
 
             <p>
-                Firmware {{ hardwareModules.leftModuleInfo.firmwareVersion }} is running on the left keyboard half.<br>
-                Firmware {{ hardwareModules.rightModuleInfo.firmwareVersion }} is running on the right keyboard half.
+                <ng-container *ngFor="let moduleInfo of hardwareModules.moduleInfos; trackBy:moduleInfoTrackByFn">
```
mondalaci, line 11: Please rather write:
Left keyboard half firmware: 8.10.7

Display the right keyboard half first.

packages/uhk-common/src/models/uhk-products.ts
ert78gb: I removed the `type` and `slotName` properties because we don't use it anymore.
The `firmwareUpgradeSupported` is a feature flag that helps to decide Agent supports the firmware update of the module or not e.g. touchpad currently
I would like to remove the `bootloaderPingReconnectMsg` property if we can figure it out an error message that meaningful and contains a place holder for the module name like: `Cannot ping the bootloader. Please remove the "${moduleName}" module, and keep reconnecting it until you see this message.`. I use this distinction because the  message of the left half is different `Cannot ping the bootloader. Please remove the bridge cable, and keep reconnecting the left keyboard half until you see this message.`

mondalaci: I only meant to generalize `bootloaderPingReconnectMsg` for modules. The suggested string is good.

mondalaci:
For the touchpad, display:

> Touchpad firmware version is unknown and it's not updatable yet.

For the other firmwares, don't put a `.` character at the end:

> Key cluster firmware: 8.10.7

PR #1488, opened 2021-04-18
Feature: Handle reorder of USB mgmt interface 

steamraven:
Handles reordering USB Management interface.  Will try to detect based on HID Usage if available.   On some Linux setups, usage may not be available, then defaults to probing interfaces, since we know that this is a UHK keyboard.

This needs testing before merge. Please Test and report!


mondalaci:
Your PR works as intended! Excellent work, but I'd rather wait until the USB reordering firmware branch proves to be actually useful based on the feedback of others because I want to make sure that this PR is justified.

sunghospark:
@mondalaci as mentioned in https://github.com/UltimateHackingKeyboard/firmware/issues/287 the usb re-ordering turned out to be helpful for KVM issues. It would be great if this PR can be merged checked against latest and merged🙏

mondalaci:
Just commented the referenced firmware issue. I agree that we should fix this issue.

Issue #617, opened 2018-04-30
Add more layers


mondalaci:
Currently, the Base, Mod, Mouse, and Fn layers are available. We'll extend these layers with 4 regular layers (Fn2, Fn3, Fn4, Fn5) and 4 modifier layers (Shift, Ctrl, Alt, Super).

All of these layers will be optional (can be added or removed per keymap) with the exception of the base layer which cannot be removed.

See the updated layer button group in which the order of Mouse and Fn is exchanged compared to the current order, and there's a dropdown arrow at the end:

![image](https://user-images.githubusercontent.com/300402/73209672-0a554580-4149-11ea-85c2-240a75a175f8.png)

The dropdown contains a checkbox for every optional layer:

![image](https://user-images.githubusercontent.com/300402/73210665-04f8fa80-414b-11ea-89d5-f3638fec60d6.png)

[The implementation of the above dropdown](https://mdbootstrap.com/snippets/jquery/laci/1684060) is not consistent with the Bootstrap 3 style as it's only meant to be a demonstration. The dropdown must not disappear upon clicking a checkbox. Unchecking a checkbox must be confirmed by a confirmation popover containing: *Do you really want to delete this layer?*

Modifier layer names must be OS-specific. So for example, Alt becomes Cmd on Mac.

The layer button group must update immediately upon toggling the layer checkboxes, and the button order must be the same as the dropdown item order.

The user configuration format will change. Right now, a keymap contains a fixed number of 4 layers. According to the above, a keymap will contain at least 1 and at most 12 layers. A layer id must be included into each layer object.

eltang:
I believe the most flexible way to implement this is to allow a key to trigger two actions (one two activate the modifier, one to switch the layer). However, it would require that keymaps be able to contain more layers than the current four.

mondalaci:
We're on the same page. The way to implement this is to make keymaps contain 4 more layers for the 4 modifiers. The firmware would switch to the relevant modifier layer if there's a key defined on that modifier layer which is not none action.

piratk:
If you intend to update the layers vs keymaps, perhaps make it possible to have the same layer in different keymaps.

mondalaci:
@piratk I'm not sure whether you mean the same layer with different content in different keymaps, or the same layer with the same content in different keymaps. The former will be implemented. I'm not comfortable with the latter because it'd make things more compicated for users, especially newcomers.

piratk:
[...]

PR #1548, opened 2021-07-11
feat: add more layers

ert78gb:
close: #617


kareltucek:
I wanted to try this branch out, but I am getting this result:

2021-08-21-135713_2508x2105_scrot

Any ideas what I am doing wrong?

mondalaci:
@kareltucek You are doing everything right. Let me talk to @ert78gb about when should the update screen trigger, and possibly how to ignore it.

ert78gb:
If the user config version of the right module is greater than the userConfig version of the agent then this screen will visible.
If you build agent from source you can change the `userConfigVersion` in the root package.json file

kareltucek:
Allright, got it running!

Some notes:
- When I add fifth layer (fn2), its checkbox does not get checked until I switch to another keymap and back.
- If I merge command-action branch with more-layer branch, run agent, and try to load config that contains some macro commands, agent does not like it. The config seems to load fine, but save button does not appear and console is full of logs:

2021-08-22-233504_3840x2160_scrot

ert78gb:
Both PR have affect to the user-config I don't know how compatible them. Would we like to merge them at once or do we merge them incrementally?

kareltucek:
Incrementally, unless Laszlo changes his mind. Which means that it is indeed nonissue at the moment.

mondalaci:
@kareltucek I'm running [common trunk](https://github.com/UltimateHackingKeyboard/firmware/pull/420), and the extra layers work as intended, but the modifier layers don't trigger the mapped actions. Are they supposed to?

We commented out the force update screen that was present in this PR previously. I didn't originally envision the configuration parser to be backward-compatible which justified it, but now that it's backward-compatible, we'll probably remove the force update Agent screen.

I can't reproduce the Fn2 checkbox issue you mentioned. Can you? Command actions work fine, too.

ert78gb:
How could it be a backward compatibility changes? We inserted the layer id/name field in the middle of the binary structure.

mondalaci:
@ert78gb The configurator parser of the firmware branches according to the user config version number.

kareltucek:
@ert78gb user config version is part of flashed eeprom image, so the firmware can interpret it accordingly.

@mondalaci as far as I remember, added layers are not par of common_trunk. You need common_trunk_upstream, or merge the layer pr separately and wait till it propagates into common_trunk.

> I didn't originally envision the configuration parser to be backward-compatible which justified it, but now that it's backward-compatible, we'll probably remove the force update Agent screen.

Well, this is an opportunity to try out the migration process in practice. If we continue to treat it as a noncompatible upgrade, it will make future config upgrades easier for us.

kareltucek:
(To clarify, I have no set-in-stone intention to maintain the parser compatibility across all future config version upgrades - it just turns out that life is easier with backward compatible parser when there is no strong reason to break the compatibility.)

mondalaci:
@kareltucek I much prefer backward compatibility, assuming that the maintenance burden is light which is according to the code. We can burn the bridges later if the maintenance burden becomes less tolerable or if we run out of flash space.

@ert78gb Please remove the force update Agent screen because we don't need it. Sorry for the extra work!

ert78gb:
np the extra work. I just would like to double check to remove this screen is a good idea.
Currently, Agent parses the user configuration based on the major version number of `userConfigVersion` of the right module.
Agent can only parse those user configurations which were previously implemented. I think Agent should notify the user that a newer version of the Agent will able to parse the user configuration.

The "Update Agent" page is visible because the value of `userConfigVersion` in the package.json is `4.1.0`. The current firmware  returns with `4.2.0` user config version . There is a bug, because only the major version numbers should be compared. I am gonna fix it.

I have 3 use cases:
**Case 1:**
User has multiple computers.
- User upgrades the Agent to the latest version on Machine A
- Upgrade firmware on Machine A
- Connect the keyboard to Machine B
Agent should notify the user about the user config incompatibility.

**Case 2:**
Downgrade Agent
- User upgrades the Agent
- User upgrades the firmware
- User downgrades the Agent
Agent should notify the user about the user config incompatibility.

**Case 3:**
Upgrade with newer firmware
- User downloads the latest alfa firmware
- Upgrade the keyboard with the latest alfa firmware
Agent should notify the user about the user config incompatibility.

mondalaci:
@ert78gb You're right, but I have some things to add.

1. We should check not only majorVersion, but majorVersion.minorVersion because [increasing the minor version extends the configuration with new values](https://github.com/UltimateHackingKeyboard/firmware/blob/master/VERSIONING.md).
2. We should only display the Agent update screen if Agent is unable to parse the user config. A new Agent version may only extend the minor version, but the new configuration values may not be used by the current UHK configuration, so let's not bother the user, only when necessary. The top Agent notification bar will appear anyways, so the user can update Agent if wanted.
3. It's possible that we break an Agent release, and some users want to use an older Agent version until we fix the bug. This should happen quite rarely, but it would be useful to be able to ignore the Agent update screen. Please implement the `--disable-agent-update-screen` for this purpose. This option should also be useful for testing and comparing older Agent versions.

PR #1632, opened 2021-11-24
Add smart macro versioning queries.

kareltucek:


mondalaci:
@ert78gb Please merge it if you have no objections.

ert78gb:
The code looks good to me.
My question is the backward compatibility. I checked only on Mac with the `8.10.12` firmware and the response of the protocol version property command is 64 bytes. I don't know the firmware sends 64 bytes from the beginning or the USB protocol extend it to 64 bytes. If the USB protocol then I think worth to read `smartMacrosVersion` if the `firmwareVersion` greater than `X`.

Issue #660, opened 2018-05-30
Customize LED segment display


MattKunze:
Currently the LED segments show the currently selected keymap ('QWM', etc). For me at least this is something that will never change, so the display doesn't really provide much information. It would be awesome if these could be overridden with a console utility (`uhk -leds 'FOO'`) to display arbitrary text.

Why? This is the ultimate *hacking* keyboard, so mainly it just seems like stuff like that should be hackable ;) I could see writing a script to update it with the current CPU load, polling the outside temperature, hooking it up to the current terminal to display additional status, who knows.

Not sure if this belongs here or the firmware repo, thought I'd get the discussion started somewhere though.

mondalaci:
Good point, and this is something that we planned. Now we also have an issue for it. Will do eventually. :)

luteijn:
To expose the LED (and future expansions, like e.g. an i2c lcd display or other gadget) it might be useful to have the uhk also present itself as a usb-serial converter so scripts have something straight-forward to talk to. Alternatively/additionally agent could act as a proxy, but I guess you already have some ideas for this.

mondalaci:
@luteijn Yep, the plan is to expose Agent as a proxy. Regarding communication, the already existing generic USB HID interface will be used. It's binary, but creating an additional USB serial interface would create a lot of bulk.

mondalaci:
@richrd Yep, the LED display segments will be individually controllable. This is much needed for showing characters that resemble the predator alphabet, and who are we to stop you from doing that. :)

So glad you're having fun with your UHK!

lrascao:
👍 i would also like to see this, btw great job!

Spark3000:
Awesome.

norrs:
This would be awesome, could use it for showing current xkbkeymap language, ie if Im currently using english keyboard or native norwegian.

FatBoyXPC:
I think it would be neat if we could include some scripts to be stored on the keyboard, triggerable by the keyboard as well. Similar to how you can change layouts with the keyboard.

mondalaci:
@FatBoyXPC This is not planned. Generally speaking, Agent is only meant to configure the UHK. You can achieve what you're after by setting up an unused shortcut (such as Ctrl+Alt+Shift+F13) for a key and bind it to your window manager to execute a script.

FatBoyXPC:
[...]

PR #1813, opened 2022-07-30
Add legend for LED display font

danpalmer:
This PR is a work in progress, and a request for feedback on whether this feature would be welcomed, and on the approach taken.

While configuring my UHK I was excited to find that the LED display is quite flexible with many characters supported. However, exploring this took some trial and error. I thought a legend/key for the font would be a nice feature, to see what characters are supported and what each character looks like without flashing the keyboard.

Currently this is implemented as an extra page in the sidebar for demo purposes, but I think this would work well as a medium sized popover next to the input field for the LED display text. Additionally it might be nice to overlay the LED display component on the keyboard diagram with the selected text.

Thoughts and ideas welcome!

Screenshot 2022-07-30 at 15 27 29


mondalaci:
Thanks for your contribution!

How would you exactly use this feature? I plan to [expose LEDs via a USB API](https://github.com/UltimateHackingKeyboard/firmware/issues/322), including the per-key RGB backlight and the LED display.

Although a nice touch, I feel uneasy about displaying the LED display in Agent, especially considering that future UHK will probably have much more sophisticated displays, and replicating them in Agent would take quite an effort.

danpalmer:
@mondalaci My primary use-case here is having some sort of guide for what characters are available and what they'll look like.

As for how this is actually integrated, I'm open to ideas. I think a small popover next to the input may be the most subtle, while still providing some value. Alternatively perhaps there could be a live preview or something like that. Another option is to perhaps hide it somewhere in the "Help" screen. I don't think this needs to be prominent in any way, in fact I think the more out of the way the better so that it's only there for reference when really needed.

What do you think? If you think the guide is not a feature that makes sense in the agent that's absolutely fine, I'm happy to close this PR. If you think it could be useful somewhere though I'm happy to iterate and see if we can find somewhere good for it.

mondalaci:
@danpalmer Thanks for elaborating on your use case! Based on customer feedback, we can see that some features are requested, but nobody has ever wanted to preview the content of the LED display, so I'd rather not merge this PR.

[This issue](https://github.com/UltimateHackingKeyboard/agent/issues/660) is the most relevant that is highly requested.

Given the above, I'm closing this PR. We appreciate your contribution regardless!

danpalmer:
Cool, thanks for the consideration. I probably won't take a look at that issue as it's a bit more than I have time for and not something I'm personally interested in. I'll keep an eye open for other things while using the software though and may contribute more as I find things. It seems like a nice codebase to work in!

Issue #1804, opened 2022-06-27
Adding more than 255 macros freezes application


provinzio:
When trying to add more than 255 macros (just clicking the + button), the app freezes eventually and does not respond anymore.

A restart is necessary.

I am using the newest agent (freshly downloaded and installed, just got my UHK ❤️ ).

If no more than 255 macros are supported, an error should be raised instead.

Edit: I use [kareltucek/firmware](https://github.com/kareltucek/firmware)

kareltucek:
I can clearly reproduce this with development Agent build (a few weeks old). Following message is shown in terminal:

```
19:33:02.263 › He.id: Integer 256 is outside the valid [0, 255] interval
```

mondalaci:
@provinzio Thanks, we should fix this. What's your use case? 255+ macros is a lot.

provinzio:
I am currently implemented the VOU-layout (https://www.maximilian-schillinger.de/vou-layout.html) and need a lot of Alt Codes for that. So I started by mashing the plus button and ended up freezing the agent.

I thought of adding all Alt+Codes first to have everything setup just in case, but I don't need that much in action. So for my usecase, it should be enough to keep the cap of 255 macros.

kareltucek:
I am still wondering if we should support unicode chars via alt-codes as a native action eventually. Not a huge fan, but it definitely is a repeatedly requested feature...

kareltucek:
> @kareltucek How often can you recall users facing this limitation? Of these cases, how many are unicode-related?

Specifically number-of-macro bounded limitation never.

Alt-code related requests turn up time to time. Maybe once a year? Requests like the Neo2 keymap, and various requests where it often is unclear whether the person just doesn't understand the difference between character and scancodes, or is actually unhappy about being limited to scancodes...

The fork now supports `tapKeySeq` as some alt-code solution was explicitly requested by multiple people. It allows typing emoji like `tapKeySeq CS-u 1 f 6 2 2 space` (linux dialect).

> I am wondering if we should 

So... probably not.

provinzio:
> or is actually unhappy about being limited to scancodes...

As in my case, as far as I understood, I am unable to write (all) unicode characters on Windows. e.g. emojis, greek letters, most math symbols.

I'd love to be able to tell the UHK to type these without using a third software (e.g. Windows ComposeKey). I don't want to install the software on all machines...

kareltucek:
Yes, your usecase is quite unambiguous.

PR #1815, opened 2022-07-31
feat: don't allow to create more than 255 macro

ert78gb:
close: #1804

### DIFF
--- a/packages/uhk-web/src/app/components/side-menu/side-menu.component.html
+++ b/packages/uhk-web/src/app/components/side-menu/side-menu.component.html
@@ -144,8 +144,12 @@
                     Macros
                     <button (click)="addMacro()"
                             class="btn btn-sm btn-secondary float-right"
-                            [class.disabled]="state.updatingFirmware">
-                        <fa-icon [icon]="faPlus"></fa-icon>
+                            [disabled]="state.updatingFirmware || state.macros.length > maxAllowedMacros">
+                        <fa-icon [icon]="faPlus"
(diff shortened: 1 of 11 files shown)


packages/uhk-web/src/app/components/side-menu/side-menu.component.html
```diff
@@ -144,8 +144,12 @@
                     Macros
                     <button (click)="addMacro()"
                             class="btn btn-sm btn-secondary float-right"
-                            [class.disabled]="state.updatingFirmware">
-                        <fa-icon [icon]="faPlus"></fa-icon>
+                            [disabled]="state.updatingFirmware || state.macros.length > maxAllowedMacros">
+                        <fa-icon [icon]="faPlus"
+                                 [ngbTooltip]="state.macros.length > maxAllowedMacros ? 'No more than 255 macros are supported.' : null"
```
mondalaci, line 149: Make the string contain ${maxAllowedMacros} to avoid duplication and set its value to 254.

ert78gb: done. And I replaced the `>` condition to `>=`

mondalaci:
Actually, maxAllowedMacros = 255 was fine, so please restore it. I hit a related bug, which can be reproduced as follows:

1. Create as many macros as possible.
2. Delete some macros.
3. Add as many macros as possible.
4. At some point, you'll be able to click on the add macro button, even though it's greyed out, and the following error will appear:
> 15:37:47.931 › He.id: Integer 256 is outside the valid [0, 255] interval

Also, the duplicate macro button should be inactive when reaching the macro limit, and the same tooltip text should appear for it as for the add macro button.

PR #1908, opened 2022-11-13
feat: allow 65 seconds as max macro delay

### LINKED ISSUE #1867: Macro delays longer than 10 seconds
This is something I am aware of for a long time but I never found time to report it.
When I am trying to add a longer delay inside a macro, if I for example try to set a delay of 100 seconds, it results in 10 seconds being added instead.
I haven't tried to modify configuration file manually to set a delay and import it but there must be a reason this is not available.
Right now I am solving this problem by creating loops with registers or just multiplying the same line over and over.

ert78gb:
closes #1867

### DIFF
--- a/packages/uhk-web/src/app/components/macro/action-editor/tab/delay/macro-delay.component.html
+++ b/packages/uhk-web/src/app/components/macro/action-editor/tab/delay/macro-delay.component.html
@@ -10,7 +10,7 @@
                 <input #macroDelayInput
                        type="number"
                        min="0"
-                       max="1000"
+                       max="65"
                        step="0.1"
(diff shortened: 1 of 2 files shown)


kareltucek:
I am quite confused by this PR. It does not seem to change the behaviour at all.

To be clear, the original problem (I believe) is that the clamping algorithm seems to do crazy and unintuitive things:

2022-11-13-213804_2437x682_scrot
2022-11-13-213812_2496x205_scrot

Similarly:
70 -> 7
100 -> 10
65 -> 65

I would expect:
70 -> 65
100 -> 65
123 -> 65

PR #1970, opened 2023-02-11
feat: Per-key colors

### LINKED ISSUE #1688: Per-key colors
### User configuration

Per-key colors will modify the user configuration, bumping its major version.

Let's add `uint8 backlightingMode` after `keyBacklightBrightness` in `UserConfiguration`. The valid enum values for this variable are `FunctionalBacklighting` and `PerKeyBacklighting`.

Following `backlightingMode`, add 8 RGB colors according to https://github.com/UltimateHackingKeyboard/firmware/blob/ff307571bcb38f54977daadbecb41455b19f642a/right/src/ledmap.c#L13-L20 and use these values as defaults. RGB colors are composed of 3 uint8 values.

Also add `ledColors[]` after `keyActions[]` in `Module`, and make it contain as many RGB values as keyActions if `backlightingMode === PerKeyBacklighting`, otherwise make its length 0.

### Settings UI

Rename the "LED brightness" page to "LED settings" including its header, filename, and side menu name.

Put the 3 sliders of the page next to each other in a 3 x 1/3 layout and add a "LED brightness" h3 header above them.

Below the sliders, add a "Backlighting mode" h3 header, and add the following radio buttons:

- Functional backlighting
- Per-key backlighting

Add a "Functional backlighting colors" h3 header below the above, list the referenced 8 color names in a 4x2 grid. Make borders around the colors, and adjust their background color according to the set colors. Set the text color of these rectangles to be the inverse of the background colors. Set the cursor style to pointer above the rectangles and open a color picker when clicking them. I like [this color picker](https://mdbootstrap.com/docs/b4/jquery/plugins/color-picker/), but others may also work.

### Color editor UI

When per-key backlighting is enabled, add a button group on the keymap page rightward of the layer button group.

Make the button group responsive, and position it under the layer button group there's not enough horizontal space available.

Display "Colors: " leftward of the button group, just as in the case of "Layers: ". Make the first 8 buttons of the button group square-shaped and make their background contain their relevant colors. Let's default to the colors of the functional backlighting, but these colors don't affect functional backlighting in any way; they're just defaults. Make these palette colors part of Agent's configuration.

Rightward of the color squares, add a font awesome plus button in the button group. When hovering on the plus button, display a tooltip for it saying "Add color to the palette.". When clicking on the plus button, invert it and display a "Click on a key to add its color to the palette." tooltip. When clicking on a key, add a new color to the palette, make the tooltip disappear, and de-invert the plus button.

When selecting a color from the palette,
1. put a font awesome check icon into it the palette color
2. animate the color of all keys of the UHK to their relevant color, and add a gray border around them
3. add a new button to the color button group displaying "Finish editing"

In color editing mode, after the plus icon, add a trash icon to the button group. When hovering over it, display the tooltip "Delete current palette color.", and make it behave accordingly.

In color editing mode, left-clicking on a key changes its color to the current palette color. The user can hold the left mouse button and paint multiple keys this way.

When clicking on the "Finish editing" button, restore the 3 changes listed above.

### Out of space UI

When backlighting mode is set to functional backlighting but the configuration contains per-key backlighting that the user has previously set, display a yellow (bootstrap warning color) "Recover LED space" button between the "Import device configuration" and "Reset device configuration" buttons. When hovering over the button, display the following tooltip "Per-key backlighting was previously active, but it's not used anymore, and per-layer LED colors still occupy {space} bytes on-board storage which you can recover.". When clicking on the button, display a confirmation popover with the "Do you want to recover {space} bytes on-board storage by deleting per-key backlighting colors?" message.

When the bottom out of space notification bar appears, instead of the "Delete some keymaps or macros to be able to save your configuration." sentence, display "[Switch to functional backlighting] and [recover per-key backlighting space], or delete some keymaps or macros to be able to save your configuration." - If per-key backlighting is active, display the first part and link it to the LED settings page. If per-key backlighting is present in the configuration, display the second part and link it to the configuration page. If only the second part is visible, don't include "and" and capitalize "recover".

ert78gb:
closes #1688


RegisHubelia:
Can't wait to get this feature!

highend:
Any idea when this will be merged?

mondalaci:
Firmware support is needed first. I think I'll have time to implement it in a month or two.

sajattack:
🎉

PR #1973, opened 2023-02-27
chore(web): resolve path to node_modules with webpack alias

NickCao:
This is the first in a series of PRs to migrate from lerna to native npm workspaces.

### DIFF
--- a/packages/uhk-web/angular.json
+++ b/packages/uhk-web/angular.json
@@ -28,11 +28,11 @@
             "optimization": false,
             "assets": [
               "src/assets",
-              { "glob": "**/*", "input": "node_modules/monaco-editor/min", "output": "assets/monaco-editor/min" },
-              { "glob": "**/*", "input": "node_modules/monaco-editor/min-maps", "output": "assets/monaco-editor/min-maps" }
+              { "glob": "**/*", "input": "monaco-editor/min", "output": "assets/monaco-editor/min" },
+              { "glob": "**/*", "input": "monaco-editor/min-maps", "output": "assets/monaco-editor/min-maps" }
             ],
             "styles": [
-              "node_modules/nouislider/dist/nouislider.min.css",
+              "nouislider/dist/nouislider.min.css",
               {
                 "input": "src/styles.scss",
                 "bundleName": "stylesLight",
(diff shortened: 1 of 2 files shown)


ert78gb:
Hi @NickCao 

I tried to test this PR, but after the following commands 
- `npm run clean`
- `npm ci`
- `npm run build`
- `npm run electron`
I got the `assets/monaco-editor/min/vs/loader.js:1          Failed to load resource: net::ERR_FILE_NOT_FOUND` error.

If I run `npm run server:web` I got the same error and the in both cases the monaco editor did not load.
Could you help me how did you solved this issue?

ert78gb:
I am being very busy on this week so I don't know how I will time to review your PRs

NickCao:
I'm building from a clean git repo (running git clean -fdx) before building, and failed to reproduce the issue you saw.

ert78gb:
We use `@angular-builders/custom-webpack` and built in angular build package to create a build of the web and electron renderer. I started to use `@angular-builders/custom-webpack` to create es5 bundles for electron because the earlier version of electron did not support esm modules in the renderer process. AFAIK the newer or current version of the electron supports esm, but I am not sure.

When you modify the configuration of the `uhk-web` package you have to think on the solution will work with both build method or not. 

Currently we does not have automated test, but every time when I upgrade a dependency or modify the release process. I check 
- add/remove keymaps
- drag and drop of key actions and macro actions
- smart macro doc
- firmware upgrade
- mouse speed test
- modify/restore user config
- create a new release. You can create only linux release on your machine because the Mac and Windows certificates are encrypted

the main functionalities of Agent. 

Please test these features when you open a PR.

When we upgrade electron, electron-builder, electron-updater packages I do it on separated `electron-qa` repo to test the auto update feature is works or not.

I just wrote these thought to show how many things have to cover when we do a major changes

NickCao:
> It turns out, when two packages rely on the same module with a compatible version number in the dependencies, npm (with workspaces) will install the package only in root/node_modules. Otherwise, it will install in the node_modules of package(s). - https://stackoverflow.com/questions/69981061/npm-workspaces-installing-packages-into-a-workspace-node-modules

Thus the currently hardcoded path would break.

NickCao:
And the difficulty we are facing is actually a long standing issue of npm: https://github.com/npm/rfcs/issues/287, maybe we should just use pnpm/yarn instead.....

ert78gb:
@NickCao Thank you for your contribution. I am closing this PR because we keep using lerna until the npm workspace does not support our setup. I would not like to introduce other package manager because they have cons too.

PR #1976, opened 2023-02-27
chore(agent): cleanup nested package

NickCao:
WARN: I don't know why the nested package was required in the first place, if this breaks something, let me know.


ert78gb:
We can't bundle `node-hid` package with webpack because it causes issues on some operating systems. This is the reason why we use 2nd package.json, because in this case the `node-hid` is just bundled in the asar file. 

I haven't time to test this PR but I think it will break or releases

ert78gb:
I tested this PR and it breaks the release process.

NickCao:
Still figuring out how to make things work with npm workspace, paths to `node_modules` are hardcoded in numerous places.

ert78gb:
@NickCao Thank you for your contribution. I am closing this PR because we keep using lerna until the npm workspace does not support our setup. I would not like to introduce other package manager because they have cons too.

PR #1991, opened 2023-03-25
chore(usb): use npx in script shebang instead of hardcoding path

NickCao:
The scripts no longer depends on ts-node to be installed into the specific path.


ert78gb:
It will be too slow. We use these script on raspberry pi and npx try to install the lastest version of the script.
2nd reason is the running of the script will indeterministic, because the esm-loader is not finalised as far as I know.

NickCao:
Shouldn't npx prefer the locally installed version? It's description says: `This command allows you to run an arbitrary command from an npm package (either one installed locally, or fetched remotely)`. And may you clarify a bit about the nondeterminism?

ert78gb:
`npx` prefers the locally installed version, but if someone forgets to run `npm ci` then` ts-node` is missing and `npx` asks to install it. If the user installs different version that is tested then maybe the script might fail, because the newer version of the `ts-node` uses a different version of node loader hook, so these scripts do not support it and we will face `weird` error messages. 
If the user forgets to install dependencies then the script returns with a very clear error message.

The other weird situation is if dependencies haven't been installed, but `ts-node` is available in the node module resolution path and `npx` uses it, we will be in the same situ as described above.

This is what I mean "nondeterminism".

Rethink my concerns, these are very rare cases and maybe we don't have to care about it.
I like the benefit of this PR, - we could call these scripts from anywhere, not just from the `packages/usb` folder- so we could give a try for it.

@mondalaci Could you test it on the raspberry pi? Do you recognize any significant performance changes?

mondalaci:
I don't want to spend time on rebuilding Agent on the Pi, but on my Ryzen 5 3600X PC, script execution time increased from about 0.6s to 1s which would likely translate to a significantly higher time difference on the Pi, making the operation of our factory team slower, so this issue should be addressed.

NickCao:
Trying another implementation with custom ESM loaders. While this implementation losses the benefit of being able to call the scripts from anywhere, it should address the performance concerns.

ert78gb:
On my mac this version faster. The `./get-variable.ts debounceTimePress` 11-15% faster
On linux the improvement is just 1-3%

If I switch branch the first run always slower. Does not matter switching from master to PR or back.

ert78gb (approved):

mondalaci:
@NickCao Thank you for your contribution! In the future, please open an issue before sending us a PR to discuss the details. @ert78gb works on Agent part-time, so his time is limited, and reviewing this PR has consumed more than an hour of this time which could have been better spent on other issues. I don't want to discourage you, but we feel that such small optimizations don't worth the time spent.

NickCao:
I'm more than willing to take the suggestion. Thanks again for your great patience in guiding me through the process!

PR #2079, opened 2023-09-02
feat: MD5 checksum-based firmware updates

### LINKED ISSUE #1663: MD5 checksum-based firmware updates
This is the Agent-side specification of checksum-based firmware updates. See the [firmware-side specification](https://github.com/UltimateHackingKeyboard/firmware/issues/440).

Agent will be able to query two kinds of module checksums:
- The actual checksums from the modules themselves
- The bundled checksums. For any given firmware version these checksums are the same as the actual checksums, but stored in the master device (right keyboard half).

When displaying the version numbers on the firmware page, display the master version number as it is.

For the slave version numbers:
- If the slave version number equals the master version number, display it as it is.
- If the slave version number doesn't equal the master version number:
    - If the actual checksum and the bundled checksum don't match, display the slave firmware version.
    - If the actual checksum and the bundled checksum match, display the master version, and underline the version number with gray dots and show the following tooltip text when hovering over it: "This module runs firmware ${actualVersionNumber}, but based on its checksum, it's the same as firmware ${bundledModuleVersion}, hence it hasn't been updated. Unnecessary firmware updates might brick your UHK and modules in very rare circumstances, and make the firmware update process longer, so Agent only updates firmwares when justified."

ert78gb:
closes #1663


AkechiShiro:
@mondalaci @ert78gb Thanks a lot for this feature, 

Is this active by default for firmware updates ? Or is this "experimental" on the latest firmware release ?

mondalaci:
The next Agent version will contain this feature.

PR #2081, opened 2023-09-09
feat: send-command.ts

### LINKED ISSUE #2080: Generic USB command for testing.
I am repeatedly running into pains regarding testing of firmware USB-related PRs...

Could I have a new script in `packages/usb/test-command.ts` that would take a list of (8bit) integers, which it would simply take and just send to the keyboard, and then print a list of results on two lines - once as a list of numbers, seconds time as a string (only valid characters so that a strings in reply are not terminated prematurely due to zero bytes)?

E.g.:
```
$ ./test-command.ts 0 8 2
foo
0 102 111 111
```

Or is there another way already established for this kind of testing?

ert78gb:
closes #2080


kareltucek:
Thanks Róbert!

ert78gb:
Anytime, just let me know if you need something like that

PR #2082, opened 2023-09-10
feat: error panel

### LINKED ISSUE #2076: Error panel
[error state](https://github.com/UltimateHackingKeyboard/firmware/blob/66a2bef451b32d8caf4ca7ae5d3a155cf9bdcc04/right/src/usb_commands/usb_command_get_device_state.c#L38) has been added recently. If a smart macro error or warning is detected, it's nonzero, in which case, make a bottom error panel visible in Agent. The error panel must be resizable, just like the smart macro sidebar. 

Use [UsbCommand_GetVariable(UsbVariable_StatusBuffer=5)](https://github.com/UltimateHackingKeyboard/firmware/blob/master/right/src/usb_commands/usb_command_get_variable.c#L28) to fetch the status buffer.

The `UsbCommand_GetKeyboardState()` payload will be refactored, and its length will be decreased. To prepare for this change, only watch the last bit of `SetUsbTxBufferUint8(1, IsEepromBusy)` and `SetUsbTxBufferUint8(2, MERGE_SENSOR_IS_MERGED)`, not the whole bytes.

ert78gb:
closes #2076

### DIFF
--- a/packages/uhk-agent/src/services/device.service.ts
+++ b/packages/uhk-agent/src/services/device.service.ts
@@ -51,6 +51,7 @@ import {
     TmpFirmware,
     UhkHidDevice,
     UhkOperations,
+    UsbVariables,
     usbDeviceJsonFormatter,
     waitForDevice
 } from 'uhk-usb';
@@ -63,8 +64,8 @@ import {
     copySmartMacroDocToWebserver,
     getBackupUserConfigurationContent,
     getDefaultFirmwarePath,
-    getUserConfigFromHistoryAsync,
     getSmartMacroDocRootPath,
+    getUserConfigFromHistoryAsync,
     loadUserConfigHistoryAsync,
     makeFolderWriteableToUserOnLinux,
     saveTmpFirmware,
@@ -80,6 +81,7 @@ import {
 export class DeviceService {
     private _pollerAllowed: boolean;
     private _uhkDevicePolling: boolean;
+    private _checkStatusBuffer: boolean;
     private queueManager = new QueueManager();
     private wasCalledSaveUserConfiguration = false;
     private isI2cDebuggingEnabled = false;
@@ -372,6 +374,7 @@ export class DeviceService {
                     this.logService.config('[DeviceService] User configuration', data.userConfig);
                     const buffer = mapObjectToUserConfigBinaryBuffer(data.userConfig);
                     await this.operations.saveUserConfiguration(buffer);
+                    this._checkStatusBuffer = true;
                     response.userConfigSaved = true;
                 }
             } else {
@@ -498,6 +501,7 @@ export class DeviceService {
                 userConfig);
             const buffer = mapObjectToUserConfigBinaryBuffer(userConfig);
             await this.operations.saveUserConfiguration(buffer);
+            this._checkStatusBuffer = true;
 
             response.modules = await this.getHardwareModules(false);
             await copySmartMacroDocToWebserver(firmwarePathData, this.logService);
@@ -632,6 +636,19 @@ export class DeviceService {
         event.sender.send(IpcEvents.device.changeKeyboardLayoutReply, response);
     }
 
+    private async checkStatusBuffer(): Promise<void> {
+        if (!this._checkStatusBuffer) {
+            return;
+        }
+
+        this._checkStatusBuffer = false;
+
+        const message = await this.operations.getVariable(UsbVariables.statusBuffer);
+        if (message) {
+            this.win.webContents.send(IpcEvents.device.statusBufferChanged, message);
+        }
+    }
+
     /**
      * HID API not support device attached and detached event.
      * This method check the keyboard is attached to the computer or not.
@@ -655,6 +672,7 @@ export class DeviceService {
 
                         if (state.hasPermission && state.zeroInterfaceAvailable) {
                             state.hardwareModules = await this.getHardwareModules(false);
+                            this._checkStatusBuffer = true;
                         } else {
                             state.hardwareModules = {
                                 moduleInfos: [],
@@ -671,6 +689,7 @@ export class DeviceService {
                     }
 
                     await this.pollDebugInfo(iterationCount);
+                    await this.checkStatusBuffer();
                 } catch (err) {
                     this.logService.error('[DeviceService] Device connection state query error', err);
                 }
@@ -712,6 +731,7 @@ export class DeviceService {
             this.logService.config('[DeviceService] User configuration will be saved', data.configuration);
             const buffer = mapObjectToUserConfigBinaryBuffer(data.configuration);
             await this.operations.saveUserConfiguration(buffer);
+            this._checkStatusBuffer = true;
 
             if (data.saveInHistory) {
                 await saveUserConfigHistoryAsync(buffer);
--- a/packages/uhk-common/src/models/application-settings.ts
+++ b/packages/uhk-common/src/models/application-settings.ts
@@ -13,6 +13,10 @@ export type AppThemeSelect = {
 
 export interface ApplicationSettings {
     checkForUpdateOnStartUp: boolean;
+    /**
+     * Application main error panel height in percent.
+     */
+    errorPanelHeight?: number;
     everAttemptedSavingToKeyboard: boolean;
     animationEnabled?: boolean;
     appTheme?: AppTheme;
--- a/packages/uhk-common/src/util/ipcEvents.ts
+++ b/packages/uhk-common/src/util/ipcEvents.ts
@@ -48,6 +48,7 @@ export class Device {
     public static readonly loadUserConfigHistoryReply = 'device-load-user-config-history-reply';
     public static readonly getUserConfigFromHistory = 'device-get-user-config-from-history';
     public static readonly getUserConfigFromHistoryReply = 'device-get-user-config-from-history-reply';
+    public static readonly statusBufferChanged = 'device-status-buffer-changed';
     public static readonly toggleI2cDebugging = 'device-toggle-i2c-debugging';
     public static readonly i2cWatchdogCounterChanged = 'device-i2c-watchdog-counter-changed';
 }
--- a/packages/uhk-usb/src/constants.ts
+++ b/packages/uhk-usb/src/constants.ts
@@ -77,7 +77,8 @@ export enum UsbVariables {
     testUsbStack = 1,
     debounceTimePress = 2,
     debounceTimeRelease = 3,
-    usbReportSemaphore= 4
+    usbReportSemaphore= 4,
+    statusBuffer= 5,
 }
 
 export const LAYER_NUMBER_TO_STRING = [
--- a/packages/uhk-usb/src/uhk-operations.ts
+++ b/packages/uhk-usb/src/uhk-operations.ts
@@ -18,9 +18,9 @@ import {
     UhkModule
 } from 'uhk-common';
 import { promisify } from 'util';
-import { DevicePropertyIds } from './constants.js';
 import {
     ConfigBufferId,
+    DevicePropertyIds,
     EepromOperation,
     EnumerationModes,
     KbootCommands,
@@ -604,10 +604,19 @@ export class UhkOperations {
         return convertSlaveI2cErrorBuffer(responseBuffer, slaveId);
     }
 
-    public async getVariable(variableId: UsbVariables): Promise<number> {
+    public async getVariable(variableId: UsbVariables, iteration: number = 0): Promise<number | string> {
         this.logService.usb('[DeviceOperation] USB[T]: get variable');
         const buffer = Buffer.from([UsbCommand.GetVariable, variableId]);
         const responseBuffer = await this.device.write(buffer);
+
+        if (variableId === UsbVariables.statusBuffer) {
+            let message = readUhkResponseAs0EndString(UhkBuffer.fromArray(convertBufferToIntArray(responseBuffer)));
+            if (message.length > 0 && iteration < 20) {
+                message += await this.getVariable(variableId, iteration + 1);
+            }
+
+            return message;
+        }
 
         return responseBuffer[1];
     }
--- a/packages/uhk-web/src/app/app.component.html
+++ b/packages/uhk-web/src/app/app.component.html
@@ -23,9 +23,17 @@
     <div class="inner-content"
          [class.top-notification-panel-visible]="isTopNotificationPanelVisible()"
          [class.second-menu-visible]="secondSideMenuVisible">
-        <router-outlet></router-outlet>
+        <as-split direction="vertical"
+                  [useTransition]="true"
+                  (dragEnd)="errorPanelSizeDragEndHandler($event)">
+            <as-split-area [size]="splitSizes.top">
+                <router-outlet></router-outlet>
+            </as-split-area>
+            <as-split-area *ngIf="statusBuffer" [size]="splitSizes.bottom">
+                <span [innerHTML]="statusBuffer | newLineToBr | safeHtml"></span>
+            </as-split-area>
+        </as-split>
     </div>
-
 </div>
 <notifier-container></notifier-container>
 <highlight-arrow *ngIf="firstAttemptOfSaveToKeyboard$ | async" [@highlightArrow]></highlight-arrow>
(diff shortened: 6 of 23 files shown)


mondalaci:
@kareltucek This feature is working, we're fine-tuning it, and just realized that when intentionally making an error, such as `blah`, it gets put twice into the status buffer:

```
Error at $onInit 1/7: Unrecognized command: blah
> 7 | blah
>   | ^
Error at $onInit 1/7: Unrecognized command: blah
> 7 | blah
>   | ^
```

The `set a = 1 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890` line puts multiple `Parameter not recognized: a ...` error messages into the status buffer.

Can you reproduce these issues? Is this a recent bug?

Also, the status buffer shouldn't be emptied upon reading it because when Agent reads it, and Agent is restarted, the error doesn't reappear like it's resolved, even though it's not.

kareltucek:
> This feature is working, we're fine-tuning it, and just realized that when intentionally making an error, such as `blah`, it gets put twice into the status buffer:
> 
> ```
> Error at $onInit 1/7: Unrecognized command: blah
> > 7 | blah
> >   | ^
> Error at $onInit 1/7: Unrecognized command: blah
> > 7 | blah
> >   | ^
> ```
> 
> The `set a = 1 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890` line puts multiple `Parameter not recognized: a ...` error messages into the status buffer.
> 
> Can you reproduce these issues? Is this a recent bug?

This seems to be Agent bug.

> Also, the status buffer shouldn't be emptied upon reading it because when Agent reads it, and Agent is restarted, the error doesn't reappear like it's resolved, even though it's not.

The status buffer is not emptied, but the firmware does not know that Agent wants to start reading it from the beginning. 

How should I detect that USB status buffer position should be resetted?

mondalaci:
If I put `blah` into `$onInit`, I get a single error message on UHK power-up. Then, if I save my configuration (even if I do it multiple times), I get two such error messages in the status buffer. How so?

Please reset the USB status buffer position when it's fully read. We also need a dirty flag so that Agent knows when the status buffer content changes. Let's store the dirty flag bit in place of `UsbCommand_GetKeyboardState()` : `getErrorState()` value.

kareltucek:
> If I put blah into $onInit, I get a single error message on UHK power-up. Then, if I save my configuration (even if I do it multiple times), I get two such error messages in the status buffer. How so?

Because once the error is produced by macro validation, second time when $onInit is executed.

kareltucek:
> Please reset the USB status buffer position when it's fully read. 

How does the Agent then know when to stop reading? If I just reset the position, it will continue reading the buffer from the beginning...

mondalaci:
Shouldn't both macro validation and $onInit execute on bootup, resulting in two duplicate errors without saving the configuration? If not, why not?

Agent can stop reading when encountering a zero byte. Then, reread the status buffer when the dirty flag is set.

kareltucek:
> Shouldn't both macro validation and $onInit execute on bootup, resulting in two duplicate errors without saving the configuration? If not, why not?

Validation is not executed on bootup. 

Why: As far as I remember, you were concerned about unnecessary power usage due to validation, and I didn't see much reason to validate on bootup anyways - since if the user decides to ignore errors/warnings upon saving configuration, he clearly doesn't want to deal with them, especially if they are from unused macros or something like that. As a result, it seemed to make sense to validate only when saving configuration.

mondalaci:
The validation is good the way it is; I didn't remember.

We should somehow prevent duplicate errors in the status buffer because they appear duplicated in Agent's newly implemented error panel, confusing users. Feel free to suggest how.

kareltucek:
Well, multiple ideas come to mind:

- Handle it in Agent. We will probably want Agent to be able to jump to precise macro and position when an error is clicked, which means that Róbert will have to parse/split the feed into separate error objects anyway, so when at it, having a hash table of already-encountered errors is not much additional work.

  (Additionally, Agent may show a little number somewhere to indicate how many times an error was encountered.)

- I can go through all errors and determine which ones throw during validation, and make them throw only during validation. (At the moment, deprecation warnings are done that way.)

mondalaci:
I prefer a firmware-based solution. That way, `printStatus` wouldn't be affected by the duplicate issue.

kareltucek:
The trouble is that it means that on the contrary `printStatus` will be missing errors from runtime execution, which may be confusing if the user uses `printStatus` to debug a misbehaving macro.

kareltucek:
Third solution... switch to reporting via macros, and make every macro declare a local static variable that will make sure that such error is printed only once.

Trouble: additional memory consumption around 100 bytes, assuming that there is no additional padding. (I suspect that there may be some requirement on 4-byte alignment, which would bump the usage to 400.)

kareltucek:
Fourth solution: Improvement of second... have a system of flags that keeps track of whether the status buffer contains validation errors or not, and report only if validation is taking place, or the status buffer does not contain validation errors. 

Trouble: The logic is complex and may confuse the user, since it will throw duplicate errors when there are no validation errors in the error buffer, but not if the buffer contains validation errors.

mondalaci:
How about skipping validation for the macro events that'd run otherwise anyway, such as `$onInit`, the current keymap's event, etc?

kareltucek:
Not a clean solution either:

- Means a lot of code clutter to decide which macros to validate and which not to. 
- Macro validation goes through the entire macro, irrespectively of control flow. Execution will skip those statements that are not executed due to conditions, goTo statements, etc.
- If we introduce some sort of code compilation/tokenization, this solution breaks down immediately.

Besides, I think that it is correct for the firmware to produce duplicate error reports in these situations, so I would prefer Agent to handle this.

kareltucek:
Of course, sixth idea (so far my favourite): to naively search the status buffer to see if there already is an error from the same macro/position.

It is a costly operation, but it should not take place in regular workflow, so it probably does not matter...

mondalaci:
Agent will filter out duplicates, as suggested.

Why does the `set a = 1 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890` command trigger 6 errors (the last one being truncated) in $onInit?

kareltucek:
> Why does the set a = 1 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 command trigger 6 errors (the last one being truncated) in $onInit?

I can't reproduce this with `printStatus`, therefore this seems like an Agent bug to me.

kareltucek:
Well, actually I can reproduce it via USB command, so it probably indeed is firmware issue. Will look into it.

ert78gb:
Try to use the ` ./get-variable.ts statusBuffer` usb script. The `statusBuffer` argument introduced in this PR.

The exact macro that cause infinity loop is
```
set a = 1 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890
asd
```

if I remove the last `1234567890` then the infinity loop disappearing. I think the line length the key factor.

ert78gb:
FYI. To prevent infinity loop the `./get-variable.ts statusBuffer` terminate after 20 iteration.

kareltucek:
Fixed on https://github.com/UltimateHackingKeyboard/firmware/pull/683 

(Both the bug, and resetting after first zero byte.)

mondalaci:
Please also store the dirty flag bit in place of `UsbCommand_GetKeyboardState()` : `getErrorState()` value.

kareltucek:
Ah, sorry, forgotten about that!

kareltucek:
Done in https://github.com/UltimateHackingKeyboard/firmware/pull/685

mondalaci:
@ert78gb Remaining issues to fix:
- Deduplicate error messages
- Animate error panel
- Don't display excess errors; The `set a = 1 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890 1234567890` command only puts duplicate errors into the status buffer, but Agent displays 8 errors.

mondalaci:
@kareltucek My `$onInit` contains:
`set module.trackpoint.axisLockFirstTickSkew 1`

When adding `blah` in a new line, I get:

```
Error at $onInit 1/2: Unrecognized command: blah
> 2 | blah
> | ^
Warning at $onInit 1/2: Unprocessed input encountered.
> 2 | blah
> | ^
```

Why the second warning? It seems redundant.

Then I append `x` to `blah`, and I get:

```
Warning at $onInit 1/1: Unprocessed input encountered.
> 1 | set module.trackpoint.axisLockFirstTickSkew 1
> | ^
Error at $onInit 1/2: Unrecognized command: blahx
> 2 | blahx
> | ^
Warning at $onInit 1/2: Unprocessed input encountered.
> 2 | blahx
> | ^
```

The newly appearing first warning doesn't seem legit.

mondalaci:
@kareltucek https://github.com/UltimateHackingKeyboard/firmware/pull/695 seems to have fully resolved the mentioned issue, and this feature is working remarkably well on my end. How about you?

kareltucek:
Yep, should be working fine.

(Sorry about that 2 hours ago comment, didn't notice it. Noticed the problem when testing rpnfan's report...)

>  this feature is working remarkably well on my end. How about you?

If you mean Agent side, there is still some space for improvement:

- Mainly, multiple spaces should not be reduced to a single space. (I.e., replace them with nbsp or whatever...)
- If error is encountered outside of validation, or if validation is triggered manually, Agent does not download the errors.
- It would be nice if clicking an error brought up the relevant macro / action and placed text cursor at the position.
- It may be nice to have some visual separation of error records. 

-------

If you mean firmware side....

Then how should *I* know? 🙃 I had fixed every problem I could find, but I am quite sure that some problems managed to slip past...

mondalaci:
> If error is encountered outside of validation, or if validation is triggered manually, Agent does not download the errors.

Please provide an example of how to reproduce this.

kareltucek:
> > If error is encountered outside of validation, or if validation is triggered manually, Agent does not download the errors.
> 
> Please provide an example of how to reproduce this.

- Have a `blah` macro.
- Open Agent.
- Use the `blah` macro.
- Observe that UHK flashes ERR, but error panel does not pop up in Agent.

mondalaci:
@kareltucek Thanks!

> - Mainly, multiple spaces should not be reduced to a single space. (I.e., replace them with nbsp or whatever...)
> - If error is encountered outside of validation, or if validation is triggered manually, Agent does not download the errors.

@ert78gb These errors must be fixed in this PR. The dirty flag of the `GetKeyboardState` USB call must be watched for the second issue.

> - It would be nice if clicking an error brought up the relevant macro / action and placed text cursor at the position.
> - It may be nice to have some visual separation of error records.

@ert78gb These suggestions can be implemented in another PR or in this one. What's your choice?

kareltucek:
> - It would be nice if clicking an error brought up the relevant macro / action and placed text cursor at the position.

Errors now don't provide numeric position on line, should I add it? (Otherwise, Róbert will have to parse that from number of spaces on the third line...)

(I.e., `Error at $onInit 1/2/1: Unrecognized command: blah` rather than `Error at $onInit 1/2: Unrecognized command: blah`.)

mondalaci:
Yes, that'd be useful.

ert78gb:
I would like to implement the following features in other PR.

> - It would be nice if clicking an error brought up the relevant macro / action and placed text cursor at the position.
> - It may be nice to have some visual separation of error records.

My questions:
- Are the errors always macro error?
- Is the error parsing logic correct? The `^(Error at |Warning at )(.*) (\d+)\/(\d+)\/(\d+):` regular expression returns with every information of the navigation.
  - 2nd group is the name of the macro
  - 3rd group is the macro index item
  - 4th error line in the macro item
  - 5th error position in the macro item

kareltucek:
> * Are the errors always macro error?

At the moment yes, but I can't rule out that we will report other problems via this same error mechanism in future.

> * Is the error parsing logic correct?

Yes!

---

Also, please assume that in some cases the following lines (those that report position visually) may be missing (Either the second one in case line is known but position on line is unknown, or both in case it is a non-macro error (which is not the case at the moment but may be the case in future).)

ert78gb:
could we add the "macro" error logic somehow to the error message? This info differentiate the errors in the future. Or is it overkill now?

ert78gb:
For example `Macro error at ....`

kareltucek:
I don't have an opinion on whether we should or not. The question itself sounds as a good one, but not very important.

@mondalaci should we?

ert78gb:
@kareltucek 
Is the following the expected behaviour:
- Start the agent
- save a macro with error
- the error panel is visible. It is the expected
- restart agent. The error panel is not visible. The status buffer reading returns with an empty string
- exit from Agent. 
- run the `./get-variable.ts statusBuffer` usb script it also returns with nothing.

I used the `#8e7b06f9` firmware.

The earlier `#573154fb` firmware kept the error buffer.

kareltucek:
Well, the dirty flag is resetted when the buffer is fully consumed, so the fact that dirty flag is zero after agent restart is correct. But reading from the status buffer variable should give you the error.

I will look into it in the evening, I am not at computer atm. Sorry for complications!

ert78gb:
no worries

kareltucek:
Well, if I use `./send-command.ts 18 5` instead of `./get-variable.ts statusBuffer`, I am getting:

```
$ ./send-command.ts 18 5
Byte result:
0 69 114 114 111 114 32 97 116 32 116 115 116 69 114 114 32 49 47 49 47 49 58 32 85 110 114 101 99 111 103 110 105 122 101 100 32 99 111 109 109 97 110 100 58 32 98 108 97 104 10 62 32 49 32 124 32 98 108 97 104 10 62 32

String result:
Error at tstErr 1/1/1: Unrecognized command: blah
> 1 | blah
>
$ ./send-command.ts 18 5
Byte result:
0 32 32 124 32 94 10 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0

String result:
  | ^

$ ./send-command.ts 18 5
Byte result:
0 69 114 114 111 114 32 97 116 32 116 115 116 69 114 114 32 49 47 49 47 49 58 32 85 110 114 101 99 111 103 110 105 122 101 100 32 99 111 109 109 97 110 100 58 32 98 108 97 104 10 62 32 49 32 124 32 98 108 97 104 10 62 32

String result:
Error at tstErr 1/1/1: Unrecognized command: blah
> 1 | blah
>
$ ./send-command.ts 18 5
Byte result:
0 32 32 124 32 94 10 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0

String result:
  | ^

 ./send-command.ts 18 5
Byte result:
0 69 114 114 111 114 32 97 116 32 116 115 116 69 114 114 32 49 47 49 47 49 58 32 85 110 114 101 99 111 103 110 105 122 101 100 32 99 111 109 109 97 110 100 58 32 98 108 97 104 10 62 32 49 32 124 32 98 108 97 104 10 62 32

String result:
Error at tstErr 1/1/1: Unrecognized command: blah
> 1 | blah
>
$ ./send-command.ts 18 5
Byte result:
0 32 32 124 32 94 10 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0

String result:
  | ^

...
```

Which seems fine to me. (There is just one error in the buffer and if I keep requesting it repeatedly, the firmware feeds it to me repeatedly.)

(Is there anything wrong with the above output?)

ert78gb:
sry. the bug was on my side. thx for the help

kareltucek:
You are welcome!

likern:
@mondalaci Could we release new version of agent and firmware? I'm not sure, but probably firmware fixes some annoying bugs (with advanced secondary roles, but I'm not sure). I'd like to test new functionality 😃

kareltucek:
Patience please, I would like to finish one more thing in firmware before the grand release...

PR #2251, opened 2024-04-27
feat: "Remap on all layers" shouldn't remap modifier layers

### LINKED ISSUE #2223: "Remap on all layers" shouldn't remap modifier layers
When using the "Remap on all layers" checkbox of the key action popover, Agent should remap all layers except modifier layers. If the user is on a modifier layer, the current modifier layer should be affected along with every non-modifier layer.

ert78gb:
closes #2223

### DIFF
--- a/packages/uhk-common/src/config-serializer/config-items/layer-name.ts
+++ b/packages/uhk-common/src/config-serializer/config-items/layer-name.ts
@@ -12,3 +12,10 @@ export enum LayerName {
     super,
     base = 255
 }
+
+export const MODIFIER_LAYER_NAMES = [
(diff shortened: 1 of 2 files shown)


mondalaci:
Not working as expected.

1. On my "QWERTY for PC" layer, I have the base, mod, mouse, and fn layers.
2. Add the fn2 and shift layers
3. Go to the base layer
4. Change the left bottom (LSuper) key to scancode "A" with "Remap on all layers" checked
5. The key is remapped on the base, mouse fn, fn2, and shift layers but not on the mod layer. It should be remapped on every layer but shift.

PR #2255, opened 2024-04-29
feat: jump to macro errors on click

### LINKED ISSUE #2092: Jump to macro errors on click
Copied from https://github.com/UltimateHackingKeyboard/agent/pull/2082#issuecomment-1722989032

- It would be nice if clicking an error brought up the relevant macro / action and placed text cursor at the position.
- It may be nice to have some visual separation of error records.

ert78gb:
closes #2092


mondalaci:
Works great!

@kareltucek Would you test this? You're much more familiar with the possible set of error messages, so you may catch something I didn't.

kareltucek:
Yep, tested, cool!

(They use a single abstraction and reporting format, so functionality should be without issues.)

PR #2270, opened 2024-05-10
feat: macro command syntax highlight and auto-completion

ert78gb:


ert78gb:
It is almost done. 
thx for the prev PR @kareltucek.
I think would be better to move the typescript of the `naive-autocompletion-parser` to the dev dependencies because typescript not really need as runtime/package dependency. 

@mondalaci @kareltucek Please play with it.

mondalaci:
The syntax highlighting and autocompletion look very cool and helpful. That's all I can say, but I'm not a heavy smart macro user.

Feel free to move naive-autocompletion-parser to devDependencies.

kareltucek:
> thx for the prev PR @kareltucek.

Thanks for finishing and refactoring the PR!

> I think would be better to move the typescript of the naive-autocompletion-parser to the dev dependencies because typescript not really need as runtime/package dependency.

Feel free to!

--------------

As for my testing, it works like a charm!

I am satisfied with it and think that this can be merged.

For the sake of completeness, as a heavy macro user, I will note known limitations:
- macro names and keymap ids are not completed with actual existing objects. This can be implemented relatively easily if we decide we want it.
- syntax highlight does weird things for key abbreviations and some identifiers. E.g., it highlights numbers as numbers even when they are used as keyid abbreviations. I may be able to figure out how to improve the matters slightly (e.g., so that keymap abbreviations like 'QT3' are highlighted consistently), but full solution is not possible due to technical reasons.
- modifier mask autocomplete does not complete modmask chains. This is limitation on the way the parser handles white characters. It may be possible to solve, but is complicated and probably not worth the time.
- the engine does not complete after certain characters - again a problem of the whitespace handling.

mondalaci:
@ert78gb I'll merge this as soon as you commit the devDependencies change.

@kareltucek I'd like to prioritize important firmware-related issues now, but feel free to implement improvements later, especially the easy ones.

PR #2276, opened 2024-05-27
feat: expose 20 mouse buttons

### LINKED ISSUE #2167: Expose 20 mouse buttons
Now that [the relevant firmware feature](https://github.com/UltimateHackingKeyboard/firmware/pull/751) has been implemented.

ert78gb:
closes #2167

Also removed the `extraMouseButtonsSupported` feature flag because Agent force to upgrade the firmware if it is not compatible.

### DIFF
--- a/packages/uhk-common/src/config-serializer/config-items/key-action/mouse-action.ts
+++ b/packages/uhk-common/src/config-serializer/config-items/key-action/mouse-action.ts
@@ -21,7 +21,19 @@ export enum MouseActionParam {
     button5,
     button6,
     button7,
-    button8
+    button8,
+    button9,
+    button10,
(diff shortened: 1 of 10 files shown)


mondalaci:
@kareltucek Mouse button 20 doesn't register for me when testing it with `xev`, but the rest does. Assuming you can reproduce the issue, please look into it. I'm wondering whether the scancode gets sent, and if so, it might be limited by USB descriptor intervals.

mondalaci:
@kareltucek Will you take a look at this? Don't wanna rush you; Just want to make sure this pending PR isn't forgotten.

mondalaci (approved):

PR #2292, opened 2024-05-31
fix: update patch level dependencies

EinfachAleks:
Update patch level devDependencies


ert78gb:
thank you for your contribution, but we don't do bulk dependency upgrade.

ert78gb:
our policy https://github.com/UltimateHackingKeyboard/agent/blob/master/NPM_UPDATES.md

PR #2303, opened 2024-06-25
feat: extend configuration history records with device type and name

### LINKED ISSUE #1937: Extend configuration history records with device type and name
@ert78gb We should also consider the device id and the compatibility of saved configurations with the current firmware. Let's talk about this before implementation.

ert78gb:
closes #1937


mondalaci:
Please use cursor: pointer style when hovering over the tabs.

ert78gb:
pointer cursor fixed

Issue #2336, opened 2024-08-21
Rhythmic freezing of Agent 


Tanz0rz:
Agent is constantly locking up in a perfect rhythm.

OS: Windows 10
Version 4.2.0

Behavior:
- Unfrozen for a half second
- Frozen for one second
- This repeats permanently  

No other part of my system is effected while the software is open. Reboots of the application and computer do not help. This happens with my UHK 2.0 connected and disconnected.

mondalaci:
Please try to reproduce this issue with another computer. This is hardly an Agent issue because no one else reported it, and the latest Agent version has been released for a while.

ert78gb:
Based on the symptom there is a slow USB devices connected to the computer. Other case in #1482.
Could check a slow device cause the freeze?

Tanz0rz:
> Please try to reproduce this issue with another computer. This is hardly an Agent issue because no one else reported it, and the latest Agent version has been released for a while.

The issue is happening exclusively with the Agent software on my Windows computer with or *without* the keyboard connected. Agent has been reinstalled, computer rebooted, etc. 

> Based on the symptom there is a slow USB devices connected to the computer. Other case in https://github.com/UltimateHackingKeyboard/agent/issues/1482.
Could check a slow device cause the freeze?

I think you are right. If I open Agent from the command line, you can see that the freezing is almost perfectly in line with the USB polling interval from Agent (video attached). 

https://github.com/user-attachments/assets/65d3181a-0185-4a09-92b5-5f5e3ea3e31a

ert78gb:
> Add a command line flag that triggers Agent to poll for USB devices exactly one time on startup and disables the persistent looping polling. That should be extremely simple and would work for people who have gone deep enough to find this issue on GitHub.

The polling not just simple poll to detect the keyboard is connected or not. The polling reads status information that affect the Agent UI.

Tanz0rz:
> > Add a command line flag that triggers Agent to poll for USB devices exactly one time on startup and disables the persistent looping polling. That should be extremely simple and would work for people who have gone deep enough to find this issue on GitHub.
> 
> The polling not just simple poll to detect the keyboard is connected or not. The polling reads status information that affect the Agent UI.

When Agent first starts, it can cache the USB device ID for the keyboard, right? It seems like the problem is due to Agent asking for information about every USB device on my computer every second after it has already found and connected to the keyboard.

ert78gb:
[...]

PR #2337, opened 2024-08-24
fix: use `devicesAsync` instead of `devices` of `node-hid`

ert78gb:
maybe fixes: #2336


mondalaci:
Works for me, and according to https://github.com/UltimateHackingKeyboard/agent/issues/2336#issuecomment-2308677924 it fixed the issue so feel free to merge it.

PR #2484, opened 2025-01-08
chore: make usb-interface command line argument optional

### LINKED ISSUE #2467: Usb scritps - device selectors; Don't require usb interface
Currently, when using vid/pid specification, usb-interface is required (e.g., ` --vid=0x37a8 --pid=9 --usb-interface=0`). 

When there is only one such device connected, this should not be necessary and Agent should automatically pick the correct one.

If I understand correctly, usb-interfaces are no longer fixed, so in various auxiliary scripts this can't be easily hardcoded.

ert78gb:
closes #2467

### DIFF
--- a/packages/uhk-agent/src/util/command-line.ts
+++ b/packages/uhk-agent/src/util/command-line.ts
@@ -66,7 +66,7 @@ const sections: commandLineUsage.Section[] = [
             },
             {
                 name: 'pid',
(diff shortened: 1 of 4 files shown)


ert78gb:
@kareltucek Perhaps you have overlooked this PR :)

kareltucek:
Oops!

kareltucek:
Works like a charm! I have just pushed a commit that removes the argument in firmware scripts...

Pleas merge this PR - I don't have permissions 🙈.

PR #2485, opened 2025-01-08
fix: erase user configuration

### LINKED ISSUE #2476: erase-user-config not working with uhk80
```
./erase-user-config.ts --serial-number=1399406988
Error: Communications error with UHK. Response code: 3
    at file:///opt/west/firmware/lib/agent/packages/uhk-usb/dist/src/uhk-hid-device.js:323:35
```

It triggers command 5, which is hardware config write, not user config:
```
        case UsbCommandId_WriteHardwareConfig:
            UsbCommand_WriteConfig(ConfigBufferId_HardwareConfig, GenericHidOutBuffer, GenericHidInBuffer);
```

and it results in the code 3, which is:

```
UsbStatusCode_WriteConfig_BufferOutOfBounds
```

ert78gb:
closes #2476


ert78gb:
@kareltucek please test it if you will have time.
If I start the Agent after I deletion then I got status buffer error. Is it the expected behaviour or the erase process has an error?

kareltucek:
Yes, works fine!

Issue #2424, opened 2024-09-27
Key map dialog appearing outside of window bounds


Zetaphor:
Running on Windows 11 until I figure out getting Agent working on NixOS

This is on a 4K display:

![image](https://github.com/user-attachments/assets/9c02c875-eb82-4a7f-a503-3f8d4c5cc0fe)

ert78gb:
@mondalaci I can't reproduce it on my windows 10 notebook. It does not have 4K display.

@Zetaphor Do you still experience the wrong positioning of the dialog with the latest Agent version?

mondalaci:
@Zetaphor Nobody else has reported this issue, and I can't reproduce it on Linux Mint, so I'm suspicious of a non-standard configuration / settings. The [high-]DPI settings used or any related settings would be welcome.

pcooke9:
I use 4k resolution with Win11 display scaling set to 150%.

I've seen similar issues manifesting in various ways for a long time. I can't currently reproduce it exactly the same as the image in the first post, but there are some glitches.

While I was trying to reproduce it just now (with recent Agent master branch), I discovered a funny little quirk. The outer edges of the keys teleport to the opposite side of the keymap when resizing the window.

This video was taken with Win 11 at 4k res and 150% display scaling, but I can reproduce it with 100% scaling as well:
https://www.youtube.com/watch?v=JuQUrXqIrQY

mondalaci:
Thanks for sharing! The teleporting issue only seems to happen when the application window is very small. I don't think users use such a small window size, so we should only address this if the fix is very easy.

pcooke9:
Of course, I can't imagine anyone actually trying to use Agent with the window that small.😆 I just thought it was interesting, and posted it in case it could potentially help with something down the road.

ert78gb:
The main issue solved in the #2610 so I close it

PR #2610, opened 2025-04-06
fix: visibility of the corner of the next/prev layer when screen is very small

ert78gb:
It is a quick fix of the https://github.com/UltimateHackingKeyboard/agent/issues/2424#issuecomment-2764764330 .

The "phantom" key is the key from the prev/next layer. This PR is the fastest solution that I could provide now. I don't know how much effort would we like to put in it. 

Screenshot 2025-04-06 at 16 49 36

### DIFF
--- a/packages/uhk-web/src/app/components/keyboard/slider/keyboard-slider.component.scss
+++ b/packages/uhk-web/src/app/components/keyboard/slider/keyboard-slider.component.scss
@@ -5,11 +5,11 @@
 
 @mixin left-position() {
     left: 0;
-    transform: translateX(-101%);
+    transform: translateX(-110%);
 }
 
 @mixin right-position() {
-    left: 101%;
+    left: 110%;
     transform: translateX(0);
 }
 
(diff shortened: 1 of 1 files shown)


mondalaci:
@pcooke9 Would you test https://github.com/UltimateHackingKeyboard/agent/actions/runs/14293667375 ?

I can reproduce this.

pcooke9:
> The "phantom" key is the key from the prev/next layer.

I didn't realize those keys were from a different layer, but you're right. It's a little weird how when you click one of them, it opens the remapping properties for the currently selected layer though...🤷‍♂️

Anyway, this version is definitely better. You have to make the window ridiculously small before they're visible now. You can still open the key remapping properties by clicking on the teleported keys, but you can no longer accidentally swap/exchange them via drag-and-drop.👍

Overall, I don't think it's much of a problem, TBH.

ert78gb:
Extra protection is if we set a minimum application window size, so the user can't set extra small / problematic window size. But I don't know what should be the minimum. The other layer getting visible around ~400px window width. The width of the side menu is 250px. 
The min windows size is just a workaround I think I have to sole it properly if it will the priority, but now we have more important things

PR #2622, opened 2025-05-01
feat: expose zephyr logs

### LINKED ISSUE #2618: Expose the Zephyr log
Image

Instead of the two checkboxes, put a "Debug I2cRecovery" toggle button, and when it's enabled, reveal the "Ring an audible bell upon I2cRecovery increments" checkbox rightward of it.

For the UHK 80, show an "Enable Zephyr logging" toggle button. When it's enabled, use the terminal to output the content of the Zephyr shell. The USB commands are already implemented in the firmware. Karel's short explanation:

> There is a variable that defines whether the backend is active or not. Then there is a flag in the device state that indicates that there is new logging data in the buffer. And finally, a command that consumes part of the buffer. (Really consumes - it frees that part of the buffer for more logs.)

Ask him if you need the details.

ert78gb:
closes #2618

### DIFF
--- a/packages/uhk-agent/package-lock.json
+++ b/packages/uhk-agent/package-lock.json
@@ -20,6 +20,7 @@
         "fs-extra": "11.3.0",
         "get-port": "7.1.0",
         "node-hid": "3.1.2",
+        "p-limit": "6.2.0",
         "serialport": "13.0.0",
         "sudo-prompt": "9.2.1",
         "tmp": "0.2.3",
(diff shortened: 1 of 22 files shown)


ert78gb:
@kareltucek How could I test zephyr log feature? When does zephyr write anything?

ert78gb:
fyi: currently works only with the right half

kareltucek:
On configuration saves. On bluetooth and connection related changes. (like connecting and disconnecting a dongle or bridge cable)

ert78gb:
@kareltucek Sometimes the zephyr logs of the Dongle returns with empty string or not visible characters. 
I forget to save the values. I can't constantly reproduce it. The zephyr log bit is 1 of the device state but the zefyr buffer "empty". I haven't experienced it with the left and right half.

PR #2646, opened 2025-06-14
fix: escape html of status buffer

ert78gb:


ert78gb:
@kareltucek could you test it, please.

kareltucek:
Works perfect! Thanks!

PR #2647, opened 2025-06-14
fix: zephyr shell: newlines

ert78gb:


ert78gb:
@kareltucek Could you test it please?. I did not test it because my UHK 80 is not with me.

PR #2649, opened 2025-06-14
fix: prevent crash when can't parse macro command line from status buffer

ert78gb:
The https://github.com/UltimateHackingKeyboard/firmware/releases/tag/v14.2.0 changed the syntax of the macro command error. 

```log
> 4 | ifHold final holduKey 1 <--- before
] 4 | ifHold final holduKey 1 <--- after
```

This PR makes prevent the crashes of the Agent if can't parse the macro line informaition


kareltucek:
Ah. I am going to revert that change. It was just a temporary workaround until the html escapes are dealt with.

mondalaci:
@kareltucek So is this PR redundant, and should be closed?

kareltucek:
No. 

It is good that things don't crash.

Issue #2650, opened 2025-06-14
Copying zephyr log and html escapes


kareltucek:
Actually, I missed one thing when testing. When I copy stuff from the zephyr log terminal: 

![Image](https://github.com/user-attachments/assets/12026179-3ec5-4905-b4ad-4a026ffa08d6)

PR #2661, opened 2025-07-05
fix: copy the pure zephyr log to the clipboard

ert78gb:
[#2650](https://github.com/UltimateHackingKeyboard/agent/issues/2650)


ert78gb:
@kareltucek please test it. I don't know how to create zephyr log with special html charachter

kareltucek:
@ert78gb works like a charm!

PR #2667, opened 2025-07-05
feat: compress key actions

### LINKED ISSUE #2421: Compress empty actions.
Rationale: there are usecases that use sparsely-populated layers. Possibly lots of them:

- nondefault layers, especially modifier layers
- modular keymap construction from smaller blocks, using `overlayLayer` command

These sparsely-populated layers still consume 1 or 4 (with perKeyRgb) bytes per action, even  when the layer is empty.

Let's add a new "action type" (6) describing that following n actions are empty with black perKeyRgb.

For instance the serialized config may contain `0x06 0x04`, meaning "set following 4 actions to none and set their perKeyRgb to black". 

(And of course, make Agent use this action for blocks of 3 or more empty actions if their perKeyRgb is 0 0 0.)

ert78gb:
closes #2421

Expected firmware https://github.com/UltimateHackingKeyboard/firmware/pull/1300

### DIFF
--- a/packages/uhk-common/src/config-serializer/config-items/module.spec.ts
+++ b/packages/uhk-common/src/config-serializer/config-items/module.spec.ts
@@ -49,7 +49,7 @@ describe('module', () => {
         expect(buffer).toEqual(expected)
     })
 
-    fit ('should compress 2 NoneAction with per key backlighting', () => {
+    it ('should compress 2 NoneAction with per key backlighting', () => {
         const noneAction1 = new NoneAction();
         noneAction1.r = 1
         noneAction1.g = 2


ert78gb:
maybe there is a backward compatibility issue. Because when I switched back to older Agent version with the user-config I don't get the correct Agent upgrade needed page. Maybe it is an other bug, but please wait with the merge until I find the root cause

ert78gb:
bug fixed. ready to test

ert78gb:
fixed the backlighting color issue.

kareltucek:
Yep, looks good now!

PR #2757, opened 2025-11-17
feat: macro arguments

### LINKED ISSUE #2635: Macro arguments
This is the UI specification of https://github.com/UltimateHackingKeyboard/agent/issues/2634

Image

[[...]](https://iconduck.com/icons/293706/data-type-array) is a toggle button.
- When no arguments are present, set it to untoggled.
- When no arguments are present and it's clicked, add $macroArg.1
- When at least an argument is present, set it to toggled.
- When at least an argument is present and it's clicked, don't do anything.
- Make the tooltip "Add macro argument" when it's untoggled.

When a macroArg.* [minus](https://fontawesome.com/icons/circle-minus?f=classic&s=solid) button is clicked, remove the relevant macro argument and recalculate the macroArg.* numbers so that they start with 1 and increase one by one. Make its tooltip "Remove macro argument"

Only use a single macroArg.* [plus](https://fontawesome.com/icons/circle-plus?f=classic&s=solid) button for the last macro argument. When clicked, add a new argument after the last one. Make its tooltip "Add macro argument"

Render macro key actions to include their arguments, such as "myMacro arg1 arg2"

ert78gb:
closes #2635

### DIFF
--- a/packages/uhk-web/src/app/components/popover/tab/macro/macro-tab.component.ts
+++ b/packages/uhk-web/src/app/components/popover/tab/macro/macro-tab.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges } from '@angular/core';
+import { ChangeDetectorRef, ChangeDetectionStrategy, Component, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges } from '@angular/core';
 import { faCirclePlus, faCircleMinus, faPlus, faUpRightFromSquare } from '@fortawesome/free-solid-svg-icons';
 import { Store } from '@ngrx/store';
 import { Subscription } from 'rxjs';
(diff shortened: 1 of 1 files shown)


ert78gb:
> For me, the actual frustration came from clicking the plus sign by accident

In this case, better to add some confirmation to the 'Create macro from this key' click.

I am not against @kareltucek request. Let's see @mondalaci opinions

kareltucek:
I still feel we are complicating it. I see no principal difference between following three scenarios, and am confident that in the first two, the macro arguments should remain unchanged. I think this is the correct behavior because I expect arguments to be used as a way to add additional metadata to a key, rather than a way to "just" configure a macro action behavior. I.e., I see them more as a property of a key rather than of the macro action.

- Assume we have a play macro action with arguments. Now just change the macro. (The arguments remain. I believe this is the desired behavior.)

- Assume we have a play macro action with arguments. Now manually go to the macro editor via left pane and create a new macro. Return back to the key popover and just change the macro of the play macro action. (This is exactly the same as the previous scenario, isn't it?)

- Assume we have a play macro action with arguments. Now click the plus button, create the macro and jump back to the original keymap. Why should this behave differently than the previous cases?

ert78gb:
Ot the tab there is the 'Jump to macro'  button. In the following use case

- Open the macro tab
- Add/modify macro argument
- Click on the 'Jump to macro'  link
- Navigate back to the key. The macro arguments modifications are gone

Most of the time I update the user configuration immediately. The key action popover is an exception because we save only valid key action. Lucky users can't set invalid PlayMacroAction.

kareltucek:
I definitely support consistent behavior, but think that in the above cases the right consistent behavior always is to persist the changes when leaving the key popover, including the jump to macro action.

ert78gb:
> the right consistent behavior always is to persist the changes when leaving the key popover, including the jump to macro action.

done

PR #2772, opened 2025-11-25
Add GUI option to reboot the connected UHK

FreedomBen:
Add a menu item called "Reboot device" under the "Device" section. This menu allows sending a reboot command to the connected UHK.

This functionality was already implemented but not directly exposed in the GUI.  It can be useful for troubleshooting a misbehaving device.

_Disclosure:  AI was used to assist in creating this PR, but the code was reviewed and tested by me (still human last I checked)_

Question:  Is there a better place for the reboot menu?


FreedomBen:
Fair enough, thank you!

PR #2798, opened 2025-12-25
Clarify Linux/Ubuntu installation instructions in README

### LINKED ISSUE #2797: AppImage not working on Ubuntu 25.10; libfuse problems
I had trouble running UHK Agent on Ubuntu 25.10 -- I first got this:

```
%` ./UHK.Agent-8.0.1-linux-x86_64.AppImage
dlopen(): error loading libfuse.so.2

AppImages require FUSE to run.
You might still be able to extract the contents of this AppImage
if you run it with the --appimage-extract option.
See https://github.com/AppImage/AppImageKit/wiki/FUSE
for more information
```

It looks like it needs `libfuse2`; I have `libfuse3` installed. (See [this issue](https://github.com/upscayl/upscayl/issues/1385)). I installed `libfuse2t64` and tried again, but I got:

```
% ./UHK.Agent-8.0.1-linux-x86_64.AppImage
The setuid sandbox is not running as root. Common causes:
  * An unprivileged process using ptrace on it, like a debugger.
  * A parent process set prctl(PR_SET_NO_NEW_PRIVS, ...)
Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
[663429:1225/055730.130874:FATAL:content/browser/zygote_host/zygote_host_impl_linux.cc:211] Check failed: . : Invalid argument (22)
zsh: trace trap (core dumped)  ./UHK.Agent-8.0.1-linux-x86_64.AppImage
```

I went down a bit of a rabbit hole from there; I tried running it with sudo. I also tried using the `--appimage-extract` thing suggested initially. That in turn complained about root ownership and 4755 permissions, but even with that it didn't work.

In the end, I got it to run with `--no-sandbox`.

It would be useful if there was some documentation on this -- maybe say something in the README?

dandrake:
Updated instructions for downloading Agent on Linux.

Addresses / fixes #2797 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@ Agent is the configuration application of the [Ultimate Hacking Keyboard](https:
 
 Please do not build Agent from source unless you want to develop it. Using an existing release is far easier for everyone involved:
 
-* If you already own a UHK then [download the latest desktop build of Agent](https://github.com/UltimateHackingKeyboard/agent/releases/latest) from the releases page. On Linux, download the AppImage, make it executable, and run it. If you don't own a UHK then you won't get past the opening screen!
+* If you already own a UHK then [download the latest desktop build of Agent](https://github.com/UltimateHackingKeyboard/agent/releases/latest) from the releases page. If you don't own a UHK then you won't get past the opening screen!
+  * On Linux, download the AppImage, make it executable, and run it with the `--no-sandbox` option. On recent versions of Ubuntu, you may need to install the `libfuse2t64` package.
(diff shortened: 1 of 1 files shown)


mondalaci:
@ert78gb @dandrake This is a first. I'm wondering under what circumstances this issue occurs. Nobody else has reported it yet.

ert78gb:
I also checked the behaviour of Ubuntu 25. It uses the fuse3, but electron uses fuse 2. The new fuse3 is not compatible with fuse 2 and newer distros does not ship fuse 2 by default. 

Honestly I don't know when will it solved because there is a closed issue in the electron https://github.com/electron/electron/issues/44131 that references to chromium https://issues.chromium.org/issues/410878548 that referencing back to electron :(  

Currently, fuse2 manual install is the solution/workaround.

Issue #2799, opened 2025-12-30
Index host connections from 1


kareltucek:
Please, in the switch action key popover tab, show indexes starting from 1 instead of 0. This is purely visual request - leave the internal representation zero indexed. 

<img width="1080" height="648" alt="Image" src="https://github.com/user-attachments/assets/eaa055cd-5fe3-461a-999e-b918b67f7e5e" />

PR #2802, opened 2026-01-08
feat: index host connections from 1

ert78gb:
close [#2799](https://github.com/UltimateHackingKeyboard/agent/issues/2799)

Also updated the index of the key visualisation 
Screenshot 2026-01-08 at 19 52 49


kareltucek:
LGTM!

PR #2815, opened 2026-01-20
fix: paste of macro-command editor

### LINKED ISSUE #2693: Pasting into macro action entry fields is broken
**Agent release v8.0.0 (Windows)**

•In Agent's Macros tab, when trying to paste into the "Add macro action -> Macro commands" entry field, both Ctrl+V and the right-click context menu work as expected.
•Pasting into the "Add macro action -> Type text" entry field works fine using Ctrl+V, but when trying to paste via the right-click context menu, no menu appears.

When trying with **Agent master branch version:** https://github.com/UltimateHackingKeyboard/agent/actions/runs/16455090428:
I can't paste anything into the "Macro commands" entry field at all. Doesn't work using the right-click context menu, or Ctrl+V.🤷‍♂️
___
I also noticed that the first attempt to bring up the right-click menu causes it to briefly appear, and then disappear. The second attempt usually brings up the menu as expected.

ert78gb:
close #2693

### DIFF
--- a/packages/uhk-agent/src/electron-main.ts
+++ b/packages/uhk-agent/src/electron-main.ts
@@ -42,6 +42,7 @@ if (options.help) {
     process.exit(0);
 }
 
+const isMac = process.platform === "darwin";
 const logger = new ElectronLogService();
 logger.setLogOptions(getLogOptions(options));
 logger.misc('[Electron Main] command line arguments', options);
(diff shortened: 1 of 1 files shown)


ert78gb:
@mondalaci @kareltucek I tested it on 3 OS and works fine. If you have time please double check. thx

PR #2858, opened 2026-03-21
feat: Typing behavior, v14

ert78gb:
closes #2844

firmware build: https://github.com/UltimateHackingKeyboard/firmware/actions/runs/24790226950?pr=1483


kareltucek:
I have pushed some rewordings and preset values.

There are some weird behaviors. 

When I open Agent, I see Simple preset with Timeout 505ms.

Often when I click a preset, I see it switch to that preset (the preset becomes blue, values change), but then it immediately switches back to Custom.

ert78gb:
@kareltucek 

> When I open Agent, I see Simple preset with Timeout 505ms.

Do you mean 350? It is comes from the default user configuration https://github.com/UltimateHackingKeyboard/agent/blob/53fd26c48d543d45f46875f2d8f452bef5173729/packages/uhk-common/user-config-80.json#L68

Maybe we have to adjust the values here too. The user configuration contains the "default values of the old advanced" strategy. I mean the default strategy is Simple but the values are of the Advanced if I remember well.

> Often when I click a preset, I see it switch to that preset (the preset becomes blue, values change), but then it immediately switches back to Custom.

Could you write the click sequence? I can't reproduce the issue.

kareltucek:
> Do you mean 350? It is comes from the default user configuration 

No, I mean 505. But I can't reproduce it now, so nevermind.



>>    Often when I click a preset, I see it switch to that preset (the preset becomes blue, values change), but then it immediately switches back to Custom.
>
> Could you write the click sequence? I can't reproduce the issue.

- Click HRM
- Click onto the Timeout bar (changing its value)
- Click Simple

Meanwhile the presets go Custom -> HRM -> Custom -> Simple -> Custom

ert78gb:
@kareltucek I still can't reproduce the bug
uhk-typings

kareltucek:
https://github.com/user-attachments/assets/9ce07f2a-116e-418b-8ee5-55b08cc337a1

mondalaci:
I can't reproduce this issue: When switching to a preset, Agent never switches back to custom without further user interaction.

kareltucek:
The 505 bug - the problem is that it shows Simple, but at the same time shows advanced role values. That's where 505 comes from.:

2026-04-27-174614_2656x2102_scrot

[UserConfiguration-ex1_505.json](https://github.com/user-attachments/files/27132418/UserConfiguration-ex1_505.json)

------------------

This is export of the config after the simple/custom issue reproduction. Values look fine. 

Another way to reproduce this issue is to simply import the config and observe that it shows Custom and not Simple.

[UserConfiguration-ex2_custom.json](https://github.com/user-attachments/files/27132504/UserConfiguration-ex2_custom.json)

ert78gb:
> The 505 bug - the problem is that it shows Simple, but at the same time shows advanced role values. That's where 505 comes from.

The `secondaryRoleStrategy` is `Simple` in the user configuration and it has precedence. So independently from the `Advanced` values. It is a backward compatible behaviour of the old UI when we just disabled the `Advanced` input controls without any value modification.

> [UserConfiguration-ex2_custom.json](https://github.com/user-attachments/files/27132504/UserConfiguration-ex2_custom.json)

Still can't reproduce the issue :(

mondalaci:
@ert78gb Which Agent build should Karel use to reproduce this issue? Please provide the relevant GitHub action link.

@kareltucek Which firmware build should be used?

kareltucek:
Hmmm. I can't reproduce it with the latest CI build either. 

Also, the CI-built Agent is running *much* faster than a local-built one.

kareltucek:
I can't reproduce it with the CI build actually. Also, the CI-built Agent is running much faster than locally-built one.

So nevermind.

I guess we can merge firmware branch into master? I am not sure if we want to merge agent side into master now or just before release.

mondalaci:
Please merge the firmware branch into master, publish a new release, and afterward we'll merge the Agent feature and publish a new Agent release.

PR #2871, opened 2026-04-20
Temporary fixes.

kareltucek:
I am a bit confused by the original PR. This is what I used for testing. Please revise and apply corresponding fixes in the original branch.

My confusion is on these two:

- Why that reenumeration? 
- Shouldn't right be flashed with kboot native too?

### DIFF
--- a/packages/usb/kboot-native-firmware-upgrade.ts
+++ b/packages/usb/kboot-native-firmware-upgrade.ts
@@ -1,39 +1,29 @@
 #!/usr/bin/env -S node --loader ts-node/esm --no-warnings=ExperimentalWarning
 
 import fs from 'fs';
-import * as path from 'path';
-import { LEFT_HALF_MODULE } from 'uhk-common';
-import { getCurrentUhkDeviceProduct, getDeviceFirmwarePath, getFirmwarePackageJson } from 'uhk-usb';
+import { RIGHT_TRACKBALL_MODULE } from 'uhk-common';
+import { getCurrentUhkDeviceProduct } from 'uhk-usb';
 
 import Uhk, { errorHandler, yargs } from './src/index.js';
 
 (async function () {
     try {
         const argv = yargs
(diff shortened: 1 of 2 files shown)


packages/usb/kboot-native-firmware-upgrade.ts
```diff
@@ -1,39 +1,29 @@
...
-        const packageJson = await getFirmwarePackageJson({
-            packageJsonPath,
-            tmpDirectory: firmwarePath
-        });
-        const rightFirmwarePath = getDeviceFirmwarePath(uhkDeviceProduct, packageJson);
-
-        await operations.updateRightFirmwareWithKboot(rightFirmwarePath, uhkDeviceProduct);
-        const leftFirmwarePath = path.join(firmwarePath, 'modules/uhk60-left.bin');
-        await operations.updateModuleWithKbootNative(leftFirmwarePath, uhkDeviceProduct, LEFT_HALF_MODULE);
+        await operations.updateModuleWithKbootNative(firmwarePath, uhkDeviceProduct, RIGHT_TRACKBALL_MODULE);
```
ert78gb, line 26: I created this script to test the feature without Agent. 

I am fine if have to provide the `firmwarePath` to the script and not the directory. If we keep the script then we have to provide the module as a command line argument too. I don't know will this script replace the old firmware upgrade process in the factory.

This script has been created only for this PR so no one use it, we could easily modify or delete it.

ert78gb:
> Why that reenumeration?

Just to be sure the Normal keyboard is enumerated. The `UhkOperations.updateModuleWithKbootNative` maybe will be called in the full firmware upgrade process where the right half is in "buspal", "bootloader" mode or just booting.


> Shouldn't right be flashed with kboot native too?

We use kboot for the right half flashing from the beginnings, so this issue is focusing only on the module firmware upgrade.

kareltucek:
Ah, right!. Closing this then.

PR #2905, opened 2026-05-04
Fix secondary role wording.

kareltucek:


ert78gb (approved):

mhantsch (changes requested):
For the HRM presets, it should be `secondaryRoleAdvancedStrategyTriggerFromSameHalf: false`. It is important that they do not accept same-side triggers (except timeout, of course).

For the Timeout preset, I guess it does not matter.

kareltucek:
Sure! Done. Merging.

PR #2913, opened 2026-05-06
tooltip rewordings on Typing behaviour page

mhantsch:
- tooltips reworded for more clarity on the Typing behaviour page
- flip semantic from Suppress to Allow for same-side triggers


kareltucek (commented):
Yep, looks good!

PR #2956, opened 2026-06-01
feat: command prompt for zephyr log

### LINKED ISSUE #2812: [Feature] Command prompt for zephyr log.
We would like to be able to executed zephyr shell commands directly from the agent log/console in advanced settings

I.e., in advanced settings, somewhere bellow the logging console, add an input field (simple text area? or does the console widget support something already?). 

After a command is entered, send it to UHK via the 0x1e usb command. It expects a c-style, null terminated string.

kareltucek:
closes: #2812

It works with the bundled firmware. Shell command to test the function is `uhk connections`

Clipboard shortcuts:
- copy: `CTRL + SHIFT + C` on all OS and `CMD + C` also works on Mac
- paste: `CTRL + SHIFT + V` on all OS, `CMD + V` also works on Mac and `CTRL + V` works on Linux and Windows

The `CTRL + C` is reserved as terminal sequence.


ert78gb:
I havent test Karel's extension but would not laggy to send every character independently to the keyboard?

kareltucek:
Yes, this version is laggy. But I think that if you tweak timings a bit (like scheduling state reads in 10 or so ms after any input, it could be pretty fine).

packages/uhk-web/src/app/components/zephyr-terminal/zephyr-terminal.component.ts
```diff
@@ -0,0 +1,149 @@
...
+                if (log.device === this.uhkDevice.logName) {
+                    this.terminal.write(log.log);
+                    foundHistory = true;
+                }
+            }
+
+            // initialize the terminal if no history found
+            // it will write the correct prompt
+            if (!foundHistory) {
+                this.dispatchTerminalInput('clear\n')
```
ert78gb: @kareltucek What do you think is there a better way to initialise the prompt? 
I don't really wanna store any device specific prompt string in Agent if not necessary.

kareltucek: Just `this.dispatchTerminalInput('\n')` please. I.e., If there is already something in the buffer, please do not clear it. Show it.

ert78gb: done. If have to modify something then let me know otherwise this PR is done from my side.

kareltucek: @ert78gb, multiple device view doesn't work on my side for some reason. Also, there still is the old prompt textarea, which is no longer relevant.

kareltucek: @ert78gb please see UHK Dev. Also, I have used claude to hotfix some issues, you can see it for reference at vt100_poc2

kareltucek:
Works like a charm now!

PR #2971, opened 2026-06-19
feat: make keyboard name persistent across UserConfig imports.

### LINKED ISSUE #2422: Move device name from UserConfig to HardwareConfig
Motivation:

2024-09-22-221939_804x364_scrot

kareltucek:
Changelog:

- Make keyboard name persist across UserConfig imports, in order to support UserConfigs shared across multiple keyboards.

Closes #2422

### DIFF
--- a/packages/uhk-web/src/app/store/effects/user-config.ts
+++ b/packages/uhk-web/src/app/store/effects/user-config.ts
@@ -316,8 +316,10 @@ export class UserConfigEffects {
     loadUserConfigurationFromFile$ = createEffect(() => this.actions$
         .pipe(
             ofType<LoadUserConfigurationFromFileAction>(ActionTypes.LoadUserConfigurationFromFile),
-            map(action => action.payload),
-            map((payload: LoadUserConfigurationFromFilePayload) => {
+            withLatestFrom(this.store.select(getUserConfiguration)),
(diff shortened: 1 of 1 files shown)


packages/uhk-web/src/app/store/effects/user-config.ts
```diff
@@ -316,8 +316,10 @@ export class UserConfigEffects {
     loadUserConfigurationFromFile$ = createEffect(() => this.actions$
         .pipe(
             ofType<LoadUserConfigurationFromFileAction>(ActionTypes.LoadUserConfigurationFromFile),
```
ert78gb, line 318: This event is dispatching when 
- restoring user configuration from history
- importing user configuration

Do we wanna overwrite the device name in both cases?
I think when we restore from history we want the exactly same user config that we restore.

And the question is same for import user config. Maybe someone do custom backup/export and when restore the config they want back the original setup.

 I could image some confirmation, but maybe it is over kill.

Otherwise the PR is ok for me

kareltucek: I think yes.

Excluding history would mean that the name would appear to change or not change at random, and the history titles would change too.

ert78gb (approved):

Issue #2692, opened 2025-07-31
[Suggestion] Names and copy-paste for layers


Totoro83:
First of all, I'm putting 2 different suggestions in the same issue, because in my use case they are connected. If you think this should be split in two issues just let me know, I'll split them.

And second, I understand that my use case is not typical. But I also think that both suggestions could be useful to others, that's why I decided to propose these two things.

Basically, I've an heavily modified keyboard. I have so many layers that I needed to create an additional keymap, as I already used all the normal layers (base, mod, mouse, and Fn from 1 to 5.) It works great most of the times for my use case, but I still am tweeking the maps here and there for things I use less, moving functionalities and opening agent from time to time to check what I changed recently. It's not something I do often, but still it's something I need to do sometimes.

But here is where things get a bit difficult. My layers are not ordered in a logical way, and I always struggle to find the right one. And their names don't help. Reading Fn1 instead of Fn3, for example, does not give me any clue. Of course I already know some of the layers very well and I'm able to figure out what their content is, but for some of them I have to browse among them for a bit, even half a minute sometimes, to figure out which layer I'm looking for. I know where these layers are in my keyboard and which keys I have to press to trigger them, but on Agent they are unlogically positioned and difficult to find. The first problem would be easy to solve in theory, it's enough to move the layers in the right place, but to copy an entire layer is very time consuming, especially when the layers can change while I work on them.

My proposals are the following:

- Add the possibility to copy and paste layers. If you copy a layer, you can paste the content in another layer on the same layout or in another one.
- Add the possibility to change names to the layers. It does not have to be necessarily a deep change, the names can be the same on macros (even if I'm not against a similar change). It would be enough to show both names when a layer is selected, and possibly the renamed name (or a short version of it) in the tabs on top. This way, the name would act as a suggestion.

Totoro83:
Yeah, it's perfectly understandable that you don't want to change the layer names to keep things simple. I still think that even just a very short description of the layer besides the name written on the tabs (like for example `Fn (Desc)`, where "Desc" can be just a few characters long) would be very useful, but I imagine that it could clutter the UI too much.

But even just having the copy-paste functionality would improve the situation drammatically.

And anyway thanks for the consideration 😁

PR #2986, opened 2026-07-07
feat: add copy/paste layer support

mondalaci:
## Summary
Implements copy/paste of layers, as requested in #2692.

- Adds **Copy** and **Paste** buttons to the layer header (to the right of the layer button group, left of the help `?` icon). The Paste button appears once a layer has been copied.
- **Copy** stores the current layer in an in-memory "virtual clipboard" (NgRx store), so it can be pasted onto any layer — including in another keymap or on another device.
- **Paste** overwrites the currently selected layer with the copied layer.

## Cross-context safety
Paste is guarded so it can never produce an invalid configuration:

- **Macros** are re-linked by name (macro names are unique). If the target config has no macro with that name, the key becomes a `NoneAction` (color preserved).
- **Switch-keymap** references are kept only if the referenced keymap abbreviation still exists; otherwise the key becomes a `NoneAction` (color preserved). This avoids serializing a `-1` keymap index.
- **Structural merge:** modules are matched by id **and** key count, so a layer copied from a different keyboard/layout (e.g. UHK60 ↔ UHK80, ANSI ↔ ISO) can never desync the target's module structure. Modules without a compatible counterpart keep their existing content.

`SwitchLayerAction`, keystroke secondary roles, and `ConnectionsAction` host-connection references are left untouched — they serialize to fixed/valid values and at worst become semantically "dead" keys until the relevant layer/connection exists (no corruption).

## Notes
- Partial pastes (some incompatible modules skipped) are intentionally silent.
- No undo for paste yet; a general undo/redo is planned separately and would cover this for free.

## Test plan
- [ ] Copy a layer, verify the Paste button appears.
- [ ] Paste onto a different layer in the same keymap; verify key actions, colors, and macro/switch-keymap keys are correct.
- [ ] Paste into a different keymap; verify macro keys re-link by name and switch-keymap keys survive/clear correctly.
- [ ] Copy on one device, paste on another (different layout); verify no structural corruption and config saves/loads cleanly.
- [ ] Verify buttons don't appear on the add-keymap preview screen.


ert78gb:
superseded by #2992

PR #2993, opened 2026-07-12
feat: auto-group macros in the sidebar

### LINKED ISSUE #1845: [Feature Request] Ability to group macros in the sidebar
I made a lot of macros lately and cluttered my sidebar. Would it be possible to add the ability to organize macros into collapsible folders/groups?

mondalaci:
## Summary
- Adds optional collapsible macro groups in the sidebar, derived from macro names at display time (no config schema change).
- Splits on non-alphanumeric characters by default; optional camelCase splitting for names like `bindMouseLeft`.
- Adds Settings controls with tooltips, dependent-option disabling, and persisted Agent preferences (off by default).
- Handles smart-macro `$` prefixes and avoids incorrect deeper nesting when siblings do not share a next segment.

Closes #1845

## Test plan
- [ ] Open Settings → Macro sidebar grouping: verify layout, tooltips, and that dependent options disable when grouping is off
- [ ] Enable grouping with macros like `Doom: Chainsaw`, `Doom: Plasma gun` → collapsible `Doom` folder
- [ ] Enable camelCase splitting with `bindMouseLeft` / `bindMouseRight` → `bind` group
- [ ] Verify `$onInit` / `$onJoin` group under `$on`, not `on`
- [ ] Set max nesting depth to 2 with `Open daily sites` / `Open weekly sites` → `Open` > `daily sites`, `weekly sites` (not full names)
- [ ] Confirm grouped macro leaves show full name in hover tooltip when shortened
- [ ] Run `npm test` in `packages/uhk-common`

### DIFF
--- a/packages/uhk-common/src/util/group-macros-by-name.ts
+++ b/packages/uhk-common/src/util/group-macros-by-name.ts
@@ -19,6 +19,9 @@ interface MacroGroupingItem<TMacro extends GroupableMacroItem> {
     macro: TMacro;
 }
 
+const MACRO_NAME_SEPARATOR = /[^\p{L}\p{N}$]+/u;
+const LEADING_MACRO_NAME_SEPARATOR = /^[^\p{L}\p{N}$]+/u;
+
 export function groupMacrosByName<TMacro extends GroupableMacroItem>(
     macros: TMacro[],
     settings: MacroGroupingSettings
@@ -111,7 +114,7 @@ function getMacroMenuTreeNodeLabel<TMacro extends GroupableMacroItem>(node: Macr
 }
 
 export function splitMacroName(name: string, camelCaseSeparation: boolean): string[] {
-    let parts = name.split(/[^a-zA-Z0-9$]+/).filter(part => part.length > 0);
+    let parts = name.split(MACRO_NAME_SEPARATOR).filter(part => part.length > 0);
 
     if (camelCaseSeparation) {
         parts = parts.flatMap(splitCamelCaseSegment);
@@ -156,7 +159,7 @@ function splitFirstSegment(
 
     const head = segments[0];
     const afterHead = name.slice(name.indexOf(head) + head.length);
-    const rest = afterHead.replace(/^[^a-zA-Z0-9$]+/, '');
+    const rest = afterHead.replace(LEADING_MACRO_NAME_SEPARATOR, '');
 
     return { head, rest };
 }
--- a/packages/uhk-common/src/util/group-macros-by-name.ts
+++ b/packages/uhk-common/src/util/group-macros-by-name.ts
@@ -7,7 +7,6 @@ export interface GroupableMacroItem {
 
 export interface MacroMenuTreeNode<TMacro extends GroupableMacroItem = GroupableMacroItem> {
     children?: MacroMenuTreeNode<TMacro>[];
-    displayName?: string;
     label?: string;
     macro?: TMacro;
     path?: string;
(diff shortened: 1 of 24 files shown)


packages/uhk-common/src/util/group-macros-by-name.ts
```diff
@@ -0,0 +1,192 @@
...
+function getMacroMenuTreeNodeLabel<TMacro extends GroupableMacroItem>(node: MacroMenuTreeNode<TMacro>): string {
+    if (node.type === 'group') {
+        return node.label || '';
+    }
+
+    return node.displayName || node.macro?.name || '';
+}
+
+export function splitMacroName(name: string, camelCaseSeparation: boolean): string[] {
+    let parts = name.split(/[^a-zA-Z0-9$]+/).filter(part => part.length > 0);
```
ert78gb: @mondalaci 
I think we would like to support non english macro names too? I mean we would like to split the `Az én macróm` => `Az, én, macróm` instead of `Az, n, macr, m`

Do we keep only ASCII numbers as part of the word or support other unicode numbers too? 
Like macro name `2²` will `2` with the current regexp. It is will less used but the non ASCII macro name is more frequent

mondalaci:
Good catch! Just committed. How about now?

kareltucek:
Two suggestions:
- Visually distinguish group headlines.
- Always list full name even inside of a group.
  - Then, if there is a macro named same as an identified group, list it *inside* of that group rather than outside.

2026-07-14-153401_315x350_scrot
2026-07-14-153351_372x340_scrot

mondalaci:
By headlines, I assume you mean non-leaf items. I've tried to make them more prominent, but it looked overkill. The current indentation and the downward-pointing arrow should indicate that they're non-leaf nodes.

Implemented the rest of your suggestions.

ert78gb:
Just a question I don't know what is the good behaviour. If I open the popover of a key action and navigate to the `Macro` tab. And Click on the `Jump to macro` button. Should the macro group expand if it was collapsed?

And may be worth to add a 3rd macro grouping option about how to show macros when Agent opens. In collapsed or opened. It is not a strong option I have to use this feature more to see better

mondalaci:
Great catch regarding the jump to macro! The macro should expand. Just submitted a commit.

Let's not add a default opening option for now. I think the current behavior is fine, and we'll add further options if requested.

kareltucek:
Looking into the code/dom now, and... it seems that *everything* is a leaf item in that tree. That greatly limits any indent-related formatting. 😑

kareltucek:
Just a suggestion:

2026-07-15-113517_1300x1899_scrot

pcooke9:
Looks nice! Is this in a usable state now, or should I wait? Because I really don't wanna wait...😁

kareltucek:
Usable state :)

packages/uhk-web/src/app/components/side-menu/side-menu.component.scss
```diff
@@ -142,8 +142,22 @@ ul {
...
+
+    &__macro-tree {
+        width: auto;
+        margin-left: $macro-tree-base-indent;
+    }
+
+    &__macro-group-children {
+        width: auto;
+        padding-left: $macro-tree-indent-step;
+        border-left: 1px solid silver;
```
ert78gb, line 160: I havent checked out the PR just looked into it. Is this silver looks goon in light and dark mode? Should not we use a predefined "silver" color?

mondalaci: Yes, the silver color does look good in light and dark modes alike.

kareltucek: > Should not we use a predefined "silver" color?

We definitely should.

PR #2998, opened 2026-07-14
feat: expand macro field limits according to user configuration data types

### LINKED ISSUE #2754: Expand macro field limits according to user configuration data types
See https://forum.uhk.io/t/move-the-cursor-to-absolute-x-y-using-macros/2730 and check every field.

mondalaci:
## Summary
- Export integer range constants from the config serializer and use them in the macro editor.
- Expand mouse move/scroll coordinate limits from ±9999 to the Int16 range (±32767).
- Expand delay limit from 65 s to 65.535 s (UInt16 milliseconds).
- Clamp out-of-range coordinate input so values like 80000 are corrected immediately.

Fixes #2754

## Test plan
- [ ] In a move-pointer macro action, enter `80000` for X or Y and confirm it clamps to `32767`.
- [ ] Enter `-40000` and confirm it clamps to `-32768`.
- [ ] Enter a delay of `100` seconds and confirm it clamps to `65.535`.
- [ ] Save the macro and verify the configuration still serializes correctly.

### DIFF
--- a/packages/uhk-common/src/config-serializer/assert.ts
+++ b/packages/uhk-common/src/config-serializer/assert.ts
@@ -1,31 +1,41 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+export const INT8_MIN = -0x80;
+export const INT8_MAX = 0x7F;
+export const INT16_MIN = -0x8000;
+export const INT16_MAX = 0x7FFF;
+export const INT32_MIN = -0x80000000;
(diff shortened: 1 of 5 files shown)


ert78gb:
@mondalaci The logic does not allow to save invalid value, but the input control allow to type values out of the range. It is not a new behaviour. Would we like to fix it in this PR or we leave it as is

ert78gb (approved):

PR #3002, opened 2026-07-15
feat: show macro key assignments on the macro page

### LINKED ISSUE #2743: On the macro page, list the keys the macro is assigned to
Also, navigate to the relevant layer and open the key action popover for the key.

See https://forum.ultimatehackingkeyboard.com/t/macro-jump-to-key/2685/12

mondalaci:
Resolves #2743

## Summary
- On macro pages, show **Used on:** with comma-separated links to each assigned key (`keymap ⭢ layer ⭢ key` using default QWERTY labels).
- Links navigate to `/keymap/{abbreviation}?layer=&module=&key=` and open the key action popover.
- Remove redundant back-navigation from the macro page and keymap popover; keep **Jump to macro** for reverse navigation.
- Fix **Jump to macro** so it no longer marks the configuration dirty or shows **Save to keyboard**.

## Test plan
- [ ] Open a macro assigned to one or more keys and verify **Used on:** links appear under the title.
- [ ] Click a **Used on** link and confirm it opens the correct keymap/layer/key popover.
- [ ] From the popover Macro tab, click **Jump to macro** and confirm navigation works without showing **Save to keyboard**.
- [ ] Assign a macro to a new key via **Assign new macro** and confirm saving still works normally.
- [ ] Verify module settings back links still render as `Back to [name] keymap`.


ert78gb:
This PR contains the commits of other PR. As far as I see the UI part of the other PR is havent fully decided so this PR also blocked.

mondalaci:
Rebased onto `master` so this PR now contains only the macro key-assignment commit (`91de3252`), without the macro sidebar grouping commits from the other PR.

The diff is 17 files / +344 −35, scoped to issue #2743 only.

ert78gb:
I summarise here my problem with this PR and macro grouping
- the `uhk-common` package contains UI related code that have to be live in the `uhk-web` package. I did not mention it in the macro grouping PR because I was happy about the unit test, but I don't want it to be a trend. I thought I will add test runner to the `uhk-web` and we will port it. 

- Earlier I followed the smart/dummy component pattern and mostly the store contained the main calculations now the deep components reads from the store and do calculations. The dummy components just render the input and propagate the event. Most of the exception is inherited. 

- uses to much custom css. Would be nice to consolidate them for a unified styling and not add more and more. Every time when have to modify the css have to think about why is it there? Can I remove or what will collapse?

- the use of optional chaining `?`  and nullish coalescing operator `??` operators is a code smell for me. Something is not properly initialised or read data from wrong selector. I maintained an old react app that was full of optional chaining and it was nightmare. Always thinking about is it intentionally optional or just the developer was lazy.

- methods/functions with more than 3 parameters are hard to maintain in long term, like `createKeymapWithMacroAssignment`

I am not agains ai development but somehow have to improve their skills to decrease the cost of afterwork. Or we don't do after work and we will see the outcome after few months. I especially like I don't have to fight with pixels but I don't like the mess and noise that generating.

mondalaci:
Addressed the architectural points from the review:

- Moved `findMacroKeyAssignments` (and its tests) from `uhk-common` into `uhk-web`
- Made `macro-header` presentational for Used on data; `macro-edit` derives the view models and passes them as `@Input`
- Replaced custom Used on CSS with Bootstrap utilities (`my-2`, `small`, `me-1`, `text-nowrap`)
- Reduced optional chaining / nullish coalescing in the new code; options objects for >3-parameter helpers
- Added Cursor rules so future Agent work follows the same boundaries, smart/dumb split, styling preference, parameter limit, and initialization guidance

packages/uhk-web/src/app/components/macro/edit/macro-edit.component.ts
```diff
@@ -59,20 +68,31 @@ export class MacroEditComponent implements OnDestroy {
...
-                private router: Router) {
-        this.subscriptions.add(store.select(getSelectedMacro)
-            .subscribe((macro: Macro) => {
+                private router: Router,
+                private mapper: MapperService) {
+        this.subscriptions.add(
+            combineLatest([
+                store.select(getSelectedMacro),
+                store.select(getKeymaps),
+                store.select(getDefaultUserConfiguration),
```
ert78gb, line 82: We load the default configuration into the store when opening the `Add Keymap` menu.
I do it to save some memory because it is not frequently used feature. We could load it and app start and refresh when new device is connected, but I haven't did it, because little bit confusing me to see the macro assigned on a non qwerty keymap and we so the qwerty key name. 
For example I assign the macro to the Dvorak O that is Qwerty S and I see `Dvorak for PC -> Base -> S` instead of `Dvorak for PC -> Base -> O`

ert78gb (approved):

kareltucek:
> On macro pages, show Used on: with comma-separated links to each assigned key (keymap ⭢ layer ⭢ key using default QWERTY labels).

That's not what it does. 

The list of usages is shown only if the macro editor is opened via the jump-to-macro icon. When the macro is opened from the list, it is *not* shown. Not sure if this was intended.

> Remove redundant back-navigation from the macro page and keymap popover; keep Jump to macro for reverse navigation.

Good luck searching the correct keymap for the "jump back" 😇: 

2026-08-06-151812_1777x670_scrot

mondalaci:
> The list of usages is shown only if the macro editor is opened via the jump-to-macro icon. When the macro is opened from the list, it is not shown. Not sure if this was intended.

For me, the list is visible either way.

Please provide your configuration and the steps to reproduce, including the exact macros/keymaps/layers/keys.

> Good luck searching the correct keymap for the "jump back" 😇:

You gotta admit you have a crazy config that isn't representative! :)

pcooke9:
> For me, the list is visible either way.

The list is available from both methods for me as well.

Karel does have a good point w.r.t. the jump back button though. Some of us are terrible at writing macros (not Karel 🤣), and forget where we were quite easily (even with only a few locations in the list 😅).

mondalaci:
> Karel does have a good point w.r.t. the jump back button though. Some of us are terrible at writing macros (not Karel 🤣), and forget where we were quite easily (even with only a few locations in the list 😅).

Do you mean the numerous jump-back links on the macro page? If so, how to improve them? It's the inevitable side-effect of complex configurations.

kareltucek:
> You gotta admit you have a crazy config that isn't representative! :)

Actually, I do admit it w.r.t. my report of missing "Used on" section 😅 - seems I got confused by macros that are not bound anywhere. I have got quite a number of these - macro events, but also indirections that are executed from other macros.

I would expect an empty list to be shown in these cases rather than nothing, but you are right that this isn't a representative observation, and isn't an important issue.

> You gotta admit you have a crazy config that isn't representative! :)

As for the "Used on" section, I am not sure my usecase is that niche. All I did was mapping some macros onto "all keymaps" and "all layers". 

Maybe folding the section by default would be more aesthetic in cases like these, but again, not a huge issue.

>  If so, how to improve them? It's the inevitable side-effect of complex configurations.

I think a "jump back" functionality and "list of binding sites" are two different features that should not be mixed. 

I would just show those two sections after each other.

pcooke9:
> Do you mean the numerous jump-back links on the macro page? If so, how to improve them? It's the inevitable side-effect of complex configurations.

As Karel's last comment pointed out, I just meant that the jump-back feature should probably have it's own dedicated button that's separate from the "Used on" list.

I see you just reimplemented it. 😉

pcooke9:
In the "Used on" list; it may be helpful to use a more visually prominent separation. Maybe a thickened grey "|" or something between the links would make them easier to distinguish from one another.

mondalaci:
Thanks for the suggestion! Implemented in #3046.

image

pcooke9:
Awesome, much better indeed! Especially in dark mode.

PR #3004, opened 2026-07-17
feat: minimize to system tray (#850)

### LINKED ISSUE #850: Linux: No minimize to tray of the agent
First, I am not quite sure if this "feature" was planned anyway.

But fact is, that I would like to have the agent to be able to be minimized to the tray icons on any operating system. On Linux with KDE/plasma environment there's no option to do that.

I haven't tested it on other operating systems.

mondalaci:
## Summary
- Add a **Minimize to system tray** setting on the Settings page (Electron only) so minimizing Agent hides it to the system tray instead of the taskbar
- Implement cross-platform tray support in the main process with left-click toggle, right-click context menu (Show / Quit), and Linux taskbar handling
- Add `--start-minimized-to-tray` CLI option to launch Agent hidden in the tray (e.g. for autostart entries)

Closes #850

## Test plan
- [ ] Enable **Minimize to system tray** in Settings, minimize the window, and confirm it appears in the system tray
- [ ] Left-click the tray icon to toggle show/hide; right-click to open the context menu with **Show UHK Agent** and **Quit**
- [ ] Launch with `--start-minimized-to-tray` and confirm Agent starts with only the tray icon visible
- [ ] Launch a second Agent instance while hidden in tray and confirm the window is restored and focused
- [ ] Verify normal minimize behavior when the setting is disabled
- [ ] Smoke test on Windows and macOS if available


pcooke9:
Current behavior is that after the initial Agent launch, you need to minimize to tray using the Windows top button at least once per session before the tray icon becomes available. If the user has already enabled "Minimize to tray" in Settings, I wonder if it would be better to have the tray icon be available immediately upon opening Agent every time?

Also, if the Agent window is in a _maximized_ state, and then you send it to tray by clicking the Windows minimize button at the top, it doesn't retain the maximized state when restoring from the tray (or when clicking the desktop launcher again). But once the tray icon is available, if you minimize it again using only the tray icon, it will then remember the maximized state. But still not when using the Windows minimize button. Is this the desired behavior?

Video of the issue:
https://www.youtube.com/watch?v=BoQCTJvQBOA

mondalaci:
@pcooke9 Thanks for the heads-up! Please test the new build, as it should work well.

pcooke9:
Latest build indeed remembers the maximized state now.

Only one nitpick; if Agent is in the maximized state when last closed, launching with `-- --start-minimized-to-tray` causes Agent to briefly flash a maximized window before it minimizes to tray. Doesn't happen if last state was not maxed.

Otherwise, I think everything works as expected in Win 11. 😊

mondalaci:
@pcooke9 Great catch! Nothing can escape your watchful eyes :)

Fixed.

pcooke9:
> Great catch! Nothing can escape your watchful eyes :)

That's only because I'm sitting less than a meter away from a 55" screen. 😶

Everything looks good now though!

PR #3027, opened 2026-08-03
fix: status buffer split and missing binding-site lines

### LINKED ISSUE #3025: Status buffer issues
Two issues:

1. sometimes, Agent splits the status buffer in halves, and shows only one of the halfs to the user.
2. some macro errors contain specification of binding site. Agent is removing these lines. This is confusing, because the logs look redundant. 

I suspect that 1 is caused by an unexpected content of 2.

mondalaci:
## Summary
- Drain the full status/shell buffer (up to 100 USB transfers) instead of stopping after ~1260 bytes, which left leftovers that later overwrote the UI
- Only re-fetch the status buffer when the device newly becomes available (or when dirty), not on every connection-state field change
- Preserve binding-site / nested location-stack lines in the status buffer error panel

Fixes #3025

## Test plan
- [ ] Run `diagnose usb` on UHK 60 with a long status buffer and confirm Agent shows the full message (not a short leftover fragment)
- [ ] Trigger a macro validation error that includes a binding-site line and confirm that line stays visible in the error panel
- [ ] Reconnect the keyboard and confirm the status buffer is still shown after subsequent connection-state updates (e.g. zephyr log flag changes)

### DIFF
--- a/packages/uhk-agent/src/services/device.service.ts
+++ b/packages/uhk-agent/src/services/device.service.ts
@@ -1237,6 +1237,9 @@ export class DeviceService {
                     const state = await this.device.getDeviceConnectionStateAsync();
                     if (!isEqual(state, this.savedState)) {
                         const newState = cloneDeep(state);
+                        const becameAvailable = state.hasPermission
+                            && state.communicationInterfaceAvailable
+                            && !(this.savedState?.hasPermission && this.savedState?.communicationInterfaceAvailable);
 
                         if (state.hasPermission && state.communicationInterfaceAvailable) {
                             state.hardwareModules = await this.getHardwareModules(false);
@@ -1279,7 +1282,12 @@ export class DeviceService {
                                 await this.dongleZephyrLogService.disable();
                             }
 
-                            this._checkStatusBuffer = true;
+                            // Only pull the status buffer when the device newly becomes available.
+                            // Re-reading on every connection-state field change drains leftovers from
+                            // a previous partial read (or an empty buffer) and overwrites the UI (#3025).
+                            if (becameAvailable) {
+                                this._checkStatusBuffer = true;
+                            }
                         } else {
                             deviceProtocolVersion = undefined;
                             state.hardwareModules = {
--- a/packages/uhk-web/src/app/util/status-buffer-parser.ts
+++ b/packages/uhk-web/src/app/util/status-buffer-parser.ts
@@ -68,6 +68,11 @@ function transformToErrorBlock(macros: Macro[], block: string): string {
 
     const url = `#/macro/${macro.id}?actionIndex=${macroActionIndex}&lineNr=${lineNr}&columnNr=${columnNr}&inlineEdit=true`;
     const newLine2 = `${escapeHtml(line1Result[1])}<a href="${url}">${escapeHtml(line1Result[2])}</a>`;
+    // Keep binding-site lines and nested location-stack lines (firmware may emit more than 3).
+    const extraLines = lines
+        .slice(3)
+        .map(line => escapeHtml(line))
+        .join('\n');
 
-    return `${escapeHtml(lines[0])}\n${newLine2}\n${escapeHtml(lines[2])}\n`;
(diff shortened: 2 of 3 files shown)


mondalaci:
@kareltucek This fix works for me. How about you?

kareltucek:
Yes, seems to work fine.

packages/uhk-agent/src/services/device.service.ts
```diff
@@ -1279,7 +1282,12 @@ export class DeviceService {
                                 await this.dongleZephyrLogService.disable();
                             }
 
-                            this._checkStatusBuffer = true;
+                            // Only pull the status buffer when the device newly becomes available.
+                            // Re-reading on every connection-state field change drains leftovers from
+                            // a previous partial read (or an empty buffer) and overwrites the UI (#3025).
+                            if (becameAvailable) {
```
ert78gb: The `getDeviceConnectionStateAsync` is definitely overloaded. It worth to split and refactor but it bigger task.
But I am not sure we have to read the status buffer only connecting the right half to the computer.

@kareltucek Please validate my knowledge?

As far as I know the firmware could generate status buffer content when left half oscillating or modules are connecting or disconnecting to each half. 
Maybe worth to extend the response of `GetDeviceState (0x09)`  with a status buffer has content flag like `isMacroStatusDirty`. Or use the `isMacroStatusDirty` to indicate the status buffer has value. 
If in introduce a status buffer flag indicator then we could delete the direct status buffer checks after the user configuration has been saved.

@kareltucek What do you think about it

packages/uhk-web/src/app/util/status-buffer-parser.ts
```diff
@@ -68,6 +68,11 @@ function transformToErrorBlock(macros: Macro[], block: string): string {
 
     const url = `#/macro/${macro.id}?actionIndex=${macroActionIndex}&lineNr=${lineNr}&columnNr=${columnNr}&inlineEdit=true`;
     const newLine2 = `${escapeHtml(line1Result[1])}<a href="${url}">${escapeHtml(line1Result[2])}</a>`;
+    // Keep binding-site lines and nested location-stack lines (firmware may emit more than 3).
```
ert78gb, line 71: As a future plan would be nice to write some test for these usecases

packages/uhk-agent/src/services/device.service.ts
kareltucek: > Maybe worth to extend the response of GetDeviceState (0x09) with a status buffer has content flag like isMacroStatusDirty.

I believe the isMacroStatusDirty (GetDeviceState (0x09) byte 7) does exactly that.

I.e., changes to 1 whenever Agent has started reading the buffer but has not finished, and whenever a new content is added into the buffer.

> But I am not sure we have to read the status buffer only connecting the right half to the computer.

We definitely have to read it whenever isMacroStatusDirty becomes true.

kareltucek: (I.e., this code is probably wrong and introduces a bug.)

ert78gb: Later in this function the https://github.com/UltimateHackingKeyboard/agent/blob/32fcc94805ada28409a864017de39aed1ddfaf5a/packages/uhk-agent/src/services/device.service.ts#L1299 line handle the `isMacroStatusDirty`. 

So this condition does not introduce bug. But based on what you wrote this condition does not need. Enough to check the `isMacroStatusDirty` byte.

thx

ert78gb: > I believe the isMacroStatusDirty (GetDeviceState (0x09) byte 7) does exactly that.

I deleted this code and I run the `diagnose usb` macro and the `isMacroStatusDirty` was false so we need to check the status buffer every time when the connection changing.

I reverted the `becameAvailable` the iteration read was that fixed the original bug.
I left the name of `isMacroStatusDirty` as is. because looks like it set the value only when macro invalid.

@kareltucek If you will have time please test again. I tested but double check is better :)

kareltucek: Tested, and looks good to me. Merging.

PR #3030, opened 2026-08-05
Host connections rehaul.

### LINKED ISSUE #3028: Host connections rehaul proposal.
Suggested changes:

- Show slot numbers, 1-indexed (iirc; so that they match UHK's oled numbering). 

- Show empty slots. Allow reordering them.

- Show new bonds in the list in correct slots.

- Add a new section (what should it be called?) with following two checkboxes:

  - always advertise

  - keep hosts connected

  Or maybe expose them as a single "multipoint pairing" switch button?

Firmware prerequisities:

- Place the two config variables into native UserConfig.

- Extend the api so that UHK also shows in which slots the bonds are placed.

kareltucek:
Closes #3028 . 

Just for reference atm.

2026-08-05-165150_1752x1252_scrot


mondalaci:
image

The toggle button to the left of the "Multipoint operation" title is an unusual UI pattern that we haven't used anywhere in Agent.

More importantly, when disabling the "Always advertise" and "Keep connections alive" options, the "Multipoint operation" option also gets disabled, suggesting it's redundant.

PR #3043, opened 2026-08-05
style: allow @angular-eslint/template/alt-text-error rule

ert78gb:


packages/uhk-web/src/app/components/popover/tab/keymap/keymap-tab.component.html
```diff
@@ -29,9 +29,6 @@
         </ng-select>
     </div>
     <div class="keymap-container">
-        <div class="empty" *ngIf="!selectedKeymap?.abbreviation">
-            <img src="assets/images/base-layer--blank.svg">
```
ert78gb, line 33: we don't have this asset so I dropped the whole outer element.
I tried to test it with manually modify the user config but it not loading if the config is invalid. so as I see it was a dead branch

PR #3045, opened 2026-08-07
feat: add back link when jumping from key popover to a macro

mondalaci:
## Summary
* When jumping to a macro from the key action popover (Jump to macro / Assign new macro), show a **Back to** link above **Used on:** using the same `keymap ⭢ layer ⭢ key` breadcrumb style.
* Keeps jump-back navigation separate from the assignment list, as suggested in https://github.com/UltimateHackingKeyboard/agent/pull/3002#issuecomment-5217158703.
* Reuses the shared `<back-to/>` query-param pattern from module settings.

## Test plan
- [ ] From a keymap key popover Macro tab, use **Jump to macro** and confirm a back link like `Back to QWERTY for PC ⭢ Fn ⭢ G key` appears above **Used on:**
- [ ] Click the back link and confirm it returns to the same keymap/layer/key with the popover open
- [ ] Use **Assign new macro** from the popover and confirm the same back link appears on the new macro page
- [ ] Open a macro from the side menu (no jump) and confirm no back link is shown; **Used on:** still works when assignments exist
- [ ] On the macro page with a back link, select/clear a macro action and confirm the back link remains
- [ ] Navigate from keymap to module settings and confirm `Back to [name] keymap` still works


pcooke9:
Seems to work as expected in Win 11, thanks!

PR #3052, opened 2026-08-14
fix: wait only 500 ms to OS initialise serial peripheral on linux when using mcu to upgrade firmware.

ert78gb:


kareltucek:
How confident are we that we should not decrease the timeout for other systems as well? 🤔

ert78gb:
I totally not confident what is the proper value. 

On Mac the 1 sec looks fine. I never experienced timeout.
On windows the first widows driver installation is slow. On windows the device appearing without driver, disappearing while windows install/initialise the driver, the device appearing again and useable.

The whole windows communication is much slower compare to other OS-es.

At the beginning the bootloader timeout was 5 seconds. I did not follow the firmware changes, but it decreased. So I dont know what is the good value. I also think about it to move this wait time to command line arguments to support special cases.

